PR #1, opened 2019-08-13
Fix for Value returning a Result

sfanxiang:
Because gtk-rs/glib#513.

cc @GuillaumeGomez @sdroege


sdroege:
:+1:

sdroege:
@GuillaumeGomez Let's get this in? CI won't succeed yet in any case (need examples and lgpl-docs first)

GuillaumeGomez:
:+1: 

I'll let @sfanxiang handle the merge. :)

PR #3, opened 2019-09-02

sdroege:
Fixes memory management in ShortcutLabel::new()

See https://github.com/gtk-rs/gtk/pull/885


sdroege:
Also that we have `ShortcutLabel` in `gtk4` already but didn't have it in `gtk` means that we'll probably also have other types generated here already that are not in old `gtk`. If someone wants to compare and sync...

sdroege:
Should probably also do regens for the other gtk4 crates. @GuillaumeGomez you'll do those when fixing the docs.rs stuff?

GuillaumeGomez:
Yes. I'm through them slowly and we'll need to check before next release that I didn't forgot one.

sdroege:
@GuillaumeGomez Hmm that CI failure should actually be fixed by the regeneration. It looks like the CI is using the wrong version of the gtk4 repo. Not sure how that stuff works :)

sdroege:
So what should we do here? :)

src/auto/icon_info.rs
```diff
@@ -2,7 +2,7 @@
 // from gir-files (https://github.com/gtk-rs/gir-files)
 // DO NOT EDIT
 
-#[cfg(feature = "futures")]
+#[cfg(any(feature = "futures", feature = "dox"))]
```
sfanxiang: When "dox" is set but "futures" is unset, it won't compile because it lacks the future crates.

GuillaumeGomez: Then futures crate should be included when dox is enabled. :)

sdroege: Ah that's what I was missing :)

Cargo.toml
```diff
@@ -25,7 +25,7 @@ name = "gtk4"
 [features]
 purge-lgpl-docs = ["gtk-rs-lgpl-docs", "gdk4/purge-lgpl-docs", "gsk4/purge-lgpl-docs"]
 embed-lgpl-docs = ["gtk-rs-lgpl-docs", "gdk4/embed-lgpl-docs", "gsk4/embed-lgpl-docs"]
-dox = ["gdk4/dox", "gsk4/dox", "gtk4-sys/dox"]
+dox = ["gdk4/dox", "gsk4/dox", "gtk4-sys/dox", "futures"]
```
GuillaumeGomez: :+1:

sdroege:
@GuillaumeGomez Green now, thanks!

GuillaumeGomez:
It's up to @sfanxiang :)

src/lib.rs
```diff
@@ -40,10 +40,10 @@ extern crate graphene;
 extern crate gsk4 as gsk;
 extern crate pango;
 
-#[cfg(feature = "futures")]
+#[cfg(any(feature = "futures", feature = "dox"))]
```
sfanxiang:
```suggestion
#[cfg(feature = "futures")]
```

"futures" is implied by "dox" in Cargo.toml. Just trying to be consistent with gtk3 here.

src/lib.rs
```diff
@@ -40,10 +40,10 @@ extern crate graphene;
...
 extern crate fragile;
-#[cfg(feature = "futures")]
-extern crate futures_core;
+#[cfg(any(feature = "futures", feature = "dox"))]
```
sfanxiang:
```suggestion
#[cfg(feature = "futures")]
```

Same here. "futures" is implied by "dox" in Cargo.toml.

sdroege: Done, thanks

sfanxiang:
Thanks @sdroege!

PR #10, opened 2019-11-19
Remove glib reexport

sfanxiang:
We no longer reexport glib from gtk4.

cc @GuillaumeGomez @sdroege

### DIFF
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -23,6 +23,3 @@ pub use text_buffer::TextBufferExtManual;
 pub use tree_sortable::TreeSortableExtManual;
 pub use tree_store::TreeStoreExtManual;
 pub use widget::WidgetExtManual;
-
-#[doc(hidden)]
-pub use glib::prelude::*;


GuillaumeGomez:
Hold on, we are still reexporting them: https://github.com/gtk-rs/gtk/blob/master/src/prelude.rs#L7-L8

sfanxiang:
@GuillaumeGomez I believe the idea is to stop reexporting from gtk4 (I remember asking @sdroege about this), but I'm not sure what your opinion is

sdroege:
We should still re-export the preludes of public dependencies from the prelude of each crate, i.e. usually glib and sometimes gio/gdk/etc.

But we don't re-export all contents of the prelude from the crate root (`lib.rs`).

That's what we do everywhere now, right?

sfanxiang:
> But we don't re-export all contents of the prelude from the crate root (lib.rs).

Yeah I wasn't entirely up to date with `gtk-rs`-wise changes.

> That's what we do everywhere now, right?

`examples4` and `gtk4` were originally written under the assumption that only `glib` exports glib symbols (modularity..). But after reading your comments I agree it might be too surprising if `gtk4` is different from others.

PR #19, opened 2020-10-09
groovy update: regenerate

bilelmoussaoui:
Supersedes #17


src/subclass/mod.rs
```diff
@@ -2,22 +2,18 @@
...
 pub mod box_;
-pub mod container;
 pub mod dialog;
 pub mod header_bar;
 pub mod widget;
 pub mod window;
 
 pub mod prelude {
-    pub use super::application::GtkApplicationImpl;
+    //pub use super::application::GtkApplicationImpl;
```
BrainBlasted, line 14: Why is this commented out?

bilelmoussaoui: Because I didn't had gio.Application in manual and it didn't compile, I will get it back.

BrainBlasted: Ah, okay :)

BrainBlasted:
Ported Solanum using this - looks good to me, based on that.

toger5:
Does this branch fix: https://github.com/gtk-rs/examples4/issues/8?

BrainBlasted:
@toger5 Sort-of. That's fixed at the `sys` level

BrainBlasted:
Did another port, using Social (a more complex interface than Solanum) and that worked out as well

Gir.toml
```diff
@@ -383,25 +354,6 @@ generate_builder = true
     name = "activate-link"
     inhibit = true
 
-[[object]]
-name = "Gtk.AccelGroup"
```
sdroege, line 580: You also need to `git rm src/auto/accel_group.rs` etc for these. Did you do that for the other g*k4 crates? :)

src/auto/activate_action.rs
```diff
@@ -0,0 +1,29 @@
...
+glib_wrapper! {
+    pub struct ActivateAction(Object<gtk_sys::GtkActivateAction, gtk_sys::GtkActivateActionClass, ActivateActionClass>) @extends ShortcutAction;
+
+    match fn {
+        get_type => || gtk_sys::gtk_activate_action_get_type(),
+    }
+}
+
+impl ActivateAction {
+    pub fn get() -> Option<ShortcutAction> {
```
sdroege, line 19: This needs a `type Gtk.ActivateAction` annotation (and same for the others)

src/auto/builder_scope.rs
```diff
@@ -0,0 +1,28 @@
...
+// from gir-files (https://github.com/gtk-rs/gir-files)
+// DO NOT EDIT
+
+use glib::object::IsA;
+use glib::translate::*;
+use gtk_sys;
+use std::fmt;
+
+glib_wrapper! {
+    pub struct BuilderScope(Interface<gtk_sys::GtkBuilderScope>);
```
sdroege, line 11: This will need some subclass defined here in gtk so that builder can automatically connect signals/etc. Best to create an issue for that.

src/auto/custom_sorter.rs
```diff
@@ -0,0 +1,41 @@
...
+glib_wrapper! {
+    pub struct CustomSorter(Object<gtk_sys::GtkCustomSorter, gtk_sys::GtkCustomSorterClass, CustomSorterClass>) @extends Sorter;
+
+    match fn {
+        get_type => || gtk_sys::gtk_custom_sorter_get_type(),
+    }
+}
+
+impl CustomSorter {
+    //pub fn new(sort_func: /*Unimplemented*/Fn(/*Unimplemented*/Option<Fundamental: Pointer>, /*Unimplemented*/Option<Fundamental: Pointer>) -> i32, user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> CustomSorter {
```
sdroege, line 20: This would be nice to implement, just like `gtk_custom_filter_new()` and `gtk_custom_layout_new()` but they probably all need manual code.

Note that the `gpointer` item parameters to these functions are always at least GObject subclasses. That needs some `type GLib.Object` annotations.

src/auto/cclosure_expression.rs
```diff
@@ -0,0 +1,28 @@
...
+// from gir-files (https://github.com/gtk-rs/gir-files)
+// DO NOT EDIT
+
+use glib::translate::*;
+use gtk_sys;
+use std::fmt;
+use Expression;
+
+glib_wrapper! {
+    pub struct CClosureExpression(Object<gtk_sys::GtkCClosureExpression, CClosureExpressionClass>) @extends Expression;
```
sdroege, line 11: I think this shouldn't be generated. AFAIU this is for wrapping C function pointers? `GtkClosureExpression` would be the one for us here :)

Gir.toml
bilelmoussaoui: Yes, the make regen_check test takes care of ensuring that (it's run in the CI now)

src/auto/constant_expression.rs
```diff
@@ -0,0 +1,48 @@
...
+        get_type => || gtk_sys::gtk_constant_expression_get_type(),
+    }
+}
+
+impl ConstantExpression {
+    //pub fn new(value_type: glib::types::Type, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs) -> ConstantExpression {
+    //    unsafe { TODO: call gtk_sys:gtk_constant_expression_new() }
+    //}
+
+    pub fn new_for_value(value: &glib::Value) -> ConstantExpression {
```
sdroege, line 25: This should probably get some addition in gir (just create an issue for now). Similar to `new_with_XXX()`, this should here become simply `for_value()` without the `new_` prefix.

src/auto/param_spec_expression.rs
```diff
@@ -0,0 +1,24 @@
...
+// from gir-files (https://github.com/gtk-rs/gir-files)
+// DO NOT EDIT
+
+use glib::translate::*;
+use gobject;
+use gtk_sys;
+use std::fmt;
+
+glib_wrapper! {
+    pub struct ParamSpecExpression(Object<gtk_sys::GtkParamSpecExpression, ParamSpecExpressionClass>) @extends gobject::ParamSpec;
```
sdroege: This does not work like this and needs to be manual.

See https://github.com/gtk-rs/glib/blob/master/src/param_spec.rs for details. This also needs a `new()` function (`gtk_param_spec_expression()`) and inside GLib I should clean up things so that there *are* `new()` functions...

Gir.toml
```diff
@@ -87,6 +87,14 @@ generate = [
     "Gtk.EventSequenceState",
     "Gtk.Expression",
```
sdroege, line 86: This is not a `GObject` but a separate fundamental type... so all these have to be manually bound. `glib::ParamSpec` is a good other example for such a thing.

Note that the `glib::Value` related functions for this are `gtk_value_XXX_expression()`.

All the expressions have to be handled specially.

src/auto/string_object.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub fn new(string: &str) -> StringObject {
+        assert_initialized_main_thread!();
+        unsafe { from_glib_full(gtk_sys::gtk_string_object_new(string.to_glib_none().0)) }
+    }
+}
+
+pub const NONE_STRING_OBJECT: Option<&StringObject> = None;
+
+pub trait StringObjectExt: 'static {
+    fn get_string(&self) -> Option<GString>;
```
sdroege, line 35: Not nullable

src/auto/tree_expander.rs
```diff
@@ -0,0 +1,178 @@
...
+        Self::new()
+    }
+}
+
+pub const NONE_TREE_EXPANDER: Option<&TreeExpander> = None;
+
+pub trait TreeExpanderExt: 'static {
+    fn get_child(&self) -> Option<Widget>;
+
+    //fn get_item(&self) -> /*Unimplemented*/Option<Fundamental: Pointer>;
```
sdroege, line 361: Needs a `type GLib.Object` annotation and then it would be generated

src/auto/builder_cscope.rs
```diff
@@ -0,0 +1,60 @@
...
+// DO NOT EDIT
+
+use glib::object::IsA;
+use glib::translate::*;
+use gtk_sys;
+use std::fmt;
+use BuilderScope;
+
+glib_wrapper! {
+    pub struct BuilderCScope(Object<gtk_sys::GtkBuilderCScope, gtk_sys::GtkBuilderCScopeClass, BuilderCScopeClass>) @implements BuilderScope;
```
sdroege: This should probably not be bound. It's for resolving via C symbols, that doesn't seem very useful in Rust and... unsafe

sdroege:
Otherwise generally looks good to me, but of course I didn't carefully review every single piece of API. I'm sure there are various wrong annotations that we'll find over time ;)

src/auto/param_spec_expression.rs
sdroege: Actually nevermind, it shouldn't have a `new()` function but basically work like https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/master/gstreamer/src/param_spec.rs#L13 . That way it will work well with the subclass property API.

src/auto/builder_cscope.rs
bilelmoussaoui: It's not bound neither, I have ended up disabling it later

src/auto/constant_expression.rs
bilelmoussaoui: Done https://github.com/gtk-rs/gir/issues/974

src/auto/param_spec_expression.rs
bilelmoussaoui: I have disabled all the Gtk.*Expression types until they are handled properly

bilelmoussaoui: Using Expressions from source code has been super weird so far, but thankfully you can use those from the UI file as well.

bilelmoussaoui: https://github.com/gtk-rs/gtk4/issues/21

Gir.toml
bilelmoussaoui: See https://github.com/gtk-rs/gtk4/issues/21

src/auto/custom_sorter.rs
bilelmoussaoui: I have opened issues to track both `CustomSorter` & `CustomLayout` #22 #23

src/auto/tree_expander.rs
bilelmoussaoui: Added to the gtk's broken annotations issue https://gitlab.gnome.org/GNOME/gtk/-/issues/3281

src/auto/builder_scope.rs
bilelmoussaoui: Done https://github.com/gtk-rs/gtk4/issues/24

src/auto/activate_action.rs
bilelmoussaoui: Reported upstream as part of the broken annotations issue 
https://gitlab.gnome.org/GNOME/gtk/-/issues/3281

sdroege:
Thanks, looks good for now. Let's get this in and then all further changes on top of this :)

bilelmoussaoui:
Thanks :tada:

PR #26, opened 2020-10-23
Widget subclass

kavanmevada:


gtk4/src/subclass/widget.rs
```diff
@@ -102,24 +518,335 @@ unsafe extern "C" fn widget_compute_expand<T: ObjectSubclass>(
...
+    ptr: *mut gtk_sys::GtkWidget,
+    direction_type_ptr: gtk_sys::GtkDirectionType,
+) -> glib_sys::gboolean
+where
+    T: WidgetImpl,
+{
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap = from_glib_borrow(ptr);
+    let direction_type = from_glib(direction_type_ptr);
```
bilelmoussaoui, line 568: there's a bunch of copy pastes here that needs to be updated

gtk4/src/subclass/widget.rs
```diff
@@ -102,24 +518,335 @@ unsafe extern "C" fn widget_compute_expand<T: ObjectSubclass>(
...
+
+unsafe extern "C" fn widget_get_request_mode<T: ObjectSubclass>(
+    ptr: *mut gtk_sys::GtkWidget,
+) -> gtk_sys::GtkSizeRequestMode
+where
+    T: WidgetImpl,
+{
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap = from_glib_borrow(ptr);
```
bilelmoussaoui, line 581: same here and for other overrides, It doesn't make sense to name this wrap :)

sdroege: How would you call it? That's how it is everywhere else :)

bilelmoussaoui: Ah no my bad. I confused the wrap function with the wrap property

bilelmoussaoui:
Can you add an examples to examples with a very simple widget subclass?

sdroege:
This needs to be updated for https://github.com/gtk-rs/gtk-rs/pull/27

bilelmoussaoui:
I went ahead and updated this one, I'm not very sure about the functions returns a bool, I wonder if they should be an `Inhibit` instead. Would appreciate a re-review here @sdroege

gtk4/src/subclass/widget.rs
```diff
@@ -1,21 +1,132 @@
 use gtk_sys;
 
+use libc::c_int;
```
sdroege, line 3: Why not use `u32` instead?

gtk4/src/subclass/widget.rs
```diff
@@ -53,19 +198,317 @@ impl<T: WidgetImpl> WidgetImplExt for T {
...
+
+    fn parent_focus(&self, widget: &Self::Type, direction_type: DirectionType) -> bool {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkWidgetClass;
+            if let Some(f) = (*parent_class).focus {
+                f(
+                    widget.unsafe_cast_ref::<Widget>().to_glib_none().0,
+                    direction_type.to_glib(),
+                ) != 0
```
sdroege, line 234: `from_glib()` instead of `!= 0`

sdroege: Also elsewhere

gtk4/src/subclass/widget.rs
```diff
@@ -53,19 +198,317 @@ impl<T: WidgetImpl> WidgetImplExt for T {
...
+                    orientation.to_glib(),
+                    for_size,
+                    minimum_size.as_mut_ptr(),
+                    natural_size.as_mut_ptr(),
+                    minimum_base_size.as_mut_ptr(),
+                    natural_base_size.as_mut_ptr(),
+                );
+            }
+            (
+                minimum_size.assume_init(),
```
sdroege, line 325: This causes UB if `parent_class.measure == NULL`. You need to implement some kind of default handling here (check what the C code does), or panic (if the C code would just crash / assert).

sdroege:
> I went ahead and updated this one, I'm not very sure about the functions that returns a bool, I wonder if they should return an `Inhibit` instead. Would appreciate a re-review here @sdroege

Those `bool` returns all seem to have other meanings. `Inhibit` is for signal handlers to continue/stop calling any further signal handlers.

gtk4/src/subclass/widget.rs
bilelmoussaoui: because the parent implementation requires a pointer

gtk4/src/subclass/widget.rs
bilelmoussaoui: There's always a measure implementation (that returns 0, 0, -1, -1) I just missed this

gtk4/src/subclass/widget.rs
sdroege: So? `*mut i32` is the same as `*mut libc::c_int` for all platforms that glib supports

gtk4/src/subclass/widget.rs
sdroege: Not if you're deriving from a subclass that has set it to `NULL` explicitly :)

bilelmoussaoui: So I asked ebassi & mclassen about this and a widget that sets the measure to null without setting a layout-manager will just make thing break and it's safe to assume to be always there

sdroege: Yes it's wrong, but you don't want to run into UB just because someone else does something wrong if you can avoid it. Just panic here or initialize the values to zeroes or so :)

sdroege: Ok, this panics now :)

PR #31, opened 2020-11-03
Update to current gir version

sophie-h:


bilelmoussaoui (approved):

bilelmoussaoui:
I was going to suggest updating the gir-files as well as they were updated for 3.99.3. But I guess that can happen later as we need to figure out the sys crate situation for gtk4 crates first.

src/auto/functions.rs
```diff
@@ -198,21 +193,15 @@ pub fn print_run_page_setup_dialog<P: IsA<Window>>(
     }
 }
 
-pub fn print_run_page_setup_dialog_async<
-    P: IsA<Window>,
-    Q: FnOnce(&PageSetup) + Send + Sync + 'static,
```
sdroege, line 203: Hrm, this looks wrong

sdroege: See https://github.com/gtk-rs/gir/pull/981

sdroege: If you update gir another time, it should be all good now :)

sophie-h: Done

sdroege (approved):

bilelmoussaoui:
Thanks!

PR #35, opened 2020-11-04
Merge repos

bilelmoussaoui:
To be done: 
- [x] Add the geneartor  from gtk-rs
- [x] Add a readme
- [x] Fix the CI


bilelmoussaoui:
The build fails currently locally with 
```
error: failed to select a version for `gio-sys`.
    ... required by package `gdk4-sys v0.1.0 (/home/bilelmoussaoui/Projects/gtk4/gdk4/sys)`
    ... which is depended on by `gdk4 v0.1.0 (/home/bilelmoussaoui/Projects/gtk4/gdk4)`
versions that meet the requirements `*` are: 0.13.0

the package `gio-sys` links to the native library `gio-2.0`, but it conflicts with a previous package which links to `gio-2.0` as well:
package `gio-sys v0.10.0 (https://github.com/gtk-rs/sys#5f35e26c)`
    ... which is depended on by `gdk4 v0.1.0 (/home/bilelmoussaoui/Projects/gtk4/gdk4)`
```
I guess the link is a bit problematic :/

sdroege:
> I guess the link is a bit problematic :/

You need to update all the dependencies to be inside the new repo (by path instead of via git!), and also point to the combined gtk-rs repo for the other deps.

bilelmoussaoui:
> > I guess the link is a bit problematic :/
> 
> You need to update all the dependencies to be inside the new repo (by path instead of via git!), and also point to the combined gtk-rs repo for the other deps.

My bad, that worked just fine

bilelmoussaoui:
This doesn't contain the examples4 repository as it's currently broken and will just break the CI.

.github/workflows/CI.yml
```diff
@@ -64,13 +59,11 @@ jobs:
         with:
           command: fmt
           args: --all -- --check
-      - run: make regen_check
```
sdroege, line 67: Why remove this one?

sdroege:
Looks good to me otherwise!

sdroege (approved):

.github/workflows/CI.yml
bilelmoussaoui: We don't use the make script any more and the generator doesn't have a assume yes flag yet. Will have to fix that first, same with gtk-rs repository

sdroege: Ok, please create an issue for that and then merge this here :)

bilelmoussaoui: Done https://github.com/gtk-rs/gtk4/issues/39

bilelmoussaoui:
Thanks for the review as always! :+1:

PR #41, opened 2020-11-05
regenerate with gtk 3.99.4

bilelmoussaoui:


gsk4/src/auto/render_node.rs
```diff
@@ -21,7 +21,7 @@ glib_wrapper! {
 }
 
 impl RenderNode {
-    //pub fn deserialize(bytes: &glib::Bytes, error_func: /*Unimplemented*/FnMut(/*Unimplemented*/Fundamental: Pointer, &glib::Error), user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> Option<RenderNode> {
+    //pub fn deserialize(bytes: &glib::Bytes, error_func: /*Unimplemented*/FnMut(/*Ignored*/gtk::CssSection, &glib::Error), user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> Option<RenderNode> {
```
sdroege, line 24: This causes a circular dependency between gtk4 and gsk4, that's rather suboptimal.

gsk4/sys/src/lib.rs
```diff
@@ -109,7 +109,7 @@ pub const GSK_TRANSFORM_CATEGORY_IDENTITY: GskTransformCategory = 6;
 
 // Callbacks
 pub type GskParseErrorFunc =
-    Option<unsafe extern "C" fn(gconstpointer, *const glib::GError, gpointer)>;
+    Option<unsafe extern "C" fn(*const gtk::GtkCssSection, *const glib::GError, gpointer)>;
```
sdroege, line 112: Does that compile? :)

sdroege:
Looks good to me otherwise

gsk4/sys/src/lib.rs
bilelmoussaoui: I don't know for now, waiting on the gtk4 docker image to get updated

sdroege (approved):

PR #42, opened 2020-11-05
gtk4: update subclasses to the latest glib

bilelmoussaoui:
cc @sdroege


gtk4/src/subclass/application.rs
```diff
@@ -19,12 +18,12 @@ pub trait GtkApplicationImpl:
     }
 }
 
-pub trait GtkApplicationImplExt {
+pub trait ApplicationImplExt {
```
sdroege, line 21: Why did you rename the trait? This is annoying now because `gio::ApplicationImplExt` and `gtk::ApplicationImplExt` :)

gtk4/src/subclass/application.rs
```diff
@@ -48,47 +47,34 @@ impl<T: GtkApplicationImpl> GtkApplicationImplExt for T {
...
-        unsafe extern "C" fn application_window_added<T: GtkApplicationImpl>(
-            ptr: *mut gtk_sys::GtkApplication,
-            wptr: *mut gtk_sys::GtkWindow,
-        ) {
-            let instance = &*(ptr as *mut T::Instance);
-            let imp = instance.get_impl();
-            let wrap: Borrowed<Application> = from_glib_borrow(ptr);
+unsafe impl<T: ApplicationImpl> IsSubclassable<T> for Application {
+    fn override_vfuncs(class: &mut ::glib::Class<Self>) {
+        <glib::Object as IsSubclassable<T>>::override_vfuncs(class);
```
sdroege, line 52:
```suggestion
        <gio::Application as IsSubclassable<T>>::override_vfuncs(class);
```

:scream:

bilelmoussaoui: Agh, my bad.

gtk4/src/subclass/application.rs
bilelmoussaoui: I've followed what's done in gtk3 here

bilelmoussaoui: Ah no. I don't know what I was doing here. Fixed!

sdroege (approved):

gtk4/src/subclass/application.rs
sdroege: Note that there's also some more magic happening there in gtk3, which I guess we'll include as port of going through all commits there and backporting as needed :)

PR #43, opened 2020-11-06
Various fixes

bilelmoussaoui:
- Sync generator
- Add regen check job to the CI
- Fix gtk-rs checker job not working
- Add a docs job to the CI
- Fix clippy errors
- Add a clippy job


gdk4/src/cairo_interaction.rs
```diff
@@ -45,7 +45,7 @@ pub trait GdkCairoContextExt {
 }
 
 impl GdkCairoContextExt for Context {
-    unsafe fn draw_from_gl(
+    fn draw_from_gl(
```
sdroege, line 48: This has to be unsafe. The `source` integer is a GL resource and if you pass random integers in there you'll have a lot of fun

sdroege:
Looks good otherwise

gdk4/src/cairo_interaction.rs
bilelmoussaoui: Aha. I've added a comment about the safety of the function. Not sure if the phrasing could be improved though

PR #44, opened 2020-11-07
CI: fix docs job

bilelmoussaoui:


sdroege:
Seems fine to me, not sure why the other doesn't work but github actions are even more strange than gitlab CI ;)

bilelmoussaoui:
It was not running because gir expected a Gir.toml file in the root tree instead of having multiple ones in subdirectories

bilelmoussaoui:
Seems to be running fine. Do you mind enabling Github Pages on this repository so that the docs are accessible to the users? Inside this repository's settings -> Github Pages -> then selecting `gh-pages` as the default branch

sdroege:
Done, where would the docs be then?

sdroege:
https://gtk-rs.org/gtk4-rs/ it says but that doesn't work :)

bilelmoussaoui:
https://gtk-rs.org/gtk4-rs/gtk4/, somehow cargo doc doesn't generate a main page when it generates docs for multiple packages in a workspace

sdroege:
@GuillaumeGomez ?

sdroege:
@bilelmoussaoui But it could be linked to already at least, and the crates cross-link each other. But only between those, you don't get e.g. GLib docs.

For gstreamer I made it so that all the dependencies are also documented because of that: https://gstreamer.pages.freedesktop.org/gstreamer-rs/gstreamer/

sdroege:
Should probably copy&paste all this to the gtk-rs CI too once it all works as expected :)

bilelmoussaoui:
> @bilelmoussaoui But it could be linked to already at least, and the crates cross-link each other. But only between those, you don't get e.g. GLib docs.
> 
> For gstreamer I made it so that all the dependencies are also documented because of that: https://gstreamer.pages.freedesktop.org/gstreamer-rs/gstreamer/

We can enable just the docs of the used dependencies from glib/gtk's ecosystem to avoid having a bunch of unrelated crates on the docs ?

> Should probably copy&paste all this to the gtk-rs CI too once it all works as expected :)

For sure

sdroege:
> We can enable just the docs of the used dependencies from glib/gtk's ecosystem to avoid having a bunch of unrelated crates on the docs ?

Ah that's possible? Sounds good :) Having only those crates enabled that appear in the public API of our crates would be useful.

GuillaumeGomez:
You can't really, but what you can do is remove those crates from the list like we do with gtk-rs. But that requires some tooling.

Issue #37, opened 2020-11-04
Generate gdk4-x11 & gdk4-wayland


bilelmoussaoui:
Requires adding the proper gir files to gir-files first 
- [x] gdk4-x11
- [ ] gdk4-wayland

PR #45, opened 2020-11-07
Generate gdk4-x11

bilelmoussaoui:
Part of #37


.github/workflows/docs.yml
```diff
@@ -24,14 +24,16 @@ jobs:
...
       - run: cargo install rustdoc-stripper
       - run: rustdoc-stripper -g -o ./gdk4/docs.md
       - run: rustdoc-stripper -g -o ./gsk4/docs.md
       - run: rustdoc-stripper -g -o ./gtk4/docs.md
+      - run: rustdoc-stripper -g -o ./gdk4-x11/docs.md
       - uses: actions-rs/cargo@v1
         with:
           command: doc
-          args: -p gtk4 -p gtk4-sys -p gsk4 -p gsk4-sys -p gdk4 -p gdk4-sys --no-deps
+          args: --workspace --no-deps
```
sdroege, line 36: Is `--workspace` the same as `--all`?

bilelmoussaoui: Yes, --all seems to be deprecated

sdroege:
Looks good to me otherwise :)

sdroege:
I.e., merge when you're happy with it and the CI is too

.github/workflows/docs.yml
sdroege: I should update the GStreamer CI then, thanks!

bilelmoussaoui:
Thanks for the quick review =) 

The CI seems to be happy & it looks fine for an initial thing. I will try to add the gdk4-wayland one as well.

PR #51, opened 2020-11-15
Add initial support for GTK composite templates

BrainBlasted:
Adds support for composite templates in gtk4-rs.
For now this only adds support for objects as template
children. We will need to figure out signals/callbacks later.

### DIFF
--- a/examples/src/bin/composite_template.rs
+++ b/examples/src/bin/composite_template.rs
@@ -2,15 +2,10 @@
 //!
 //! This sample demonstrates how to create a widget using GTK's composite templates.
 
-#[macro_use]
-extern crate glib;
-#[macro_use]
-extern crate gtk;
-
 use gio::{prelude::ApplicationExtManual, ApplicationExt};
 use glib::subclass::prelude::*;
-use glib::{Cast, StaticType};
-use gtk::WidgetExt;
+use glib::{glib_object_subclass, glib_wrapper, Cast, StaticType};
+use gtk::{CompositeTemplate, WidgetExt};
 
 mod imp {
     use super::*;
--- a/examples/src/bin/composite_template.rs
+++ b/examples/src/bin/composite_template.rs
@@ -8,68 +8,74 @@ extern crate glib;
 extern crate gtk;
 
 use gio::{prelude::ApplicationExtManual, ApplicationExt};
-use glib::subclass;
 use glib::subclass::prelude::*;
 use glib::{Cast, StaticType};
-use gtk::subclass::prelude::*;
-use gtk::subclass::widget::*;
 use gtk::WidgetExt;
 
-/// The private struct, which can hold widgets and other data.
-#[derive(Debug, CompositeTemplate)]
-pub struct ExApplicationWindowPriv {
-    // The #[template_child] attribute tells the CompositeTemplate macro
-    // that a field is meant to be a child within the template.
-    #[template_child(id = "headerbar")]
-    headerbar: TemplateChild<gtk::HeaderBar>,
-    #[template_child(id = "label")]
-    label: TemplateChild<gtk::Label>,
-    #[template_child(id = "subtitle_label")]
-    subtitle: TemplateChild<gtk::Label>,
-}
+mod imp {
+    use super::*;
+    use glib::subclass;
+    use gtk::subclass::prelude::*;
+    use gtk::subclass::widget::*;
+    use gtk::WidgetExt;
 
-impl ObjectSubclass for ExApplicationWindowPriv {
-    const NAME: &'static str = "ExApplicationWindow";
-    type Type = ExApplicationWindow;
-    type ParentType = gtk::ApplicationWindow;
-    type Instance = subclass::simple::InstanceStruct<Self>;
-    type Class = subclass::simple::ClassStruct<Self>;
+    /// The private struct, which can hold widgets and other data.
+    #[derive(Debug, CompositeTemplate)]
+    pub struct ExApplicationWindow {
+        // The #[template_child] attribute tells the CompositeTemplate macro
+        // that a field is meant to be a child within the template.
+        #[template_child(id = "headerbar")]
+        pub headerbar: TemplateChild<gtk::HeaderBar>,
+        #[template_child(id = "label")]
+        pub label: TemplateChild<gtk::Label>,
+        #[template_child(id = "subtitle_label")]
+        pub subtitle: TemplateChild<gtk::Label>,
+    }
 
-    glib_object_subclass!();
+    impl ObjectSubclass for ExApplicationWindow {
+        const NAME: &'static str = "ExApplicationWindow";
+        type Type = super::ExApplicationWindow;
+        type ParentType = gtk::ApplicationWindow;
+        type Instance = subclass::simple::InstanceStruct<Self>;
+        type Class = subclass::simple::ClassStruct<Self>;
 
-    fn new() -> Self {
-        Self {
-            headerbar: TemplateChild::default(),
-            label: TemplateChild::default(),
-            subtitle: TemplateChild::default(),
+        glib_object_subclass!();
+
+        fn new() -> Self {
+            Self {
+                headerbar: TemplateChild::default(),
+                label: TemplateChild::default(),
+                subtitle: TemplateChild::default(),
+            }
+        }
+
+        // Within class_init() you must set the template
+        // and bind it's children. The CompositeTemplate
+        // derive macro provides a convenience function
+        // bind_template_children() to bind all children
+        // at once.
+        fn class_init(klass: &mut Self::Class) {
+            let template = include_bytes!("composite_template.ui");
+            klass.set_template(template);
+            Self::bind_template_children(klass);
         }
     }
 
-    // Within class_init() you must set the template
-    // and bind it's children. The CompositeTemplate
-    // derive macro provides a convenience function
-    // bind_template_children() to bind all children
-    // at once.
-    fn class_init(klass: &mut Self::Class) {
-        let template = include_bytes!("composite_template.ui");
-        klass.set_template(template);
-        Self::bind_template_children(klass);
+    impl ObjectImpl for ExApplicationWindow {
+        fn constructed(&self, obj: &Self::Type) {
+            obj.init_template();
+            obj.init_label();
+            self.parent_constructed(obj);
+        }
     }
-}
 
-impl ObjectImpl for ExApplicationWindowPriv {
-    fn constructed(&self, obj: &Self::Type) {
-        obj.init_template();
-        obj.init_label();
-        self.parent_constructed(obj);
-    }
+    impl WidgetImpl for ExApplicationWindow {}
+    impl WindowImpl for ExApplicationWindow {}
+    impl ApplicationWindowImpl for ExApplicationWindow {}
 }
-impl WidgetImpl for ExApplicationWindowPriv {}
-impl WindowImpl for ExApplicationWindowPriv {}
-impl ApplicationWindowImpl for ExApplicationWindowPriv {}
 
 glib_wrapper! {
-    pub struct ExApplicationWindow(ObjectSubclass<ExApplicationWindowPriv>)
+    pub struct ExApplicationWindow(ObjectSubclass<imp::ExApplicationWindow>)
         @extends gtk::Widget, gtk::Window, gtk::ApplicationWindow, @implements gio::ActionMap, gio::ActionGroup;
 }
 
@@ -84,7 +90,7 @@ impl ExApplicationWindow {
     pub fn init_label(&self) {
         // To access fields such as template children, you must get
         // the private struct.
-        let self_ = ExApplicationWindowPriv::from_instance(self);
+        let self_ = imp::ExApplicationWindow::from_instance(self);
         self_
             .subtitle
             .get()
--- a/examples/src/bin/composite_template.rs
+++ b/examples/src/bin/composite_template.rs
@@ -57,7 +57,13 @@ impl ObjectSubclass for ExApplicationWindowPriv {
     }
 }
 
-impl ObjectImpl for ExApplicationWindowPriv {}
+impl ObjectImpl for ExApplicationWindowPriv {
+    fn constructed(&self, obj: &Self::Type) {
+        obj.init_template();
+        obj.init_label();
+        self.parent_constructed(obj);
+    }
+}
 impl WidgetImpl for ExApplicationWindowPriv {}
 impl WindowImpl for ExApplicationWindowPriv {}
 impl ApplicationWindowImpl for ExApplicationWindowPriv {}
@@ -69,14 +75,10 @@ glib_wrapper! {
 
 impl ExApplicationWindow {
     pub fn new<P: glib::IsA<gtk::Application> + glib::value::ToValue>(app: &P) -> Self {
-        let win = glib::Object::new(Self::static_type(), &[("application", app)])
+        glib::Object::new(Self::static_type(), &[("application", app)])
             .expect("Failed to create ExApplicationWindow")
             .downcast::<ExApplicationWindow>()
-            .expect("Created object is of wrong type");
-        // init_template() must be run after the object is constructed.
-        win.init_template();
-        win.init_label();
-        win
+            .expect("Created object is of wrong type")
     }
 
     pub fn init_label(&self) {
--- a/examples/src/bin/composite_template.rs
+++ b/examples/src/bin/composite_template.rs
@@ -52,7 +52,7 @@ impl ObjectSubclass for ExApplicationWindowPriv {
     // at once.
     fn class_init(klass: &mut Self::Class) {
         let template = include_bytes!("composite_template.ui");
-        Self::set_template(klass, template);
+        klass.set_template(template);
         Self::bind_template_children(klass);
     }
 }
--- a/examples/src/bin/composite_template.rs
+++ b/examples/src/bin/composite_template.rs
@@ -1,0 +1,106 @@
+//! # Composite Template Example
+//!
+//! This sample demonstrates how to create a widget using GTK's composite templates.
+
+#[macro_use]
+extern crate glib;
+#[macro_use]
+extern crate gtk;
+
+use gio::{prelude::ApplicationExtManual, ApplicationExt};
+use glib::subclass;
+use glib::subclass::prelude::*;
+use glib::{Cast, StaticType};
+use gtk::subclass::prelude::*;
+use gtk::subclass::widget::*;
+use gtk::WidgetExt;
+
+/// The private struct, which can hold widgets and other data.
+#[derive(Debug, CompositeTemplate)]
+pub struct ExApplicationWindowPriv {
+    // The #[template_child] attribute tells the CompositeTemplate macro
+    // that a field is meant to be a child within the template.
+    #[template_child(id = "headerbar")]
+    headerbar: TemplateChild<gtk::HeaderBar>,
+    #[template_child(id = "label")]
+    label: TemplateChild<gtk::Label>,
+    #[template_child(id = "subtitle_label")]
+    subtitle: TemplateChild<gtk::Label>,
+}
+
+impl ObjectSubclass for ExApplicationWindowPriv {
+    const NAME: &'static str = "ExApplicationWindow";
+    type Type = ExApplicationWindow;
+    type ParentType = gtk::ApplicationWindow;
+    type Instance = subclass::simple::InstanceStruct<Self>;
+    type Class = subclass::simple::ClassStruct<Self>;
+
+    glib_object_subclass!();
+
+    fn new() -> Self {
+        Self {
+            headerbar: TemplateChild::default(),
+            label: TemplateChild::default(),
+            subtitle: TemplateChild::default(),
+        }
+    }
+
+    // Within class_init() you must set the template
+    // and bind it's children. The CompositeTemplate
+    // derive macro provides a convenience function
+    // bind_template_children() to bind all children
+    // at once.
+    fn class_init(klass: &mut Self::Class) {
+        let template = include_bytes!("composite_template.ui");
+        Self::set_template(klass, template);
+        Self::bind_template_children(klass);
+    }
+}
+
+impl ObjectImpl for ExApplicationWindowPriv {}
+impl WidgetImpl for ExApplicationWindowPriv {}
+impl WindowImpl for ExApplicationWindowPriv {}
+impl ApplicationWindowImpl for ExApplicationWindowPriv {}
+
+glib_wrapper! {
+    pub struct ExApplicationWindow(ObjectSubclass<ExApplicationWindowPriv>)
+        @extends gtk::Widget, gtk::Window, gtk::ApplicationWindow, @implements gio::ActionMap, gio::ActionGroup;
+}
+
+impl ExApplicationWindow {
+    pub fn new<P: glib::IsA<gtk::Application> + glib::value::ToValue>(app: &P) -> Self {
+        let win = glib::Object::new(Self::static_type(), &[("application", app)])
+            .expect("Failed to create ExApplicationWindow")
+            .downcast::<ExApplicationWindow>()
+            .expect("Created object is of wrong type");
+        // init_template() must be run after the object is constructed.
+        win.init_template();
+        win.init_label();
+        win
+    }
+
+    pub fn init_label(&self) {
+        // To access fields such as template children, you must get
+        // the private struct.
+        let self_ = ExApplicationWindowPriv::from_instance(self);
+        self_
+            .subtitle
+            .get()
+            .set_text("This is an example window made using composite templates");
+    }
+}
+
+fn main() {
+    let application = gtk::Application::new(
+        Some("com.github.gtk-rs.examples.composite_template"),
+        Default::default(),
+    )
+    .expect("Failed to initialize application");
+
+    application.connect_activate(|app| {
+        let win = ExApplicationWindow::new(app);
+        win.show();
+    });
+
+    application.run(&std::env::args().collect::<Vec<_>>());
+}
--- a/examples/src/bin/composite_template.rs
+++ b/examples/src/bin/composite_template.rs
@@ -2,17 +2,16 @@
 //!
 //! This sample demonstrates how to create a widget using GTK's composite templates.
 
-use gio::{prelude::ApplicationExtManual, ApplicationExt};
+use gio::prelude::*;
 use glib::subclass::prelude::*;
-use glib::{glib_object_subclass, glib_wrapper, Cast, StaticType};
-use gtk::{CompositeTemplate, WidgetExt};
+use glib::{glib_object_subclass, glib_wrapper};
+use gtk::{prelude::*, CompositeTemplate};
 
 mod imp {
     use super::*;
     use glib::subclass;
     use gtk::subclass::prelude::*;
     use gtk::subclass::widget::*;
-    use gtk::WidgetExt;
 
     /// The private struct, which can hold widgets and other data.
     #[derive(Debug, CompositeTemplate)]
--- a/examples/src/bin/composite_template.rs
+++ b/examples/src/bin/composite_template.rs
@@ -74,7 +74,7 @@ glib_wrapper! {
 }
 
 impl ExApplicationWindow {
-    pub fn new<P: glib::IsA<gtk::Application> + glib::value::ToValue>(app: &P) -> Self {
+    pub fn new<P: glib::IsA<gtk::Application>>(app: &P) -> Self {
         glib::Object::new(Self::static_type(), &[("application", app)])
             .expect("Failed to create ExApplicationWindow")
             .downcast::<ExApplicationWindow>()
--- a/gtk4-macros/Cargo.toml
+++ b/gtk4-macros/Cargo.toml
@@ -1,0 +1,28 @@
+[package]
+documentation = "https://gtk-rs.org/docs/gtk4-macros/"
+categories = ["api-bindings", "gui"]
+license = "MIT"
+description = "Rust bindings for the GTK 4 library"
+homepage = "https://gtk-rs.org/"
+name = "gtk4-macros"
+version = "0.1.0"
+authors = ["The Gtk-rs Project Developers"]
+edition = "2018"
+keywords = ["gtk", "gtk4", "gtk-rs", "gnome", "GUI"]
+repository = "https://github.com/gtk-rs/gtk4-rs"
+exclude = [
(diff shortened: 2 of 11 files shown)


gtk4/src/subclass/widget.rs
```diff
@@ -123,3 +126,127 @@ unsafe extern "C" fn widget_direction_changed<T: ObjectSubclass>(
 
     imp.direction_changed(&wrap, dirwrap)
 }
+
+pub trait WidgetSubclass
+where
+    Self: ObjectSubclass + WidgetImpl,
+{
+    unsafe fn set_template_bytes(
```
BrainBlasted, line 134: API question: should these functions be `unsafe` for the developers, or should they be "safe" and use `unsafe` internally like other bindings?

gtk4/src/subclass/widget.rs
```diff
@@ -123,3 +126,127 @@ unsafe extern "C" fn widget_direction_changed<T: ObjectSubclass>(
...
+            name.to_glib_none().0,
+            false as glib_sys::gboolean,
+            private_offset + (offset.get_byte_offset() as isize),
+        )
+    }
+}
+
+#[derive(Debug)]
+#[repr(transparent)]
+pub struct TemplateWidget<T>
```
BrainBlasted, line 208: This may work for more than widgets - perhaps `TemplateChild` would be more appropriate.

gtk4/src/subclass/widget.rs
```diff
@@ -123,3 +126,126 @@ unsafe extern "C" fn widget_direction_changed<T: ObjectSubclass>(
 
     imp.direction_changed(&wrap, dirwrap)
 }
+
+pub trait WidgetSubclass
```
sdroege, line 130: The naming scheme is different :) The way how such things are named elsewhere would be `WidgetClassSubclassExt`

gtk4/src/subclass/widget.rs
sdroege: The only place where you can safely get a mutable reference to the class struct is in `class_init()`. Is it safe to use there in any case?

gtk4/src/subclass/widget.rs
sdroege: And this needs a blanket impl `unsafe impl WidgetClassSubclassExt for glib::Class<Widget> {}`

(And the crate itself should be unsafe, the functions probably not)

gtk4/src/subclass/widget.rs
```diff
@@ -123,3 +126,126 @@ unsafe extern "C" fn widget_direction_changed<T: ObjectSubclass>(
...
+        gtk_sys::gtk_widget_class_bind_template_child_full(
+            widget_class,
+            name.to_glib_none().0,
+            false as glib_sys::gboolean,
+            private_offset + (offset.get_byte_offset() as isize),
+        )
+    }
+}
+
+#[derive(Debug)]
```
sdroege, line 206: `PartialEq` and `Eq` could also be implemented, and `PartialEq<Widget> for TemplateWidget`

gtk4/src/subclass/widget.rs
sdroege: Can `gtk::Builder` be used for anything but widgets?

gtk4/src/subclass/widget.rs
```diff
@@ -123,3 +126,126 @@ unsafe extern "C" fn widget_direction_changed<T: ObjectSubclass>(
...
+        }
+    }
+}
+
+#[allow(unused)]
+impl<T> TemplateWidget<T>
+where
+    T: ObjectType + FromGlibPtrNone<*mut <T as ObjectType>::GlibType>,
+{
+    fn new() -> Self {
```
sdroege, line 231: Not really needed if you have `Default` :)

gtk4/src/lib.rs
```diff
@@ -17,6 +17,10 @@ extern crate libc;
 #[macro_use]
 extern crate bitflags;
 extern crate once_cell;
+#[doc(hidden)]
+#[allow(unused_imports)]
```
sdroege, line 21: ` #[allow(unused_imports)]` shouldn't be needed, maybe that's part of the problem :)

gtk4/src/lib.rs
```diff
@@ -17,6 +17,10 @@ extern crate libc;
...
+#[macro_use]
+pub extern crate field_offset;
```
sdroege, line 24: You also want to re-export the `gtk4-macros` macros from here, just like it's done in `glib` with `glib-macros`

gtk4/src/subclass/widget.rs
```diff
@@ -123,3 +126,126 @@ unsafe extern "C" fn widget_direction_changed<T: ObjectSubclass>(
...
+                as *mut gtk_sys::GtkWidgetClass;
+        gtk_sys::gtk_widget_class_bind_template_child_full(
+            widget_class,
+            name.to_glib_none().0,
+            false as glib_sys::gboolean,
+            0,
+        );
+    }
+
+    unsafe fn bind_template_child_with_offset<T>(
```
sdroege, line 185: This function should be unsafe

BrainBlasted:
That is where you are supposed to use them, according to the Gtk docsOn Nov 15, 2020 2:01 AM, Sebastian Dröge <notifications@github.com> wrote:
@sdroege commented on this pull request.



In gtk4/src/subclass/widget.rs:
> @@ -123,3 +126,127 @@ unsafe extern "C" fn widget_direction_changed<T: ObjectSubclass>(
 
     imp.direction_changed(&wrap, dirwrap)
 }
+
+pub trait WidgetSubclass
+where
+    Self: ObjectSubclass + WidgetImpl,
+{
+    unsafe fn set_template_bytes(

The only place where you can safely get a mutable reference to the class struct is in class_init(). Is it safe to use there in any case?

BrainBlasted:
Yes - for Menus, gestures, etc.On Nov 15, 2020 2:03 AM, Sebastian Dröge <notifications@github.com> wrote:
@sdroege commented on this pull request.



In gtk4/src/subclass/widget.rs:
> +        let widget_class =
+            gobject_sys::g_type_check_class_cast(type_class, gtk_sys::gtk_widget_get_type())
+                as *mut gtk_sys::GtkWidgetClass;
+        let private_offset = Self::type_data().as_ref().private_offset;
+        gtk_sys::gtk_widget_class_bind_template_child_full(
+            widget_class,
+            name.to_glib_none().0,
+            false as glib_sys::gboolean,
+            private_offset + (offset.get_byte_offset() as isize),
+        )
+    }
+}
+
+#[derive(Debug)]
+#[repr(transparent)]
+pub struct TemplateWidget<T>

Can gtk::Builder be used for anything but widgets?

gtk4-macros/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+categories = ["api-bindings", "gui"]
+license = "MIT"
+description = "Rust bindings for the GTK 4 library"
+homepage = "https://gtk-rs.org/"
+name = "gtk4-macros"
+version = "0.1.0"
+authors = ["The Gtk-rs Project Developers"]
+edition = "2018"
+keywords = ["gtk", "gtk4", "gtk-rs", "gnome", "GUI"]
+repository = "https://github.com/gtk-rs/gtk4"
```
bilelmoussaoui, line 12:
```suggestion
repository = "https://github.com/gtk-rs/gtk4-rs"
```

gtk4-macros/src/meson.build
```diff
@@ -0,0 +1,5 @@
+template_rust_sources = files(
```
bilelmoussaoui, line 1: This file can be removed

gtk4/src/subclass/widget.rs
bilelmoussaoui: Vala uses `[GtkChild]` annotation while Python uses `Gtk.Template.Child()`

sdroege:
> That is where you are supposed to use them, according to the Gtk docs

Well, then all functions except for the one with the offset can be safe :)

> Yes - for Menus, gestures, etc.

And they can be defined in templates? Then yes, widget -> child everywhere

sdroege:
Can you also include an example in here? Maybe start a new `examples4` subdirectory for that :)

bilelmoussaoui:
> Can you also include an example in here? Maybe start a new `examples4` subdirectory for that :)

I will add a very basic example separately today, then we can rebase this one & add a new template example

gtk4/src/lib.rs
BrainBlasted: I re-exported the macros and was able to get it to compile, but the warning persists unless I do this; Should I plug `#[allow(unused_imports)]` back in?

gtk4/src/subclass/widget.rs
BrainBlasted: > And this needs a blanket impl unsafe impl WidgetClassSubclassExt for glib::Class<Widget> {}

These functions should only be used in the context of subclassing during `class_init()`; Would that blanket impl apply it to every widget?

BrainBlasted: It doesn't work with the trait bounds ` ObjectSubclass + WidgetImpl`

BrainBlasted: Which in turn means we can't use `Self::Class` in `bind_template_children()`

sdroege: Check gstreamer-rs for various patterns for class methods that should only be used for subclassing :)

gtk4/src/lib.rs
sdroege: I guess so

gtk4/src/subclass/widget.rs
BrainBlasted: I've made a change so that it's `unsafe impl<T: ObjectSubclass + WidgetImpl> WidgetClassSubclassExt for T {}`. Is that okay?

gtk4/src/lib.rs
```diff
@@ -17,6 +17,17 @@ extern crate libc;
 #[macro_use]
 extern crate bitflags;
 extern crate once_cell;
+#[macro_use]
+#[allow(unused_imports)]
+pub extern crate field_offset;
```
sdroege, line 22: `#[doc(hidden)]` maybe

sdroege: Also below

gtk4/src/subclass/widget.rs
```diff
@@ -123,3 +126,127 @@ unsafe extern "C" fn widget_direction_changed<T: ObjectSubclass>(
...
+        unsafe {
+            Option::<T>::from_glib_none(self.ptr)
+                .expect("Failed to retrieve template child. Please check that it has been bound.")
+        }
+    }
+}
+
+pub trait CompositeTemplate
+where
+    Self: WidgetClassSubclassExt,
```
sdroege, line 249: Why not
```rust
pub trait CompositeTemplate: WidgetClassSubclassExt { ... }
```
?

Also that can just be `WidgetImpl` instead of the extension trait. Equivalent but probably slightly easier to resolve for the compiler.

gtk4/src/subclass/widget.rs
```diff
@@ -123,3 +126,127 @@ unsafe extern "C" fn widget_direction_changed<T: ObjectSubclass>(
 
     imp.direction_changed(&wrap, dirwrap)
 }
+
+pub unsafe trait WidgetClassSubclassExt
+where
+    Self: ObjectSubclass + WidgetImpl,
+{
+    fn set_template_bytes(
+        klass: &mut subclass::simple::ClassStruct<Self>,
```
sdroege, line 135:
```suggestion
        klass: &mut glib::Class<Self::Type>,
```

or

```suggestion
        klass: &mut glib::Class<<Self as ObjectSubclass>::Type>,
```

Here and in the other functions.

gtk4/src/subclass/widget.rs
sdroege: `unsafe impl WidgetClassSubclassExt for glib::Class<gtk::Widget> {}` did not work?

Another variant that I have is `unsafe impl<T: ClassStruct> WidgetClassSubclassExt for T where     T::Type: WidgetImpl`.

And then the variant I added in the new comment :)

But I think in the end these are all equivalent in practice. But it would be good to have an example of this, and especially one that makes sure this actually doesn't only work for direct subclasses of `gtk::Widget` but also e.g. `gtk::Box`.

gtk4/src/subclass/widget.rs
sdroege: No, actually this should be:

```suggestion
        klass: &mut <Self as ObjectSubclass>::Class,
```

gtk4/src/subclass/widget.rs
sdroege: And it would be good to have it implemented on the class struct in some way, so that inside `class_init()` you can just call `class.blabla()`.

Anyway, if you provide an example in the repo I can take a look at such things. Otherwise I can just give suggestions that might not work :)

gtk4/src/subclass/widget.rs
BrainBlasted: Can it? bind_template_children() in the macro relies on the functions from WidgetClassSubclassExt, so these bounds ensure that it's implemented.

sdroege: That should be something the compiler should be able to figure out, but I've to see it in practice :)

gtk4/src/subclass/widget.rs
```diff
@@ -106,3 +107,115 @@ unsafe extern "C" fn widget_direction_changed<T: WidgetImpl>(
...
+            gtk_sys::gtk_widget_class_bind_template_child_full(
+                widget_class,
+                name.to_glib_none().0,
+                false as glib_sys::gboolean,
+                0,
+            );
+        }
+    }
+
+    unsafe fn bind_template_child_with_offset<T>(
```
bilelmoussaoui, line 163:
```suggestion
    #[allow(clippy::missing_safety_doc)]
    unsafe fn bind_template_child_with_offset<T>(
```

gtk4/src/subclass/widget.rs
BrainBlasted: Added an example.

examples/src/bin/composite_template.rs
```diff
@@ -0,0 +1,106 @@
...
+use glib::subclass;
+use glib::subclass::prelude::*;
+use glib::{Cast, StaticType};
+use gtk::subclass::prelude::*;
+use gtk::subclass::widget::*;
+use gtk::WidgetExt;
+
+/// The private struct, which can hold widgets and other data.
+#[derive(Debug, CompositeTemplate)]
+pub struct ExApplicationWindowPriv {
```
sdroege, line 20: The pattern we use here and elsewhere (see docs) is

```rust
mod imp {
    pub struct ExApplicationWindow {
        [...]
    }

    impl ObjectSubclass for ExApplicationWindow {
        type Type = super::ExApplicationWindow;
    }
}

glib_wrapper! {
    pub struct ExApplicationWindow(ObjectSubclass<imp::ExApplicationWindow>)
        @extends gtk::Widget, gtk::Window, gtk::ApplicationWindow, @implements gio::ActionMap, gio::ActionGroup;
}
```

We have proper namespacing in Rust, no need to abuse type names for that :)

examples/src/bin/composite_template.rs
```diff
@@ -0,0 +1,106 @@
...
+impl WindowImpl for ExApplicationWindowPriv {}
+impl ApplicationWindowImpl for ExApplicationWindowPriv {}
+
+glib_wrapper! {
+    pub struct ExApplicationWindow(ObjectSubclass<ExApplicationWindowPriv>)
+        @extends gtk::Widget, gtk::Window, gtk::ApplicationWindow, @implements gio::ActionMap, gio::ActionGroup;
+}
+
+impl ExApplicationWindow {
+    pub fn new<P: glib::IsA<gtk::Application> + glib::value::ToValue>(app: &P) -> Self {
```
sdroege, line 71:
```suggestion
    pub fn new<P: glib::IsA<gtk::Application>>(app: &P) -> Self {
```

After https://github.com/gtk-rs/gtk-rs/pull/156

examples/src/bin/composite_template.rs
```diff
@@ -0,0 +1,106 @@
...
+        let win = glib::Object::new(Self::static_type(), &[("application", app)])
+            .expect("Failed to create ExApplicationWindow")
+            .downcast::<ExApplicationWindow>()
+            .expect("Created object is of wrong type");
+        // init_template() must be run after the object is constructed.
+        win.init_template();
+        win.init_label();
```
sdroege, line 78: Oh really? You can't do that in `ObjectImpl::constructed()` after `self.parent_constructed()`?

sdroege: Actually this is too late here:

> It is important to call this function in the instance initializer of a GtkWidget subclass and not in GObject.constructed() or GObject.constructor() for two reasons.

That means you must call it in `instance_init()`. There is currently no hook for this (and that would also be unsafe). See `glib/src/subclass/types.rs:instance_init()`.

Would it be a problem to always call that for widget subclasses automatically?

examples/src/bin/composite_template.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    // Within class_init() you must set the template
+    // and bind it's children. The CompositeTemplate
+    // derive macro provides a convenience function
+    // bind_template_children() to bind all children
+    // at once.
+    fn class_init(klass: &mut Self::Class) {
+        let template = include_bytes!("composite_template.ui");
+        Self::set_template(klass, template);
+        Self::bind_template_children(klass);
```
sdroege, line 56: Also can this be automatically be called inside `override_vfuncs` in `widget.rs`?

examples/src/bin/composite_template.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    // Within class_init() you must set the template
+    // and bind it's children. The CompositeTemplate
+    // derive macro provides a convenience function
+    // bind_template_children() to bind all children
+    // at once.
+    fn class_init(klass: &mut Self::Class) {
+        let template = include_bytes!("composite_template.ui");
+        Self::set_template(klass, template);
```
sdroege, line 55: For other subclasses we're trying to move from class functions to associated functions or types on the `Impl` struct. Could something similar be done here instead?

gtk4/src/subclass/widget.rs
sdroege: Something like this would work for example. Then you can call functions directly on the class struct. Only ugly part left then is `Self::bind_template_children(klass)` but I'm not entirely sure how to solve that.

```diff
diff --git a/examples/src/bin/composite_template.rs b/examples/src/bin/composite_template.rs
index de22d7f..6bec20f 100644
--- a/examples/src/bin/composite_template.rs
+++ b/examples/src/bin/composite_template.rs
@@ -52,7 +52,7 @@ impl ObjectSubclass for ExApplicationWindowPriv {
     // at once.
     fn class_init(klass: &mut Self::Class) {
         let template = include_bytes!("composite_template.ui");
-        Self::set_template(klass, template);
+        klass.set_template(template);
         Self::bind_template_children(klass);
     }
 }
diff --git a/gtk4-macros/src/composite_template_derive.rs b/gtk4-macros/src/composite_template_derive.rs
index 9b95527..45a3365 100644
--- a/gtk4-macros/src/composite_template_derive.rs
+++ b/gtk4-macros/src/composite_template_derive.rs
@@ -28,8 +28,7 @@ fn gen_template_child_bindings(fields: &syn::Fields) -> TokenStream {
             }
 
             quote! {
-                Self::bind_template_child_with_offset(
-                    klass,
+                klass.bind_template_child_with_offset(
                     &#value_id,
                     #crate_ident::offset_of!(Self => #ident),
                 );
diff --git a/gtk4/src/subclass/widget.rs b/gtk4/src/subclass/widget.rs
index 965b335..962dbf2 100644
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -108,10 +108,10 @@ unsafe extern "C" fn widget_direction_changed<T: WidgetImpl>(
     imp.direction_changed(wrap.unsafe_cast_ref(), dirwrap)
 }
 
-pub unsafe trait WidgetClassSubclassExt: ObjectSubclass + WidgetImpl {
-    fn set_template_bytes(klass: &mut <Self as ObjectSubclass>::Class, template: &glib::Bytes) {
+pub unsafe trait WidgetClassSubclassExt: ClassStruct {
+    fn set_template_bytes(&mut self, template: &glib::Bytes) {
         unsafe {
-            let type_class = klass as *mut _ as *mut gobject_sys::GTypeClass;
+            let type_class = self as *mut _ as *mut gobject_sys::GTypeClass;
             let widget_class =
                 gobject_sys::g_type_check_class_cast(type_class, gtk_sys::gtk_widget_get_type())
                     as *mut gtk_sys::GtkWidgetClass;
@@ -119,22 +119,19 @@ pub unsafe trait WidgetClassSubclassExt: ObjectSubclass + WidgetImpl {
         }
     }
 
-    fn set_template(klass: &mut <Self as ObjectSubclass>::Class, template: &[u8]) {
+    fn set_template(&mut self, template: &[u8]) {
         let template_bytes = glib::Bytes::from(template);
-        Self::set_template_bytes(klass, &template_bytes);
+        self.set_template_bytes(&template_bytes);
     }
 
-    fn set_template_static(klass: &mut <Self as ObjectSubclass>::Class, template: &'static [u8]) {
+    fn set_template_static(&mut self, template: &'static [u8]) {
         let template_bytes = glib::Bytes::from_static(template);
-        Self::set_template_bytes(klass, &template_bytes);
+        self.set_template_bytes(&template_bytes);
     }
 
-    fn set_template_from_resource(
-        klass: &mut <Self as ObjectSubclass>::Class,
-        resource_name: &str,
-    ) {
+    fn set_template_from_resource(&mut self, resource_name: &str) {
         unsafe {
-            let type_class = klass as *mut _ as *mut gobject_sys::GTypeClass;
+            let type_class = self as *mut _ as *mut gobject_sys::GTypeClass;
             let widget_class =
                 gobject_sys::g_type_check_class_cast(type_class, gtk_sys::gtk_widget_get_type())
                     as *mut gtk_sys::GtkWidgetClass;
@@ -145,9 +142,9 @@ pub unsafe trait WidgetClassSubclassExt: ObjectSubclass + WidgetImpl {
         }
     }
 
-    fn bind_template_child(klass: &mut <Self as ObjectSubclass>::Class, name: &str) {
+    fn bind_template_child(&mut self, name: &str) {
         unsafe {
-            let type_class = klass as *mut _ as *mut gobject_sys::GTypeClass;
+            let type_class = self as *mut _ as *mut gobject_sys::GTypeClass;
             let widget_class =
                 gobject_sys::g_type_check_class_cast(type_class, gtk_sys::gtk_widget_get_type())
                     as *mut gtk_sys::GtkWidgetClass;
@@ -162,17 +159,19 @@ pub unsafe trait WidgetClassSubclassExt: ObjectSubclass + WidgetImpl {
 
     #[allow(clippy::missing_safety_doc)]
     unsafe fn bind_template_child_with_offset<T>(
-        klass: &mut <Self as ObjectSubclass>::Class,
+        &mut self,
         name: &str,
-        offset: field_offset::FieldOffset<Self, TemplateChild<T>>,
+        offset: field_offset::FieldOffset<Self::Type, TemplateChild<T>>,
     ) where
         T: ObjectType + FromGlibPtrNone<*mut <T as ObjectType>::GlibType>,
     {
-        let type_class = klass as *mut _ as *mut gobject_sys::GTypeClass;
+        let type_class = self as *mut _ as *mut gobject_sys::GTypeClass;
         let widget_class =
             gobject_sys::g_type_check_class_cast(type_class, gtk_sys::gtk_widget_get_type())
                 as *mut gtk_sys::GtkWidgetClass;
-        let private_offset = Self::type_data().as_ref().private_offset;
+        let private_offset = <Self::Type as ObjectSubclass>::type_data()
+            .as_ref()
+            .private_offset;
         gtk_sys::gtk_widget_class_bind_template_child_full(
             widget_class,
             name.to_glib_none().0,
@@ -182,7 +181,7 @@ pub unsafe trait WidgetClassSubclassExt: ObjectSubclass + WidgetImpl {
     }
 }
 
-unsafe impl<T: ObjectSubclass + WidgetImpl> WidgetClassSubclassExt for T {}
+unsafe impl<T: ClassStruct> WidgetClassSubclassExt for T where T::Type: WidgetImpl {}
 
 #[derive(Debug, PartialEq, Eq)]
 #[repr(transparent)]
@@ -217,6 +216,6 @@ where
     }
 }
 
-pub trait CompositeTemplate: WidgetClassSubclassExt {
+pub trait CompositeTemplate: WidgetImpl {
     fn bind_template_children(klass: &mut Self::Class);
 }
```

examples/src/bin/composite_template.rs
sdroege: I guess the best for now would be to call it from `ObjectImpl::constructed()` before `parent_constructed()`. That's as early as we can do currently.

And then create an issue for figuring out a way to call it even earlier.

sdroege: That means that you can also do `init_label()` from `constructed()` already btw

examples/src/bin/composite_template.rs
sdroege: We can also worry about that separately afterwards, but in that case create an issue :)

gtk4/src/subclass/widget.rs
BrainBlasted: I'm not sure we can do the second part, but we are able to derive PartialEq and Eq; For comparing the inner child I think
developers could use `child.get() == obj`

gtk4-macros/src/util.rs
```diff
@@ -0,0 +1,82 @@
+use anyhow::{bail, Result};
```
bilelmoussaoui, line 1: I would avoid adding a dependency on anyhow here and use a proper error type

examples/src/bin/composite_template.rs
sdroege: That's annoying :)

gtk4-macros/src/util.rs
sdroege: glib-macros also depends on anyhow so :man_shrugging:

bilelmoussaoui: Ah right, well, it doesn't matter then

examples/src/bin/composite_template.rs
```diff
@@ -0,0 +1,114 @@
+//! # Composite Template Example
+//!
+//! This sample demonstrates how to create a widget using GTK's composite templates.
+
+#[macro_use]
+extern crate glib;
+#[macro_use]
+extern crate gtk;
```
bilelmoussaoui, line 8:
```suggestion
```

examples/src/bin/composite_template.rs
```diff
@@ -0,0 +1,114 @@
...
+
+mod imp {
+    use super::*;
+    use glib::subclass;
+    use gtk::subclass::prelude::*;
+    use gtk::subclass::widget::*;
+    use gtk::WidgetExt;
+
+    /// The private struct, which can hold widgets and other data.
+    #[derive(Debug, CompositeTemplate)]
```
bilelmoussaoui, line 23:
```suggestion
    #[derive(Debug, gtk::CompositeTemplate)]
```

examples/src/bin/composite_template.rs
```diff
@@ -0,0 +1,114 @@
...
+            self.parent_constructed(obj);
+        }
+    }
+
+    impl WidgetImpl for ExApplicationWindow {}
+    impl WindowImpl for ExApplicationWindow {}
+    impl ApplicationWindowImpl for ExApplicationWindow {}
+}
+
+glib_wrapper! {
```
bilelmoussaoui, line 77:
```suggestion
glib::glib_wrapper! {
```

sdroege (approved):

examples/src/bin/composite_template.rs
sdroege: Indeed, I missed the Rust 2015-isms in the example. Thanks :)

examples/src/bin/composite_template.rs
BrainBlasted: I've changed this in a different way - mind taking a look?

examples/src/bin/composite_template.rs
BrainBlasted: I've changed this in a different way - mind taking a look?

bilelmoussaoui (approved):

bilelmoussaoui:
@BrainBlasted thanks a lot for working on this! Do you mind opening an issue so we can track the missing/possible future improvements

sdroege:
> Do you mind opening an issue so we can track the missing/possible future improvements ?

Yes please, otherwise we just forget all that :) There were about 3-4 different things mentioned in the different discussions above.

examples/src/bin/composite_template.rs
```diff
@@ -2,15 +2,10 @@
...
-#[macro_use]
-extern crate glib;
-#[macro_use]
-extern crate gtk;
-
 use gio::{prelude::ApplicationExtManual, ApplicationExt};
 use glib::subclass::prelude::*;
-use glib::{Cast, StaticType};
-use gtk::WidgetExt;
+use glib::{glib_object_subclass, glib_wrapper, Cast, StaticType};
```
sdroege, line 7: For `Cast`, `StaticType`, please just `use glib::prelude::*`

examples/src/bin/composite_template.rs
```diff
@@ -2,15 +2,10 @@
...
+use gtk::{CompositeTemplate, WidgetExt};
```
sdroege, line 8: And here use `gtk::prelude::*` for `WidgetExt`

examples/src/bin/composite_template.rs
```diff
@@ -2,15 +2,10 @@
 //!
 //! This sample demonstrates how to create a widget using GTK's composite templates.
 
-#[macro_use]
-extern crate glib;
-#[macro_use]
-extern crate gtk;
-
 use gio::{prelude::ApplicationExtManual, ApplicationExt};
```
sdroege, line 5: And here `use gio::prelude::*`

The reason is that things might move between different extension traits but they'll always be part of the prelude. We don't want to change all the code everywhere whenever that happens.

examples/src/bin/composite_template.rs
sdroege: Seems OK to me too

BrainBlasted:
Issue created: https://github.com/gtk-rs/gtk4-rs/issues/77

sdroege:
Good for me then, @bilelmoussaoui you merge when you and the CI are happy?

PR #55, opened 2020-11-16
Resync with gtk-rs & regen

bilelmoussaoui:


gdk4-wayland/sys/src/lib.rs
```diff
@@ -8,6 +8,7 @@
     clippy::type_complexity,
     clippy::unreadable_literal
 )]
+#![cfg_attr(feature = "dox", feature(doc_cfg))]
```
sdroege, line 11: This probably breaks the CI now :)

sdroege:
Seems good to me otherwise

gdk4-wayland/sys/src/lib.rs
bilelmoussaoui: Fixed it

sdroege (approved):

Issue #56, opened 2020-11-16
Subclassing support: virtual methods


bilelmoussaoui:
Here's a list of Classes/Interfaces that along with their associate virtual methods

- [ ] Buildable - `add_child`
- [ ] Buildable - `construct_child`
- [ ] Buildable - `custom_finished`
- [ ] Buildable - `custom_tag_end`
- [ ] Buildable - `custom_tag_start`
- [ ] Buildable - `get_id`
- [ ] Buildable - `get_internal_child`
- [ ] Buildable - `parser_finished`
- [ ] Buildable - `set_buildable_property`
- [ ] Buildable - `set_id`
- [ ] ShortcutManager - `add_controller`
- [ ] ShortcutManager - `remove_controller`
- [ ] Widget - `compute_expand`
- [ ] Widget - `contains`
- [ ] Widget - `css_changed`
- [ ] Widget - `direction_changed`
- [ ] Widget - `focus`
- [ ] Widget - `get_request_mode`
- [ ] Widget - `grab_focus`
- [ ] Widget - `hide`
- [ ] Widget - `keynav_failed`
- [ ] Widget - `map`
- [ ] Widget - `measure`
- [ ] Widget - `mnemonic_activate`
- [ ] Widget - `move_focus`
- [ ] Widget - `query_tooltip`
- [ ] Widget - `realize`
- [ ] Widget - `root`
- [ ] Widget - `set_focus_child`
- [ ] Widget - `show`
- [ ] Widget - `size_allocate`
- [ ] Widget - `snapshot`
- [ ] Widget - `state_flags_changed`
- [ ] Widget - `system_setting_changed`
- [ ] Widget - `unmap`
- [ ] Widget - `unrealize`
- [ ] Widget - `unroot`
- [ ] Window - `activate_default`
- [ ] Window - `activate_focus`
- [ ] Window - `close_request`
- [ ] Window - `enable_debugging`
- [ ] Window - `keys_changed`
- [ ] Actionable - `get_action_name`
- [ ] Actionable - `get_action_target_value`
- [ ] Actionable - `set_action_name`
- [ ] Actionable - `set_action_target_value`
- [ ] Adjustment - `changed`
- [ ] Adjustment - `value_changed`
- [ ] Dialog - `close`
- [ ] Dialog - `response`
- [ ] Application - `window_added`
- [ ] Application - `window_removed`
- [ ] LayoutManager - `allocate`
- [ ] LayoutManager - `create_layout_child`
[...]

PR #58, opened 2020-11-18
Subclassing round 2

bilelmoussaoui:
Add subclassing support of the following widgets/types

- `GtkPopover`
- `GtkToggleButton`
- `GtkCheckButton`
- `GtkEntry`
- `GtkNativeDialog`
- `GtkStyleContext`

Fixes/updates 
- `GtkDialog`

Part of #56

### DIFF
--- a/gtk4/src/subclass/native_dialog.rs
+++ b/gtk4/src/subclass/native_dialog.rs
@@ -1,0 +1,108 @@
+use gtk_sys;
+
+use glib::subclass::object::ObjectImpl;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, Object};
+
+use NativeDialog;
+use ResponseType;
+
+pub trait NativeDialogImpl: NativeDialogImplExt + ObjectImpl {
+    fn response(&self, dialog: &Self::Type, response: ResponseType) {
+        self.parent_response(dialog, response)
+    }
+
+    fn show(&self, dialog: &Self::Type) {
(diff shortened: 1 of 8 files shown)


gtk4/src/subclass/native_dialog.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    fn parent_show(&self, dialog: &Self::Type) {
+        unsafe {
+            let data = T::type_data();
+            let parent_class =
+                data.as_ref().get_parent_class() as *mut gtk_sys::GtkNativeDialogClass;
+            let f = (*parent_class)
+                .show
+                .expect("No parent class impl for \"show\"");
```
bilelmoussaoui, line 53: Just to make sure I understood how things works properly, this can't be `NULL` per https://github.com/GNOME/gtk/blob/master/gtk/gtknativedialog.c#L308 ?

sdroege: Yes, that's correct

sdroege: You could also do nothing here though, it doesn't matter much because you don't have to invent a return value out of thin air

sdroege (approved):

gtk4/src/subclass/native_dialog.rs
bilelmoussaoui: Right, thanks!

PR #60, opened 2020-11-19
Subclassing part 3

bilelmoussaoui:
Add 
- `GtkRecentManager`
- `GtkMediaFile`
- `GtkActionable`
- `GtkEditable`

part of #56

### DIFF
--- a/gtk4/src/subclass/editable.rs
+++ b/gtk4/src/subclass/editable.rs
@@ -1,0 +1,156 @@
+use crate::Editable;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, GString};
+use libc::{c_char, c_int};
+
+pub trait EditableImpl: ObjectImpl {
+    fn insert_text(&self, editable: &Self::Type, text: &str, length: i32, position: &mut i32);
+    fn delete_text(&self, editable: &Self::Type, start_position: i32, end_position: i32);
+    fn changed(&self, editable: &Self::Type);
+    fn get_text(&self, editable: &Self::Type) -> GString;
+    fn do_insert_text(&self, editable: &Self::Type, text: &str, length: i32, position: &mut i32);
+    fn do_delete_text(&self, editable: &Self::Type, start_position: i32, end_position: i32);
+    fn get_selection_bounds(&self, editable: &Self::Type) -> Option<(i32, i32)>;
+    fn set_selection_bounds(&self, editable: &Self::Type, start_position: i32, end_position: i32);
+}
+
+unsafe impl<T: EditableImpl> IsImplementable<T> for Editable {
+    unsafe extern "C" fn interface_init(
+        iface: glib::ffi::gpointer,
+        _iface_data: glib::ffi::gpointer,
+    ) {
+        let editable_iface = &mut *(iface as *mut ffi::GtkEditableInterface);
+
+        editable_iface.insert_text = Some(editable_insert_text::<T>);
+        editable_iface.delete_text = Some(editable_delete_text::<T>);
+        editable_iface.changed = Some(editable_changed::<T>);
+        editable_iface.get_text = Some(editable_get_text::<T>);
+        editable_iface.do_insert_text = Some(editable_do_insert_text::<T>);
+        editable_iface.do_delete_text = Some(editable_do_delete_text::<T>);
+        editable_iface.get_selection_bounds = Some(editable_get_selection_bounds::<T>);
+        editable_iface.set_selection_bounds = Some(editable_set_selection_bounds::<T>);
+    }
+}
+
+unsafe extern "C" fn editable_insert_text<T: EditableImpl>(
+    editable: *mut ffi::GtkEditable,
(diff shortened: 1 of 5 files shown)


gtk4/src/subclass/editable.rs
```diff
@@ -0,0 +1,154 @@
...
+    start_position: *mut i32,
+    end_position: *mut i32,
+) -> glib_sys::gboolean {
+    let instance = &*(editable as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    imp.get_selection_bounds(
+        from_glib_borrow::<_, Editable>(editable).unsafe_cast_ref(),
+        &mut *start_position,
+        &mut *end_position,
```
sdroege, line 136: Either or both of the pointers can be `NULL`. You need to check before assigning the results to them

sdroege (commented):
Ok otherwise

gtk4/src/subclass/editable.rs
bilelmoussaoui: Should be ok now

gtk4/src/subclass/editable.rs
```diff
@@ -0,0 +1,165 @@
...
+    fn delete_text(&self, editable: &Self::Type, start_position: i32, end_position: i32);
+    fn changed(&self, editable: &Self::Type);
+    fn get_text(&self, editable: &Self::Type) -> GString;
+    fn do_insert_text(&self, editable: &Self::Type, text: &str, length: i32, position: &mut i32);
+    fn do_delete_text(&self, editable: &Self::Type, start_position: i32, end_position: i32);
+    fn get_selection_bounds(
+        &self,
+        editable: &Self::Type,
+        start_position: &mut i32,
+        end_position: &mut i32,
```
sdroege, line 21: Should this maybe return an `Option<(i32, i32)>` instead of returning a bool and having two out parameters? That's at least what we would convert it to for the normal bindings code.

sdroege: > start_pos will be filled with the start of the selection and end_pos with end. If no text was selected both will be identical and FALSE will be returned.

This probably means that the most reasonable here would actually be a `Result<(i32, i32), i32>` on the Rust side. You should update the normal bindings accordingly too.

sdroege:
Fixed compilation of the editable subclassing but the API should still be reconsidered.

gtk4/src/subclass/editable.rs
bilelmoussaoui: Wouldn't an option makes more sense as editable doesn't really fail to get the selection bounds?

sdroege: It says that when `FALSE` is returned, both positions will be the same. So some position still has to be returned, it's just the same for both values

bilelmoussaoui: right, but that from our side means we have "no selection bound" because both values are the same. If we are to provide a better API around this one, we should compare the values & return an Option. Does that sound sane to you? :)

sdroege: Wouldn't that lose the information about which same value both positions are? Does that same position have any actual meaning? I don't know this API or how it's used so you'll have to figure that one out :)

bilelmoussaoui: Hmm very valid point, I will see if there are any usages of the API in the gtk-demos

bilelmoussaoui: From the researches I have done it seems the api is used for two cases:
- Detect whether something is selected or not, in such case, a `NULL`  pointer is passed to both start_pos & end_pos and the usage only cares about whether it returns true or false. 

- Get the position of what's selected if there was anything selected: always checks first if the methods returns `true` and either adds/deletes text or emit a signal. 

So I think that same position value isn't actually useful as part of how this api is intended to be used. There's a `cursor-position` prop on the widgets that implements the Editable interface which should give that information to the user in case needed. I guess we should go with returning a Option<(i32, i32)> which looks the most accurate to me compared to the C API.

sdroege: That makes sense to me at least. Hard to say from the docs if that's the only intended usage, but worst case this can be improved later.

bilelmoussaoui: Definitely. I will go ahead and implement that then :) Once I'm more or less happy with the progress on #56 I will start porting some of the custom & complex widgets from the C demo

gtk4/src/subclass/editable.rs
```diff
@@ -0,0 +1,154 @@
...
+        if !start_position.is_null() {
+            *start_position = start_pos;
+        }
+
+        if !end_position.is_null() {
+            *end_position = end_pos;
+        }
+        true.to_glib()
+    } else {
+        false.to_glib()
```
sdroege, line 137: You might want to set the two values to `0` here as the docs at least say that the positions should have the same value when returning `FALSE`

sdroege:
Looks good to me otherwise

PR #62, opened 2020-11-20
Subclassing round 4

bilelmoussaoui:
- Add `GtkLayoutManager` 
- Add `GtkScrollable`
- Add `GtkFilter`
- Add `GtkSorter`

Part of #56

### DIFF
--- a/gtk4/src/subclass/layout_manager.rs
+++ b/gtk4/src/subclass/layout_manager.rs
@@ -1,0 +1,364 @@
+use glib::subclass::object::ObjectImpl;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, Object};
+use libc::c_int;
+
+use crate::{LayoutChild, LayoutManager, Orientation, SizeRequestMode, Widget};
+
+pub trait LayoutManagerImpl: LayoutManagerImplExt + ObjectImpl {
+    fn allocate(
+        &self,
+        layout_manager: &Self::Type,
+        widget: &Widget,
+        width: i32,
+        height: i32,
+        baseline: i32,
+    ) {
+        self.parent_allocate(layout_manager, widget, width, height, baseline)
+    }
+
+    fn create_layout_child(
+        &self,
+        layout_manager: &Self::Type,
+        widget: &Widget,
+        for_child: &Widget,
+    ) -> LayoutChild {
+        self.parent_create_layout_child(layout_manager, widget, for_child)
+    }
+
+    fn get_request_mode(&self, layout_manager: &Self::Type, widget: &Widget) -> SizeRequestMode {
+        self.parent_get_request_mode(layout_manager, widget)
+    }
+
+    fn measure(
+        &self,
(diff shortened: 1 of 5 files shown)


gtk4/src/subclass/layout_manager.rs
```diff
@@ -0,0 +1,294 @@
...
+        height: i32,
+        baseline: i32,
+    ) {
+        self.parent_allocate(layout_manager, widget, width, height, baseline)
+    }
+    fn create_layout_child(
+        &self,
+        layout_manager: &Self::Type,
+        widget: &Widget,
+        for_child: &Widget,
```
sdroege, line 31: Can any of those widgets be `NULL`?

gtk4/src/subclass/scrollable.rs
```diff
@@ -0,0 +1,36 @@
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::Cast;
+use gtk_sys;
+
+use Border;
+use Scrollable;
+
+pub trait ScrollableImpl: ObjectImpl {
+    fn get_border(&self, scrollable: &Self::Type, border: &mut Border) -> bool;
```
sdroege, line 10: Should this return an `Option<Border>` maybe?

sdroege:
Looks good otherwise :)

gtk4/src/subclass/scrollable.rs
bilelmoussaoui: I did an attempt at this, not sure if the code looks that good

gtk4/src/subclass/layout_manager.rs
bilelmoussaoui: From what I understood from the C API the widget calls this function from the layout manager by passing it a pointer of itself as `widget` & the child widget as `for_child` . So they can't be nullable

gtk4/src/subclass/scrollable.rs
```diff
@@ -0,0 +1,35 @@
...
+    scrollable: *mut ffi::GtkScrollable,
+    borderptr: *mut ffi::GtkBorder,
+) -> glib::ffi::gboolean {
+    let instance = &*(scrollable as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    let border = imp.get_border(from_glib_borrow::<_, Scrollable>(scrollable).unsafe_cast_ref());
+    border
+        .map(|b| {
+            *borderptr = *b.to_glib_full();
```
sdroege, line 31: You should probably set the pointer to `NULL` if there is no border, just in case.

gtk4/src/subclass/scrollable.rs
sdroege: I think that looks much better, but there's a bug in it :)

gtk4/src/subclass/scrollable.rs
bilelmoussaoui: Good point, done

sdroege (approved):

PR #67, opened 2020-11-21
Add some trust_return_value_nullability's for gtk

sophie-h:


sdroege:
@sophie-h You might want to create an issue like https://github.com/gtk-rs/gtk-rs/issues/163 also here in gtk4-rs to keep track of this :)

gtk4/Gir.toml
```diff
@@ -404,33 +404,48 @@ status = "generate"
...
 generate_builder = true
+# upstream_needs_fix
+trust_return_value_nullability = true
     [[object.signal]]
     name = "activate-link"
     inhibit = true
+    [[object.function]]
+    name = "get_program_name"
+        [object.function.return]
+        nullable = true
```
sdroege, line 415: Please link to an upstream issue/commit for these things, or mention an upstream version where it's fixed. That way we can properly keep track of it and remove the override when possible.

sdroege (commented):
Looks good to me otherwise, thanks for taking care of this :)

gtk4/Gir.toml
sophie-h: Should me much clearer now.

generator.py
```diff
@@ -171,7 +171,7 @@ def main():
         conf["gir_path"] = "./gir/target/release/gir"
 
     print('=> Regenerating crates...')
-    if not regen_crates(".", conf):
+    if not regen_crates("gtk4", conf):
```
bilelmoussaoui, line 174: I don't think this change is needed?

sophie-h: Oh wow, no! that was just for speeding things up. Will add another MR to make this an option.

sdroege (approved):

bilelmoussaoui:
Thanks!

PR #69, opened 2020-11-21
[generator.py] Use argparse and option for paths

sophie-h:


bilelmoussaoui:
Overall the changes looks very good to me, mind updating the one in gtk-rs/gtk-rs as well?

sophie-h:
Good point @bilelmoussaoui, I included the last change from gtk-rs. I will create a PR after this is approved.

generator.py
```diff
@@ -1,7 +1,9 @@
-#!/usr/bin/env python3
```
GuillaumeGomez, line 1: Please revert this change. The path isn't always `/usr/bin` for python3 (in my case it's in `sbin` for example :p)

generator.py
```diff
@@ -78,102 +83,72 @@ def regen_crates(path, conf, level=0):
         if isdir(entry_file):
             if level < 2 and not regen_crates(entry_file, conf, level + 1):
                 return False
-        elif entry == 'Gir.toml':
+        elif fnmatch.fnmatchcase(entry, "Gir*.toml"):
```
GuillaumeGomez, line 86: Seems a bit overkill compared to just `elif entry.startwith("Gir") and entry.endswith(".toml"):`...

generator.py
```diff
@@ -10,6 +12,9 @@
 NEED_UPDATE = 1
 FAILURE = 2
 
+DEFAULT_GIR_FILES = 'gir-files'
```
GuillaumeGomez, line 15: What do you think about `DEFAULT_GIR_FILES_FOLDER` instead?

generator.py
```diff
@@ -78,102 +83,72 @@ def regen_crates(path, conf, level=0):
...
-    if conf["gir_files"] is None:
+    if conf.gir_files == DEFAULT_GIR_FILES:
         if def_check_submodule("gir-files", conf) == FAILURE:
             return 1
-        conf["gir_files"] = "gir-files"
-    if conf["gir_path"] is None:
+    else:
+       if not isdir(conf.gir_files):
+            print("`{}` dir doesn't exist. Aborting...".format(path))
+            return 1
```
GuillaumeGomez, line 136:
```suggestion
    elif not isdir(conf.gir_files):
        print("`{}` dir doesn't exist. Aborting...".format(path))
        return 1
```

generator.py
```diff
@@ -78,102 +83,72 @@ def regen_crates(path, conf, level=0):
...
+
+    if conf.gir_path == DEFAULT_GIR_PATH:
         if not build_gir_if_needed(def_check_submodule("gir", conf)):
             return 1
-        conf["gir_path"] = "./gir/target/release/gir"
+    else:
+       if not isfile(conf.gir_path):
+            print("`{}` file doesn't exist. Aborting...".format(path))
+            return 1
```
GuillaumeGomez, line 144:
```suggestion
   elif not isfile(conf.gir_path):
        print("`{}` file doesn't exist. Aborting...".format(path))
        return 1
```

generator.py
```diff
@@ -78,102 +83,72 @@ def regen_crates(path, conf, level=0):
...
 
     print('=> Regenerating crates...')
-    if not regen_crates(".", conf):
-        return 1
-    if conf["run_fmt"] is True and not run_command(['cargo', 'fmt']):
+    for path in conf.path:
```
GuillaumeGomez, line 147: Nice add!

GuillaumeGomez (changes requested):
Lot of small things but overall it's pretty nice! I especially like the new `path` option.

generator.py
sophie-h: That's the only thing I wanted to change and I thought it would be simpler using argparse xD

generator.py
sophie-h: This is named after "--gir-files". Would only rename it if we rename the option.

sophie-h: And `--gir-files-dir` would probably be clearer

generator.py
sophie-h: I would call it python style, but I don't care.

generator.py
sophie-h: Ups

generator.py
sophie-h: Fixed

generator.py
GuillaumeGomez: Then please do the change here and open an issue (or even a PR) on `gir`. :)

sophie-h: Wait, I'm confused. What does gir have to do with this?

sophie-h: If we do it consistent with `gir`, it should be `--gir-directory`

GuillaumeGomez: Arf sorry, don't know why I thought you were talking about `gir`... So yes, please feel free to update the option name here.

GuillaumeGomez: Fine by me!

sophie-h:
Done from my side

generator.py
```diff
@@ -78,102 +82,70 @@ def regen_crates(path, conf, level=0):
...
+    return parser.parse_args()
 
 
 def main():
     gir_path = None
 
     conf = parse_args(sys.argv[1:])
-    if conf is None:
-        return 1
+    print(conf)
```
GuillaumeGomez, line 127: Maybe remove this line? ;)

sophie-h: How pedantic of you :angry: :laughing:

GuillaumeGomez (approved):

generator.py
GuillaumeGomez: I like `print`, but too much `print` kills my brain. :p

GuillaumeGomez:
Thanks!

PR #70, opened 2020-11-21
Add flatpak manifest

sophie-h:


examples/com.github.gtk4_rs.examples.json
```diff
@@ -0,0 +1,46 @@
...
+    "runtime" : "org.gnome.Platform",
+    "runtime-version" : "master",
+    "sdk" : "org.gnome.Sdk",
+    "sdk-extensions" : [
+        "org.freedesktop.Sdk.Extension.rust-stable"
+    ],
+    "tags" : [
+        "development"
+    ],
+    "desktop-file-name-suffix" : " ☢",
```
bilelmoussaoui, line 12:
```suggestion
```

examples/com.github.gtk4_rs.examples.json
```diff
@@ -0,0 +1,46 @@
...
+        "--socket=wayland",
+        "--talk-name=org.a11y.Bus"
+    ],
+    "build-options" : {
+        "append-path" : "/usr/lib/sdk/rust-stable/bin",
+        "build-args" : [
+            "--share=network"
+        ]
+    },
+    "command": "basic",
```
bilelmoussaoui, line 26: Move this one before finish-args, it's way less noticeable here

bilelmoussaoui:
I think it's probably better to provide a proper example inside the examples directory with the flatpak manifest, desktop file & the sources necessary (main.rs/Cargo.toml) and keep the other examples separate.  

@BrainBlasted @sdroege Any opinions?

BrainBlasted:
I think that would be good too, as an example of how to create a full app using gtk-rs.

sdroege:
Seems useful as part of the docs at least, yes

alatiera:
I would *highly* suggest depending on a stable runtime and bundling the same version/commit of gtk4 we generated the girs from.

bilelmoussaoui:
@sophie-h how can we move forward with this? now that we have a subdir per example, I think we can now have a flatpak example with a very simple application, a flatpak manifest & whatever other stuff we need for it to work. GTK 4 is also now available in the GNOME 40 runtime which solves @alatiera's remark.

sophie-h:
Idk, I think would just close this PR. @BrainBlasted's idea of a 'full app' example seems reasonable. But this also has some overlap with your gtk-rust-template.

Maybe a more complete example with Flatpak support would make sense as part of the book? @Hofer-Julian any thoughts?

bilelmoussaoui:
The book already redirects users to gtk-rust-template, see https://gtk-rs.org/gtk4-rs/git/book/prerequisites.html?highlight=template#cargo--meson. We could definitely have a chapter about how to Flatpak your application based on the article I wrote a year ago.

I have also thought in the past of moving the template to gtk4-rs & have it generated using `cargo generate` but that's unrelated to this PR.

sophie-h:
Yeah, really not sure how this could fit in. Closing for now.

Hofer-Julian:
Yeah, seems like a good candidate for a bonus chapter

Issue #68, opened 2020-11-21
trust_return_value_nullability


bilelmoussaoui:
- [ ] gdk
- [x] gdk-wayland
- [x] gdk-x11
- [x] gsk
- [ ] gtk ([1](https://github.com/gtk-rs/gtk4-rs/pull/67))

cc @sophie-h @sdroege

bilelmoussaoui:
I have reached out to mclassen and there are plans for an rc release soon( by December 1st), it would be nice if we can send as many fixes as we can.

PR #71, opened 2020-11-22
[gtk] Default to trust_return_value_nullability

sophie-h:
Can we do it this way round? Here, every Object marked with `trust_return_value_nullability = false` should be checked and fixed at some point.

#68

### DIFF
--- a/gtk4/Gir.toml
+++ b/gtk4/Gir.toml
@@ -8,6 +8,7 @@ work_mode = "normal"
 generate_safety_asserts = true
 deprecate_by_min_version = true
 single_version_file = true
+trust_return_value_nullability = true
 
 generate = [
     # "Gtk.BuilderCScope", not useful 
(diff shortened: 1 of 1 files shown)


sdroege:
Makes it more obvious which ones still need to be taken care of, so I like it

bilelmoussaoui:
Sounds good to me, but let's wait until the regen with 2018 edition in first.

bilelmoussaoui:
The regen with 2018 edition is now, this can now be rebased. It seems that the bindings were not re-generated btw

sophie-h:
The point of this PR is that nothing in the bindings changes at all. It's just a change in how we use `trust_return_value_nullability`.

bilelmoussaoui:
Ah my bad. Didn't notice the global change of `trust_return_value_nullability`

PR #72, opened 2020-11-22
2018 migration fixes

bilelmoussaoui:
not sure how ci didn't detect any of those


gdk4/src/lib.rs
```diff
@@ -12,12 +12,11 @@
...
+extern crate bitflags;
+extern crate cairo;
 extern crate gdk_pixbuf;
 extern crate gio;
-#[macro_use]
 extern crate glib;
-extern crate bitflags;
-extern crate cairo;
 extern crate libc;
 extern crate pango;
```
sdroege, line 21: Hrm you shouldn't have any `extern crate` at all afterwards :) They're redundant with Rust 2018

bilelmoussaoui: Oh I wasn't aware of that, thanks!

bilelmoussaoui:
This turned out to be more of a generic cleanup/fixes :)

gdk4/src/lib.rs
sdroege: Just like `use $crate;` it's not needed anymore. Dependencies are always in scope

PR #78, opened 2020-11-24
subclass/widget: Rework the default measure() vfunc

BrainBlasted:
The way we previously handled measure didn't work
with the way measure actually works. It caused
a critical issue as widget subclasses were using
a measure func that did not behave properly.

This would lead to a crash.

With these changes we now handle measure in a simple
"default" way.

### DIFF
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -1,5 +1,3 @@
-use std::mem;
-
 use glib::prelude::*;
 use glib::subclass::prelude::*;
 use glib::translate::*;
@@ -53,8 +51,12 @@ pub trait WidgetImpl: WidgetImplExt + ObjectImpl {
         widget: &Self::Type,
         orientation: Orientation,
         for_size: i32,
-    ) -> (i32, i32, i32, i32) {
-        self.parent_measure(widget, orientation, for_size)
+        min: &mut i32,
+        nat: &mut i32,
+        min_base: &mut i32,
+        nat_base: &mut i32,
+    ) {
+        self.parent_measure(widget, orientation, for_size, min, nat, min_base, nat_base)
     }
 
     fn mnemonic_activate(&self, widget: &Self::Type, group_cycling: bool) -> bool {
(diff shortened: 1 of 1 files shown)


sdroege (approved):
Seems fine to me but needs some checking if it's guaranteed that all callers initialize the 4 values

bilelmoussaoui:
> Seems fine to me but needs some checking if it's guaranteed that all callers initialize the 4 values

Does it matter that much? because we have default values that corresponds to the C implementation ones unless I'm missing some case.  I have tried it with various widgets on a application I'm porting to gtk4 and it seems to works fine.

I will go ahead and merge this one, so that we can discover others :P

bilelmoussaoui (approved):

sdroege:
> Does it matter that much?

If it's not guaranteed then we would initialize on our side with random values and that potentially breaks things in a hard to debug way.

bilelmoussaoui:
From the examples of code i've found in gtk 's source tree, the only the two latest parameters that can expect a NULL as a value which defaults to -1. 
Per the docs a widget doesn't have to provide a measure function if it uses a layout manager but a layout manager doesn't care about the measure implementation at all so it shouldn't cause an issue from that side.

sdroege:
> From the examples of code i've found in gtk 's source tree, only the two latest parameters that can expect a NULL as a value which defaults to -1.

Them being `NULL` is not so much a problem. A problem would be if they're not `NULL` but the caller does not initialize them to something meaningful (0/0/-1/-1).

>  Per the docs a widget doesn't have to provide a measure function if it uses a layout manager but a layout manager doesn't care about the measure implementation at all so it shouldn't cause an issue from that side.

`GtkWidget` implements that virtual method, so every widget will have some implementation of it. And (some?) layout managers seem to be calling the widget implementations of this function to figure out things.

BrainBlasted:
I think those initial values are right, and then calling `imp.measure()` will mutate them according to the handling of the parent class if the user does not override `measure()`. I'm not sure where a problem comes in. This fixes the issue we had before with incorrect measure, I think.

sdroege:
The issue would be if C code calls `klass->measure()` with pointers to uninitialized integers and assumes that the function initializes them without reading first. That does not seem to be the case in callers so it's probably fine here.

PR #81, opened 2020-11-24
Fix some nullable return annotations 2

sophie-h:
Issue #68

### DIFF
--- a/gtk4/Gir.toml
+++ b/gtk4/Gir.toml
@@ -54,6 +54,7 @@ generate = [
     "Gtk.CellRendererState",
     "Gtk.Clipboard",
     "Gtk.ConstraintAttribute",
+    "Gtk.ConstraintLayout",
     "Gtk.ConstraintLayoutChild",
     "Gtk.ConstraintRelation",
     "Gtk.ConstraintStrength",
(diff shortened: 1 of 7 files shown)


gtk4/src/auto/header_bar.rs
```diff
@@ -36,7 +36,7 @@ impl HeaderBar {
         unsafe { Widget::from_glib_none(ffi::gtk_header_bar_new()).unsafe_cast() }
     }
 
-    pub fn get_decoration_layout(&self) -> Option<glib::GString> {
+    pub fn get_decoration_layout(&self) -> glib::GString {
```
bilelmoussaoui, line 39: This should be an option per your upstream patch

bilelmoussaoui:
looks good to me otherwise, do you want to exchange reviews with me on #83 ? :)

gtk4/src/auto/header_bar.rs
sophie-h: True, upstream is correct. Thanks!

bilelmoussaoui:
Thanks!

sophie-h:
Sorry @bilelmoussaoui didn't had time for a review. Since upstream didn't merged your complete MR, does this need some kind of update?

PR #84, opened 2020-11-24
gtk: fix widget's set_focus_child implementation

bilelmoussaoui:
the focused child ptr can be null in case there are no currently focused child

Fixes an issue discovered by Chris


bilelmoussaoui:
cc @brainblasted

BrainBlasted:
This fixes an issue that could cause a crash on opening a popover.

bilelmoussaoui:
Thanks for confirming that the patch fixes your issue! Merging so that it won't block your work :) sorry for the inconvenience :P

PR #87, opened 2020-11-25
gtk: add TextBuffer/TextView subclassing support

bilelmoussaoui:
Part of #56

### DIFF
--- a/gtk4/src/subclass/text_buffer.rs
+++ b/gtk4/src/subclass/text_buffer.rs
@@ -1,0 +1,550 @@
+use crate::{TextBuffer, TextChildAnchor, TextIter, TextMark, TextTag};
+use glib::subclass::object::ObjectImpl;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, Object};
+
+pub trait TextBufferImpl: TextBufferImplExt + ObjectImpl {
+    fn apply_tag(&self, text_buffer: &Self::Type, tag: &TextTag, start: &TextIter, end: &TextIter) {
+        self.parent_apply_tag(text_buffer, tag, start, end)
+    }
+    fn begin_user_action(&self, text_buffer: &Self::Type) {
+        self.parent_begin_user_action(text_buffer)
+    }
+    fn changed(&self, text_buffer: &Self::Type) {
+        self.parent_changed(text_buffer)
+    }
+    fn delete_range(&self, text_buffer: &Self::Type, start: &mut TextIter, end: &mut TextIter) {
+        self.parent_delete_range(text_buffer, start, end)
+    }
+    fn end_user_action(&self, text_buffer: &Self::Type) {
+        self.parent_end_user_action(text_buffer)
+    }
+    fn insert_child_anchor(
+        &self,
+        text_buffer: &Self::Type,
+        iter: &mut TextIter,
+        anchor: &TextChildAnchor,
+    ) {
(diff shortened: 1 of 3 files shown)


sdroege:
Let me know when it's ready for review or you have questions :)

gtk4/src/subclass/text_view.rs
```diff
@@ -0,0 +1,424 @@
...
+                    granularity.to_glib(),
+                    location.to_glib_none().0,
+                    start.to_glib_none_mut().0,
+                    end.to_glib_none_mut().0,
+                )))
+            } else {
+                glib::signal::Inhibit(false)
+            }
+        }
+    }
```
bilelmoussaoui, line 175: I think this should return glib::Inhibit but not very sure if I've understood that correctly from the C implementation. The corresponding C code is at https://gitlab.gnome.org/GNOME/gtk/-/blob/master/gtk/gtktextview.c#L7070

bilelmoussaoui:
As those two types have a bunch of vfuncs, I will keep the PR this small & work on others separately. Other than the mentioned issue I think the rest should be ok

gtk4/src/subclass/text_buffer.rs
```diff
@@ -0,0 +1,543 @@
...
+    end_ptr: *mut ffi::GtkTextIter,
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<TextBuffer> = from_glib_borrow(ptr);
+
+    imp.delete_range(
+        &wrap.unsafe_cast_ref(),
+        &mut from_glib_none(start_ptr),
+        &mut from_glib_none(end_ptr),
```
sdroege, line 403: This is not sufficient. You need to write back to the pointers afterwards if the iters have changed (also for other mutable text iters in the other functions).

In gtk3 I fixed that somewhere at some point, it's all a bit awful to get right. `gtk_text_iter_assign()` can be useful here.

sdroege: The problem basically is that `from_glib_none()` crates a copy, `from_glib_borrow()` is immutable and not a copy.

gtk4/src/subclass/text_view.rs
sdroege: In GTK4 this is fortunately a bit clearer:

> GDK_EVENT_STOP to stop other handlers from being invoked for the event. GDK_EVENT_PROPAGATE to propagate the event further.

That's exactly `Inhibit`. You can map to `glib::Inhibit` here or alternatively also create a new enum in gdk4 for that.

And you'll want to also properly map such bools to `Inhibit` for signal handlers in `src/auto` etc :)

bilelmoussaoui: I guess something like https://github.com/gtk-rs/gtk-rs/issues/91 would make sense here as well

gtk4/src/subclass/text_buffer.rs
bilelmoussaoui: Aha right, thanks, should be properly handled now

gtk4/src/subclass/text_view.rs
bilelmoussaoui: I will keep using glib::Inhibit for now here and opened https://github.com/gtk-rs/gtk4-rs/issues/88 to track and fix that issue properly

sdroege: Makes sense, yes

sdroege (approved):

PR #89, opened 2020-11-26
gtk: subclassing round 6

bilelmoussaoui:
Add subclassing support of  the following types:

- GtkColorChooser
- GtkComboBox
- GtkEntryBuffer
- GtkRange
- GtkScale
- GtkScaleButton

Part of #56

### DIFF
--- a/gtk4/src/subclass/color_chooser.rs
+++ b/gtk4/src/subclass/color_chooser.rs
@@ -1,0 +1,90 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::{ColorChooser, Orientation};
+use gdk::RGBA;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::Cast;
+
+pub trait ColorChooserImpl: ObjectImpl {
+    fn add_palette(
+        &self,
+        color_chooser: &Self::Type,
+        orientation: Orientation,
+        colors_per_line: i32,
+        colors: &[RGBA],
+    );
+    fn color_activated(&self, color_chooser: &Self::Type, rgba: RGBA);
+    fn get_rgba(&self, color_chooser: &Self::Type) -> RGBA;
+    fn set_rgba(&self, color_chooser: &Self::Type, rgba: RGBA);
+}
+
+unsafe impl<T: ColorChooserImpl> IsImplementable<T> for ColorChooser {
+    unsafe extern "C" fn interface_init(
+        iface: glib::ffi::gpointer,
+        _iface_data: glib::ffi::gpointer,
+    ) {
+        let color_chooser_iface = &mut *(iface as *mut ffi::GtkColorChooserInterface);
+
+        color_chooser_iface.add_palette = Some(color_chooser_add_palette::<T>);
+        color_chooser_iface.color_activated = Some(color_chooser_color_activated::<T>);
+        color_chooser_iface.get_rgba = Some(color_chooser_get_rgba::<T>);
+        color_chooser_iface.set_rgba = Some(color_chooser_set_rgba::<T>);
+    }
+}
+
+unsafe extern "C" fn color_chooser_add_palette<T: ColorChooserImpl>(
+    color_chooser: *mut ffi::GtkColorChooser,
+    orientation: ffi::GtkOrientation,
+    colors_per_line: i32,
+    total: i32,
+    colorsptr: *mut gdk::ffi::GdkRGBA,
+) {
+    let instance = &*(color_chooser as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    let colors = std::slice::from_raw_parts(colorsptr as *const RGBA, total as usize);
+    imp.add_palette(
+        from_glib_borrow::<_, ColorChooser>(color_chooser).unsafe_cast_ref(),
+        from_glib(orientation),
+        colors_per_line,
+        &colors,
+    );
+}
+
+unsafe extern "C" fn color_chooser_color_activated<T: ColorChooserImpl>(
+    color_chooser: *mut ffi::GtkColorChooser,
+    rgba: *const gdk::ffi::GdkRGBA,
+) {
+    let instance = &*(color_chooser as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    imp.color_activated(
+        from_glib_borrow::<_, ColorChooser>(color_chooser).unsafe_cast_ref(),
+        from_glib_none(rgba),
+    )
+}
+
+unsafe extern "C" fn color_chooser_get_rgba<T: ColorChooserImpl>(
+    color_chooser: *mut ffi::GtkColorChooser,
+    rgbaptr: *const gdk::ffi::GdkRGBA,
+) {
+    let instance = &*(color_chooser as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    let rgba = imp.get_rgba(from_glib_borrow::<_, ColorChooser>(color_chooser).unsafe_cast_ref());
+    *(rgbaptr as *mut gdk::ffi::GdkRGBA) = *rgba.to_glib_none().0;
+}
+
+unsafe extern "C" fn color_chooser_set_rgba<T: ColorChooserImpl>(
+    color_chooser: *mut ffi::GtkColorChooser,
+    rgba: *const gdk::ffi::GdkRGBA,
+) {
+    let instance = &*(color_chooser as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    imp.set_rgba(
+        from_glib_borrow::<_, ColorChooser>(color_chooser).unsafe_cast_ref(),
+        from_glib_none(rgba),
+    )
+}
--- a/gtk4/src/subclass/combo_box.rs
+++ b/gtk4/src/subclass/combo_box.rs
@@ -1,0 +1,80 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, GString};
+
+use super::widget::WidgetImpl;
+use crate::{ComboBox, Widget};
+
+pub trait ComboBoxImpl: ComboBoxImplExt + WidgetImpl {
+    fn changed(&self, combo_box: &Self::Type) {
+        self.parent_changed(combo_box)
+    }
+    fn format_entry_text(&self, combo_box: &Self::Type, path: &str) -> Option<GString> {
+        self.parent_format_entry_text(combo_box, path)
+    }
+}
+
+pub trait ComboBoxImplExt: ObjectSubclass {
+    fn parent_changed(&self, combo_box: &Self::Type);
+
+    fn parent_format_entry_text(&self, combo_box: &Self::Type, path: &str) -> Option<GString>;
+}
+
+impl<T: ComboBoxImpl> ComboBoxImplExt for T {
+    fn parent_changed(&self, combo_box: &Self::Type) {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkComboBoxClass;
+            if let Some(f) = (*parent_class).changed {
+                f(combo_box.unsafe_cast_ref::<ComboBox>().to_glib_none().0)
+            }
+        }
+    }
+    fn parent_format_entry_text(&self, combo_box: &Self::Type, path: &str) -> Option<GString> {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkComboBoxClass;
+            if let Some(f) = (*parent_class).format_entry_text {
+                return Some(from_glib_full(f(
+                    combo_box.unsafe_cast_ref::<ComboBox>().to_glib_none().0,
+                    path.to_glib_none().0,
+                )));
+            }
+            None
+        }
+    }
+}
+
+unsafe impl<T: ComboBoxImpl> IsSubclassable<T> for ComboBox {
+    fn override_vfuncs(class: &mut glib::Class<Self>) {
+        <Widget as IsSubclassable<T>>::override_vfuncs(class);
+
+        let klass = class.as_mut();
+        klass.changed = Some(combo_box_changed::<T>);
+        klass.format_entry_text = Some(combo_box_format_entry_text::<T>);
+    }
+}
+
+unsafe extern "C" fn combo_box_changed<T: ComboBoxImpl>(ptr: *mut ffi::GtkComboBox) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<ComboBox> = from_glib_borrow(ptr);
+
+    imp.changed(wrap.unsafe_cast_ref())
+}
+
+unsafe extern "C" fn combo_box_format_entry_text<T: ComboBoxImpl>(
+    ptr: *mut ffi::GtkComboBox,
+    pathptr: *const libc::c_char,
+) -> *mut libc::c_char {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<ComboBox> = from_glib_borrow(ptr);
+
+    let path: Borrowed<GString> = from_glib_borrow(pathptr);
+
+    imp.format_entry_text(wrap.unsafe_cast_ref(), &path.as_str())
+        .to_glib_full()
+}
--- a/gtk4/src/subclass/entry_buffer.rs
+++ b/gtk4/src/subclass/entry_buffer.rs
@@ -1,0 +1,332 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::EntryBuffer;
+use glib::subclass::object::ObjectImpl;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, GString, Object, ObjectExt};
+use once_cell::sync::Lazy;
+
+pub trait EntryBufferImpl: EntryBufferImplExt + ObjectImpl {
+    fn delete_text(&self, entry_buffer: &Self::Type, position: u32, n_chars: Option<u32>) -> u32 {
+        self.parent_delete_text(entry_buffer, position, n_chars)
+    }
+
+    fn deleted_text(&self, entry_buffer: &Self::Type, position: u32, n_chars: Option<u32>) {
+        self.parent_deleted_text(entry_buffer, position, n_chars)
+    }
+
+    fn get_length(&self, entry_buffer: &Self::Type) -> u32 {
+        self.parent_get_length(entry_buffer)
+    }
+
+    fn get_text(&self, entry_buffer: &Self::Type) -> GString {
+        self.parent_get_text(entry_buffer)
+    }
+    fn insert_text(&self, entry_buffer: &Self::Type, position: u32, chars: &str) -> u32 {
+        self.parent_insert_text(entry_buffer, position, chars)
+    }
+
+    fn inserted_text(&self, entry_buffer: &Self::Type, position: u32, chars: &str) {
+        self.parent_inserted_text(entry_buffer, position, chars)
+    }
+}
+
+pub trait EntryBufferImplExt: ObjectSubclass {
+    fn parent_delete_text(
+        &self,
+        entry_buffer: &Self::Type,
+        position: u32,
+        n_chars: Option<u32>,
+    ) -> u32;
+    fn parent_deleted_text(&self, entry_buffer: &Self::Type, position: u32, n_chars: Option<u32>);
+    fn parent_get_length(&self, entry_buffer: &Self::Type) -> u32;
+    fn parent_get_text(&self, entry_buffer: &Self::Type) -> GString;
+    fn parent_insert_text(&self, entry_buffer: &Self::Type, position: u32, chars: &str) -> u32;
+    fn parent_inserted_text(&self, entry_buffer: &Self::Type, position: u32, chars: &str);
+}
+
+impl<T: EntryBufferImpl> EntryBufferImplExt for T {
+    fn parent_delete_text(
+        &self,
+        entry_buffer: &Self::Type,
+        position: u32,
+        n_chars: Option<u32>,
+    ) -> u32 {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkEntryBufferClass;
+            let f = (*parent_class)
+                .delete_text
+                .expect("No parent class impl for \"delete_text\"");
+            f(
+                entry_buffer
+                    .unsafe_cast_ref::<EntryBuffer>()
+                    .to_glib_none()
+                    .0,
+                position,
+                n_chars.unwrap_or(u32::MAX),
+            )
+        }
+    }
+
+    fn parent_deleted_text(&self, entry_buffer: &Self::Type, position: u32, n_chars: Option<u32>) {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkEntryBufferClass;
+            if let Some(f) = (*parent_class).deleted_text {
+                f(
+                    entry_buffer
+                        .unsafe_cast_ref::<EntryBuffer>()
+                        .to_glib_none()
+                        .0,
+                    position,
+                    n_chars.unwrap_or(u32::MAX),
+                )
+            }
+        }
+    }
+
+    fn parent_get_length(&self, entry_buffer: &Self::Type) -> u32 {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkEntryBufferClass;
+            let f = (*parent_class)
+                .get_length
+                .expect("No parent class impl for \"get_length\"");
+            f(entry_buffer
+                .unsafe_cast_ref::<EntryBuffer>()
+                .to_glib_none()
+                .0)
+        }
+    }
+
+    fn parent_get_text(&self, entry_buffer: &Self::Type) -> GString {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkEntryBufferClass;
+            let f = (*parent_class)
+                .get_text
+                .expect("No parent class impl for \"get_text\"");
+            let mut n_bytes = 0;
+            let res = f(
+                entry_buffer
+                    .unsafe_cast_ref::<EntryBuffer>()
+                    .to_glib_none()
+                    .0,
+                &mut n_bytes,
+            );
+            FromGlibContainer::from_glib_none_num(res, n_bytes as usize)
+        }
+    }
+
+    fn parent_insert_text(&self, entry_buffer: &Self::Type, position: u32, text: &str) -> u32 {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkEntryBufferClass;
+            let f = (*parent_class)
+                .insert_text
+                .expect("No parent class impl for \"insert_text\"");
+
+            f(
+                entry_buffer
+                    .unsafe_cast_ref::<EntryBuffer>()
+                    .to_glib_none()
+                    .0,
+                position,
+                text.to_glib_none().0,
+                text.chars().count() as u32,
+            )
+        }
+    }
+
+    fn parent_inserted_text(&self, entry_buffer: &Self::Type, position: u32, text: &str) {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkEntryBufferClass;
+            if let Some(f) = (*parent_class).inserted_text {
+                f(
+                    entry_buffer
+                        .unsafe_cast_ref::<EntryBuffer>()
+                        .to_glib_none()
+                        .0,
+                    position,
+                    text.to_glib_none().0,
+                    text.chars().count() as u32,
+                )
+            }
+        }
+    }
+}
+
+unsafe impl<T: EntryBufferImpl> IsSubclassable<T> for EntryBuffer {
+    fn override_vfuncs(class: &mut glib::Class<Self>) {
+        <Object as IsSubclassable<T>>::override_vfuncs(class);
+
+        let klass = class.as_mut();
+        klass.delete_text = Some(entry_buffer_delete_text::<T>);
+        klass.deleted_text = Some(entry_buffer_deleted_text::<T>);
+        klass.get_length = Some(entry_buffer_get_length::<T>);
+        klass.get_text = Some(entry_buffer_get_text::<T>);
+        klass.insert_text = Some(entry_buffer_insert_text::<T>);
+        klass.inserted_text = Some(entry_buffer_inserted_text::<T>);
+    }
+}
+
+unsafe extern "C" fn entry_buffer_delete_text<T: EntryBufferImpl>(
+    ptr: *mut ffi::GtkEntryBuffer,
+    position: u32,
+    n_chars: u32,
+) -> u32 {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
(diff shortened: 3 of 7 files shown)


gtk4/src/subclass/color_chooser.rs
```diff
@@ -0,0 +1,94 @@
...
+    let imp = instance.get_impl();
+
+    let colors: Borrowed<Vec<RGBA>> = from_glib_borrow(colors);
+    imp.add_palette(
+        from_glib_borrow::<_, ColorChooser>(color_chooser).unsafe_cast_ref(),
+        from_glib(orientation),
+        colors_per_line,
+        &colors,
+    );
+    *total = *(colors.len() as i32);
```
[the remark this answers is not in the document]
sdroege: That's an interesting one. I'll get back to you after cooking :)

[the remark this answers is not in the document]
bilelmoussaoui: Enjoy your meal :)

gtk4/src/subclass/combo_box.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+    fn format_entry_text(&self, combo_box: &Self::Type, path: &str) -> Option<GString> {
+        self.parent_format_entry_text(combo_box, path)
+    }
+}
+
+pub trait ComboBoxImplExt: ObjectSubclass {
+    fn parent_changed(&self, combo_box: &Self::Type);
+
+    fn parent_format_entry_text(&self, combo_box: &Self::Type, path: &str) -> Option<GString>;
```
sdroege, line 20: Did you check if `path` is indeed not `nullable` and the return value is?

gtk4/src/subclass/color_chooser.rs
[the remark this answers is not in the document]
sdroege: > I wonder how can I turn the *mut gdk::ffi::GdkRGBA into a Vec<RGBA>

`FromGlibContainer::from_glib_none_num(colors, total as usize)` should do the job but that copies. As this is all read-only... `std::slice::from_raw_parts(colors as *const gdk4::RGBA, total)` would also work without copying.

> I also wonder if I should update the total value with the length of the array

No, you can't. It's not a pointer / out parameter, so it can't really be changed :) This is all read-only.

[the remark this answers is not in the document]
sdroege: Note that the cast only works because `RGBA` is the same as the C struct and `repr(C)` / `repr(transparent)`.

gtk4/src/subclass/color_chooser.rs
```diff
@@ -0,0 +1,94 @@
...
+unsafe extern "C" fn color_chooser_color_activated<T: ColorChooserImpl>(
+    color_chooser: *mut ffi::GtkColorChooser,
+    rgba: *const gdk::ffi::GdkRGBA,
+) {
+    let instance = &*(color_chooser as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    imp.color_activated(
+        from_glib_borrow::<_, ColorChooser>(color_chooser).unsafe_cast_ref(),
+        &from_glib_borrow(rgba),
```
sdroege, line 63: Strange that this is a pointer. Might want to pass an `RGBA` through Rust code as it's a `Copy` type

sdroege: I.e. make it not a reference in Rust

gtk4/src/subclass/color_chooser.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+unsafe extern "C" fn color_chooser_get_rgba<T: ColorChooserImpl>(
+    color_chooser: *mut ffi::GtkColorChooser,
+    rgba: *const gdk::ffi::GdkRGBA,
+) {
+    let instance = &*(color_chooser as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    if let Some(color) = imp.get_rgba(
```
sdroege, line 74: AFAIU this is not nullable and you must always always return some color here

gtk4/src/subclass/color_chooser.rs
```diff
@@ -0,0 +1,94 @@
...
+    rgba: *const gdk::ffi::GdkRGBA,
+) {
+    let instance = &*(color_chooser as *mut T::Instance);
+    let imp = instance.get_impl();
+    
+    let color: Borrowed<Option<RGBA>> = from_glib_borrow(rgba);
+
+    imp.set_rgba(
+        from_glib_borrow::<_, ColorChooser>(color_chooser).unsafe_cast_ref(),
+        color.as_ref().as_ref(),
```
sdroege, line 92: Also this does not seem to be nullable, and also here you probably want to pass a non-reference `RGBA` through Rust code.

gtk4/src/subclass/color_chooser.rs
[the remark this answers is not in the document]
bilelmoussaoui: Thanks for the detailed informations! :)

gtk4/src/subclass/color_chooser.rs
bilelmoussaoui: Ah. I guess we need to update the auto color_chooser as well.

gtk4/src/subclass/combo_box.rs
bilelmoussaoui: The return value can be null if the ComboBox doesn't set a `entry-text-column` property. Although I didn't find any examples of usages with NULL passed as `path`.

gtk4/src/subclass/color_chooser.rs
bilelmoussaoui: Just checked the C implementation, indeed this doesn't return a nullable

gtk4/src/subclass/color_chooser.rs
```diff
@@ -0,0 +1,87 @@
...
+
+unsafe extern "C" fn color_chooser_get_rgba<T: ColorChooserImpl>(
+    color_chooser: *mut ffi::GtkColorChooser,
+    _rgba: *const gdk::ffi::GdkRGBA,
+) {
+    let instance = &*(color_chooser as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    imp.get_rgba(from_glib_borrow::<_, ColorChooser>(color_chooser).unsafe_cast_ref());
+}
```
bilelmoussaoui, line 74: I'm not very sure about this implementation at all. As the parent impl returns an RGBA; but what i'm supposed to do with it? :)

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,233 @@
...
+
+unsafe extern "C" fn entry_buffer_get_text<T: EntryBufferImpl>(
+    ptr: *mut ffi::GtkEntryBuffer,
+    n_bytes: *mut usize,
+) -> *const libc::c_char {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<EntryBuffer> = from_glib_borrow(ptr);
+
+    imp.get_text(wrap.unsafe_cast_ref(), &mut (n_bytes as usize))
```
bilelmoussaoui, line 193: Not sure if `n_bytes as usize` is right way to do cast my pointer to a usize?

gtk4/src/subclass/combo_box.rs
sdroege: Ok, worst case it will panic and can be fixed then :)

gtk4/src/subclass/color_chooser.rs
sdroege: Quite a few options :)

```rust
let res = imp.get_rgba(...);
*(rgba as *mut gdk4::RGBA) = rgba;
```

Seems the easiest.

Not sure why the parameter here is `*const GdkRGBA`. It's not `const` in C so this looks like a code generation bug. But you can work around it (create an issue though) with some casting as for raw pointers `const`/`mut` have no meaning for the compiler.

```rust
let res = imp.get_rgba(...);
*(rgba as *mut _ as *mut gdk4::RGBA) = rgba;
```

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    fn parent_get_text(&self, entry_buffer: &Self::Type, n_bytes: &mut usize) -> GString {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkEntryBufferClass;
+            let f = (*parent_class)
+                .get_text
+                .expect("No parent class impl for \"get_text\"");
+            from_glib_none(f(
```
sdroege, line 98: Use `FromGlibContainer::from_glib_none_num()` here.

```rust
let mut n_bytes = 0;
let res = f(..., &n_bytes);
FromGlibContainer::from_glib_none_num(res, n_bytes as usize)
```

And only return a `GString` and get rid of the `&mut` parameter. It's for the length of the text AFAIU, and that's part of the string anyway...

gtk4/src/subclass/entry_buffer.rs
sdroege: No, see above. You'd get back a `GString` from `get_text()` and then return the pointer to it and store the length in `n_bytes`.

It's an out parameter for the length of the string. What you do right now is unsound and you also pass an uninitialized value around :)

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,233 @@
...
+    fn inserted_text(&self, entry_buffer: &Self::Type, position: u32, chars: &str) {
+        self.parent_inserted_text(entry_buffer, position, chars)
+    }
+}
+
+pub trait EntryBufferImplExt: ObjectSubclass {
+    fn parent_delete_text(&self, entry_buffer: &Self::Type, position: u32, n_chars: u32) -> u32;
+    fn parent_deleted_text(&self, entry_buffer: &Self::Type, position: u32, n_chars: u32);
+    fn parent_get_length(&self, entry_buffer: &Self::Type) -> u32;
+    fn parent_get_text(&self, entry_buffer: &Self::Type, n_bytes: &mut usize) -> GString;
```
sdroege, line 36: This is all a bit suboptimal as you'd always create multiple copies of the string here on the way out. This needs to be refactored a bit later (create an issue please).

The idea of the C class is that you return a plain borrowed reference to the string buffer here without any copying, but I'm not entirely sure how that can be safely done here. Typical C-isms in GTK like all those `transfer none` return values there and praying that nothing explodes.

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,233 @@
...
+    position: u32,
+    charsptr: *const libc::c_char,
+    _length: u32,
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<EntryBuffer> = from_glib_borrow(ptr);
+    let chars: Borrowed<GString> = from_glib_borrow(charsptr);
+
+    imp.inserted_text(wrap.unsafe_cast_ref(), position, &chars)
```
sdroege, line 232: If `length == -1` then this is correct, otherwise you need to only pass the first `length` bytes (or characters? the docs are not exactly clear if they mean "character" or C `char` there...)

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,233 @@
...
+    let wrap: Borrowed<EntryBuffer> = from_glib_borrow(ptr);
+
+    imp.get_length(wrap.unsafe_cast_ref())
+}
+
+unsafe extern "C" fn entry_buffer_insert_text<T: EntryBufferImpl>(
+    ptr: *mut ffi::GtkEntryBuffer,
+    position: u32,
+    charsptr: *const libc::c_char,
+    _n_chars: u32,
```
sdroege, line 211: Same here and probably elsewhere

gtk4/src/subclass/range.rs
```diff
@@ -0,0 +1,179 @@
...
+                .expect("No parent class impl for \"change_value\"");
+            from_glib(f(
+                range.unsafe_cast_ref::<Range>().to_glib_none().0,
+                scroll_type.to_glib(),
+                new_value,
+            ))
+        }
+    }
+
+    fn parent_get_range_border(&self, range: &Self::Type) -> Option<Border> {
```
sdroege, line 74: I don't think this is nullable

gtk4/src/subclass/scale.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+impl<T: ScaleImpl> ScaleImplExt for T {
+    fn parent_get_layout_offsets(&self, scale: &Self::Type) -> Option<(i32, i32)> {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkScaleClass;
+            if let Some(f) = (*parent_class).get_layout_offsets {
+                let x = std::ptr::null_mut();
+                let y = std::ptr::null_mut();
```
sdroege, line 25: That does not work like this. You pass `NULL` to the C function and it can't modify that as you don't pass a pointer to anything it could modify.

```rust
let mut x = 0;
let mut y = 0;
f(..., &mut x, &mut y);
```

And then this should probably also not be nullable

gtk4/src/subclass/scale.rs
```diff
@@ -0,0 +1,61 @@
...
+unsafe extern "C" fn scale_get_layout_offsets<T: ScaleImpl>(
+    ptr: *mut ffi::GtkScale,
+    x_ptr: *mut libc::c_int,
+    y_ptr: *mut libc::c_int,
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<Scale> = from_glib_borrow(ptr);
+
+    if let Some((x, y)) = imp.get_layout_offsets(wrap.unsafe_cast_ref()) {
```
sdroege, line 54: Not nullable but you can always assign to the `x_ptr` / `y_ptr`. They're not optional from what I see in the C code.

bilelmoussaoui: Per the docs 
>  * If the #GtkScale:draw-value property is %FALSE, the return values are undefined.
So I'm not sure if we can always to the x_ptr/y_ptr?

sdroege:
```cpp
void 
gtk_scale_get_layout_offsets (GtkScale *scale,
                              int      *x,
                              int      *y)
{
  int local_x = 0; 
  int local_y = 0;

  g_return_if_fail (GTK_IS_SCALE (scale));

  if (GTK_SCALE_GET_CLASS (scale)->get_layout_offsets)
    (GTK_SCALE_GET_CLASS (scale)->get_layout_offsets) (scale, &local_x, &local_y);

  if (x)
    *x = local_x;
  
  if (y)
    *y = local_y;
}
```

is the only caller, and that assumes that some value is always returned and it always passes non-`NULL` pointers to the vfunc

sdroege: That also means that the function is optional and `0`/`0` is the default, btw.

bilelmoussaoui: I somehow completely missed the ` if (GTK_SCALE_GET_CLASS (scale)->get_layout_offsets)`, thanks!

gtk4/src/subclass/color_chooser.rs
bilelmoussaoui: Opened an issue in https://github.com/gtk-rs/gir/issues/1012. Thanks!

gtk4/src/subclass/entry_buffer.rs
bilelmoussaoui: This doesn't seem to work the trait `glib::translate::FromGlibContainer<_, *const i8>` is not implemented for `glib::GString`

sdroege: Then you need to implement it :) The problem is that you must only create a string for given length here, not the whole thing until the NUL terminator

gtk4/src/subclass/entry_buffer.rs
bilelmoussaoui: Seems to be the number of characters per 
>  * Note that @n_chars is in characters, not in bytes.

sdroege: That doesn't make it easier to handle, but at least `std::str` has useful unicode API.

What does characters mean though? Same as a Rust `char`?

bilelmoussaoui: I guess so, because it falls back to `g_utf8_strlen` in case the provided length is equal `< 0`

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,242 @@
...
+                .expect("No parent class impl for \"get_text\"");
+            let mut n_bytes = 0;
+            let res = f(
+                entry_buffer
+                    .unsafe_cast_ref::<EntryBuffer>()
+                    .to_glib_none()
+                    .0,
+                &mut n_bytes,
+            );
+            FromGlibContainer::from_glib_none_num(res, n_bytes as usize)
```
sdroege, line 106: This needs some more work AFAIU. What you get back from the C function is the number of *characters*, but for creating the sub-slice we need to number of bytes (you correctly name it `n_bytes` here).

You need to create a `&str` from the value returned by C (making use of the `NUL` terminator to never read too much) and then use the unicode APIs to know what byte position corresponds to the n-th character.

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,242 @@
...
+                .expect("No parent class impl for \"insert_text\"");
+
+            f(
+                entry_buffer
+                    .unsafe_cast_ref::<EntryBuffer>()
+                    .to_glib_none()
+                    .0,
+                position,
+                chars.to_glib_none().0,
+                chars.len() as u32,
```
sdroege, line 125: Bytes / characters mismatch here too. `len()` is in bytes

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,242 @@
...
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkEntryBufferClass;
+            if let Some(f) = (*parent_class).inserted_text {
+                f(
+                    entry_buffer
+                        .unsafe_cast_ref::<EntryBuffer>()
+                        .to_glib_none()
+                        .0,
+                    position,
+                    chars.to_glib_none().0,
+                    chars.len() as u32,
```
sdroege, line 142: And here

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,242 @@
...
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<EntryBuffer> = from_glib_borrow(ptr);
+
+    imp.deleted_text(wrap.unsafe_cast_ref(), position, n_chars)
+}
+
+unsafe extern "C" fn entry_buffer_get_text<T: EntryBufferImpl>(
+    ptr: *mut ffi::GtkEntryBuffer,
+    _n_bytes: *mut usize,
```
sdroege, line 189: You have to return the number of bytes here. Interestingly this is not the number of characters but bytes.

Please double-check above for all my comments if the C API actually works with bytes or characters there...

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,242 @@
...
+) -> *const libc::c_char {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<EntryBuffer> = from_glib_borrow(ptr);
+
+    imp.get_text(wrap.unsafe_cast_ref()).to_glib_full()
```
sdroege, line 195: Welcome to stupid C APIs :) The C function is `transfer none` in its return value. That means that you are currently creating a new copy of the string and then leaking it, which is not optimal.

You could store the `to_glib_full()` return value in `g_object_set_qdata_full()` (use `g_free()` as free function) and replace that on every call. Then it's at least kept valid until the next call to `get_text()`, which should be sufficient here.

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,242 @@
...
+) -> u32 {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<EntryBuffer> = from_glib_borrow(ptr);
+    let chars: Borrowed<GString> = from_glib_borrow(charsptr);
+
+    imp.insert_text(
+        wrap.unsafe_cast_ref(),
+        position,
+        &chars[0..(n_chars as usize)],
```
sdroege, line 222: Indexing `&str` works in bytes, not characters. You need to use the unicode APIs here.

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,242 @@
...
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<EntryBuffer> = from_glib_borrow(ptr);
+    let chars: Borrowed<GString> = from_glib_borrow(charsptr);
+
+    imp.inserted_text(
+        wrap.unsafe_cast_ref(),
+        position,
+        &chars[0..(length as usize)],
```
sdroege, line 240: Is this in bytes or chars? See above if it's in chars

gtk4/src/subclass/entry_buffer.rs
sdroege: I think there's an example of this pattern in `gio` in `IOStream` for the getters of the input/output stream.

gtk4/src/subclass/entry_buffer.rs
bilelmoussaoui: It's in chars, the last push was to be able to switch branches for #94. I will get back to this one once the glib one is in

gtk4/src/subclass/entry_buffer.rs
bilelmoussaoui: > You could store the `to_glib_full()` return value in `g_object_set_qdata_full()` (use `g_free()` as free function) and replace that on every call. Then it's at least kept valid until the next call to `get_text()`, which should be sufficient here.

We already have safe bindings for `set_qdata` =p

gtk4/src/subclass/entry_buffer.rs
bilelmoussaoui: Strangely, this method returns the number of bytes and not the number of characters like the `insert|delete{_ed}_text` ones

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,256 @@
...
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<EntryBuffer> = from_glib_borrow(ptr);
+    let text: Borrowed<GString> = from_glib_borrow(charsptr);
+
+    let pos_end = text
+        .char_indices()
+        .map(|(i, _char)| i)
+        .nth(n_chars as usize)
+        .unwrap();
```
bilelmoussaoui, line 233: I'm not very sure about the unwrap here, should we assume that the returned text's characters length > n_chars here?

gtk4/src/subclass/entry_buffer.rs
sdroege: Might be worth reporting this inconsistency to the GTK people :)

gtk4/src/subclass/entry_buffer.rs
sdroege: If you check the docs, `n_chars` can also be `-1`. So that needs to be handled at least and means "all characters".

Otherwise I don't know, seems like a bug if there are fewer characters available so panic?

sdroege: Also check if any of the other n_bytes/n_chars can be -1 please

bilelmoussaoui: They all do except the n_bytes one, I wonder how the delete_text|deleted_text should be handled ? I have also a question regarding this one: it seems the generated code uses an unsigned integer, can I just cast as a signed one to check if it's negative?

sdroege: Yes or check against `u32::MAX` (that's the same as `-1i32`).

For these cases I usually make it an `Option<u32>` on the Rust side and make the `-1` to `None`.

sdroege:
Is this ready for another review or not yet? :)

bilelmoussaoui:
> Is this ready for another review or not yet? :)

It should be :)

gtk4/src/subclass/entry_buffer.rs
sdroege: Too late for GTK4 at this point

sdroege:
Can you resolve all the discussions that are handled?

gtk4/src/subclass/entry_buffer.rs
bilelmoussaoui: I actually asked mclassen about this one before gtk4 release, it's intended for it to be n_bytes & not n_chars

gtk4/src/subclass/entry_buffer.rs
bilelmoussaoui: I think this one is handled as well now?

bilelmoussaoui:
> Can you resolve all the discussions that are handled?

Should be the case now :)

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,284 @@
...
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<EntryBuffer> = from_glib_borrow(ptr);
+    let text: Borrowed<GString> = from_glib_borrow(charsptr);
+
+    let mut char_indices = text.char_indices().map(|(i, _char)| i);
+    let pos_end = if n_chars != u32::MAX {
+        char_indices.nth(n_chars as usize)
+    } else {
+        char_indices.last()
```
sdroege, line 257: This gives you the start byte of the last character. What you actually want here is the byte after the end of the last character, or otherwise your `text[0..pos_end]` below cuts the last character

sdroege: And the other case probably also needs a `+1`

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,284 @@
...
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<EntryBuffer> = from_glib_borrow(ptr);
+    let text: Borrowed<GString> = from_glib_borrow(charsptr);
+
+    let mut char_indices = text.char_indices().map(|(i, _char)| i);
+    let pos_end = if length != u32::MAX {
+        char_indices.nth(length as usize)
+    } else {
+        char_indices.last()
```
sdroege, line 279: Same

gtk4/src/subclass/entry_buffer.rs
sdroege: Well it's still suboptimal, there are many copies here :) One for the `to_glib_full()`, one for getting the return value from the parent class' impl, another for getting the value from the Rust impl.

gtk4/src/subclass/entry_buffer.rs
sdroege: Now you increase the index in bytes, not characters :) So you'd end up in the middle of a character if you're unlucky.

bilelmoussaoui: I have did another attempt at it. I think it's probably correct :)

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,286 @@
...
+    n_chars: u32,
+) -> u32 {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<EntryBuffer> = from_glib_borrow(ptr);
+    let text: Borrowed<GString> = from_glib_borrow(charsptr);
+
+    let chars =
+        if n_chars != u32::MAX {
+            let pos_end = text.char_indices().take(n_chars as usize).next().expect(
```
sdroege, line 255: I think this is almost correct.

```suggestion
            let mut iter = text.char_indices().skip(n_chars as usize);
            iter.next().expect("blabla"); // This would fail if there are no `n_chars` characters
            let pos_end = iter.next().unwrap_or_else(|| text.len()); // If `n_chars` is the number of characters then there is no next character and you should take the whole string
```

sdroege: Same in the other places

sdroege (commented):
Almost :)

gtk4/src/subclass/entry_buffer.rs
sdroege: Why didn't you just paste the code I wrote above? :) Now it's wrong. `pos_end` is again the byte index of the start of the last character. But you want the byte index of the start of the next character (or the length of the whole bytes slice).

Note that I called `next()` twice above and have `unwrap_or_else()` in there too.

sdroege: This code is apparently so complicated that it needs a test :) Maybe refactor it into a helper function in this file and write a few tests around it with some mixed ASCII/UTF-8 strings.

bilelmoussaoui: Makes sense; I will do that

bilelmoussaoui: > Why didn't you just paste the code I wrote above? :) Now it's wrong. `pos_end` is again the byte index of the start of the last character. But you want the byte index of the start of the next character (or the length of the whole bytes slice).
> 
> Note that I called `next()` twice above and have `unwrap_or_else()` in there too.

The double .next() wasn't giving the expected results, see the results of the tests I have added
```
thread 'subclass::entry_buffer::test::n_chars_max_length_ascii' panicked at 'assertion failed: `(left == right)`
  left: `"gtk-rs "`,
 right: `"gtk-rs"`', gtk4/src/subclass/entry_buffer.rs:295:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- subclass::entry_buffer::test::n_chars_max_length_utf8 stdout ----
thread 'subclass::entry_buffer::test::n_chars_max_length_utf8' panicked at 'assertion failed: `(left == right)`
  left: `"👨👩👧"`,
 right: `"👨👩"`', gtk4/src/subclass/entry_buffer.rs:310:9

---- subclass::entry_buffer::test::n_chars_max_length_utf8_ascii stdout ----
thread 'subclass::entry_buffer::test::n_chars_max_length_utf8_ascii' panicked at 'assertion failed: `(left == right)`
  left: `"👨g👩"`,
 right: `"👨g"`', gtk4/src/subclass/entry_buffer.rs:317:9

```
As you can see, there's always an extra character than the expected result

sdroege: Indeed, the skip always skips over that so one more is not needed.

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -0,0 +1,326 @@
...
+    if n_chars != u32::MAX {
+        let mut iter = text
+            .char_indices()
+            .skip(n_chars as usize)
+            .map(|(pos, _)| pos);
+        let pos_end = iter
+            .next()
+            .expect(
+                "The passed text to EntryBuffer contains fewer characters than what's passed as a length",
+            );
```
sdroege, line 282:
```suggestion
            .map(|(idx, _char)| idx)
            .unwrap_or_else(|| text.len());
```

Otherwise this fails: `assert_eq!(get_text_n_chars("👨👩👧👦", 4), "👨👩👧👦");`

However that would mean that we always take the whole string if it has fewer characters too. I guess we need to skip `n_char - 1` instead and then call `next()` once to check that the character exists and another time to go behind it with `unwrap_or_else()`.

bilelmoussaoui: The `n_chars - 1` did the trick

sdroege: Now you only need to also guard against `n_chars == 0`, which would wrap around now and cause problems :)

Otherwise looks good now

sdroege:
> This branch has conflicts that must be resolved

sdroege (approved):

PR #91, opened 2020-11-27
gtk: subclassing round 7

bilelmoussaoui:
Add subclassing support of the following types:
- GtkMediaStream
- GtkShortcutManager
- GtkFrame
- GtkGLArea

part of #56

### DIFF
--- a/gtk4/src/subclass/frame.rs
+++ b/gtk4/src/subclass/frame.rs
@@ -1,0 +1,55 @@
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::Cast;
+
+use super::widget::WidgetImpl;
+use crate::{Allocation, Frame, Widget};
+
+pub trait FrameImpl: FrameImplExt + WidgetImpl {
+    fn compute_child_allocation(&self, frame: &Self::Type) -> Allocation {
+        self.parent_compute_child_allocation(frame)
+    }
+}
+
+pub trait FrameImplExt: ObjectSubclass {
+    fn parent_compute_child_allocation(&self, frame: &Self::Type) -> Allocation;
+}
+
+impl<T: FrameImpl> FrameImplExt for T {
+    fn parent_compute_child_allocation(&self, frame: &Self::Type) -> Allocation {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkFrameClass;
+            let f = (*parent_class)
+                .compute_child_allocation
+                .expect("No parent class impl for \"compute_child_allocation\"");
+            let mut allocation = Allocation::uninitialized();
+            f(
+                frame.unsafe_cast_ref::<Frame>().to_glib_none().0,
+                allocation.to_glib_none_mut().0,
+            );
+            allocation
+        }
+    }
+}
+
+unsafe impl<T: FrameImpl> IsSubclassable<T> for Frame {
+    fn override_vfuncs(class: &mut glib::Class<Self>) {
+        <Widget as IsSubclassable<T>>::override_vfuncs(class);
+
+        let klass = class.as_mut();
+        klass.compute_child_allocation = Some(frame_compute_child_allocation::<T>);
+    }
+}
+
+unsafe extern "C" fn frame_compute_child_allocation<T: FrameImpl>(
+    ptr: *mut ffi::GtkFrame,
+    allocationptr: *mut ffi::GtkAllocation,
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<Frame> = from_glib_borrow(ptr);
+
+    let allocation = imp.compute_child_allocation(wrap.unsafe_cast_ref());
+    *allocationptr = *allocation.to_glib_none().0;
+}
--- a/gtk4/src/subclass/gl_area.rs
+++ b/gtk4/src/subclass/gl_area.rs
@@ -1,0 +1,116 @@
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::Cast;
+
+use super::widget::WidgetImpl;
+use crate::{GLArea, Widget};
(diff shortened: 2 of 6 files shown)


gtk4/src/subclass/frame.rs
```diff
@@ -0,0 +1,52 @@
...
+
+unsafe extern "C" fn frame_compute_child_allocation<T: FrameImpl>(
+    ptr: *mut ffi::GtkFrame,
+    allocationptr: *mut ffi::GtkAllocation,
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<Frame> = from_glib_borrow(ptr);
+
+    imp.compute_child_allocation(wrap.unsafe_cast_ref(), &mut from_glib_none(allocationptr));
```
sdroege, line 51: Is there always an initialized allocation passed in here by the C callers? Or are we supposed to create one?

gtk4/src/subclass/gl_area.rs
```diff
@@ -0,0 +1,116 @@
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::Cast;
+
+use super::widget::WidgetImpl;
+use crate::{GLArea, Widget};
+use gdk::GLContext;
+
+pub trait GLAreaImpl: GLAreaImplExt + WidgetImpl {
```
sdroege, line 9: Nice, now you can implement GStreamer's `gtkglsink` in Rust :)

gtk4/src/subclass/media_file.rs
```diff
@@ -42,7 +42,7 @@ impl<T: MediaFileImpl> MediaFileImplExt for T {
 
 unsafe impl<T: MediaFileImpl> IsSubclassable<T> for MediaFile {
     fn override_vfuncs(class: &mut glib::Class<Self>) {
-        <Object as IsSubclassable<T>>::override_vfuncs(class);
+        <MediaStream as IsSubclassable<T>>::override_vfuncs(class);
```
sdroege, line 45: Oops

sdroege:
Looks good to me otherwise

gtk4/src/subclass/frame.rs
bilelmoussaoui: GtkFrame is default implementation directly assign the values to the `Allocation` 
```c
static void
gtk_frame_real_compute_child_allocation (GtkFrame      *frame,
                                         GtkAllocation *child_allocation)
{
  GtkFramePrivate *priv = gtk_frame_get_instance_private (frame);
  int frame_width, frame_height;
  int height;

  frame_width = gtk_widget_get_width (GTK_WIDGET (frame));
  frame_height = gtk_widget_get_height (GTK_WIDGET (frame));

  if (priv->label_widget)
    {
      int nat_width, width;

      gtk_widget_measure (priv->label_widget, GTK_ORIENTATION_HORIZONTAL, -1,
                          NULL, &nat_width, NULL, NULL);
      width = MIN (frame_width, nat_width);
      gtk_widget_measure (priv->label_widget, GTK_ORIENTATION_VERTICAL, width,
                          &height, NULL, NULL, NULL);
    }
  else
    height = 0;

  child_allocation->x = 0;
  child_allocation->y = height;
  child_allocation->width = MAX (1, frame_width);
  child_allocation->height = MAX (1, frame_height - height);
}
```
So I guess the passed value is always going to be an initialized allocation unless I'm misunderstanding something

sdroege: It's allocated memory but it might have random uninitialized values inside it. It's UB in Rust to have a reference to uninitialized memory (a raw pointer is fine).

If the callers in C always initialize the allocation with something reasonable then what you do is fine. Otherwise you need to change it so that an `Allocation` is returned and directly stored in the pointer.

What can also help to understand what is the case here is to check if there are implementations of the virtual method that only read and adjust the allocation passed in. If such implementations exist then your current approach is correct.

bilelmoussaoui: The thing is that I can't find anything that uses it in any parts of the code other in gtkframe.c itself which makes understanding what to expect harder. I guess creating an empty allocation with (0, 0, 1, 1) as initial values and passing it to function would work to avoid some random values?

sdroege:
```cpp
static void
gtk_frame_size_allocate (GtkWidget *widget,
                         int        width,
                         int        height,
                         int        baseline)
{
  GtkFrame *frame = GTK_FRAME (widget);
  GtkFramePrivate *priv = gtk_frame_get_instance_private (frame);
  GtkAllocation new_allocation;   <----------- uninitialized

  gtk_frame_compute_child_allocation (frame, &new_allocation);
```

So I think this should simply return an `Allocation`, not pass mutable references around.

bilelmoussaoui: That actually makes more sense. Thanks a lot for the help! should be ready now

gtk4/src/subclass/frame.rs
```diff
@@ -0,0 +1,55 @@
...
+
+impl<T: FrameImpl> FrameImplExt for T {
+    fn parent_compute_child_allocation(&self, frame: &Self::Type) -> Allocation {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkFrameClass;
+            let f = (*parent_class)
+                .compute_child_allocation
+                .expect("No parent class impl for \"compute_child_allocation\"");
+            let mut allocation = Allocation::uninitialized();
```
sdroege, line 26: Not a fan of this and we want to get rid of `uninitialiezd()` in the long term.

```rust
let mut allocation = std::mem::MaybeUninit::uninit();
f(..., allocation.as_mut_ptr());
allocation.assume_init()
```

sdroege: Hmm that probably does not work here. The last would have to be `Allocation(allocation.assume_init())` or something like that.

Maybe something for another time...

PR #94, opened 2020-11-29
Examples: add video player/entry 

bilelmoussaoui:
port two sample examples from C

1 - video player: I decided to use a composite template to avoid having a bunch of Rc<T>

2 - entry that exposes the new history manager on gtktextview/gtkeditable

### DIFF
--- a/examples/src/bin/video_player.rs
+++ b/examples/src/bin/video_player.rs
@@ -1,0 +1,126 @@
+//! # Video Player
+//!
+//! This sample demonstrates how to use GtkVideo to play videos
+
+use glib::clone;
+use glib::subclass::prelude::*;
+use gtk::{gio, glib, prelude::*, CompositeTemplate};
+
+mod imp {
+    use super::*;
+    use glib::subclass;
+    use gtk::subclass::prelude::*;
+
+    #[derive(Debug, CompositeTemplate)]
+    pub struct VideoPlayerWindow {
+        #[template_child(id = "video")]
+        pub video: TemplateChild<gtk::Video>,
+        pub dialog: gtk::FileChooserNative,
+    }
+
+    impl ObjectSubclass for VideoPlayerWindow {
+        const NAME: &'static str = "VideoPlayerWindow";
+        type Type = super::VideoPlayerWindow;
+        type ParentType = gtk::ApplicationWindow;
+        type Instance = subclass::simple::InstanceStruct<Self>;
+        type Class = subclass::simple::ClassStruct<Self>;
+
+        glib::object_subclass!();
+
+        fn new() -> Self {
+            let dialog = gtk::FileChooserNative::new(
+                Some("Open File"),
+                gtk::NONE_WINDOW,
+                gtk::FileChooserAction::Open,
+                Some("Open"),
+                Some("Cancel"),
+            );
+            dialog.set_modal(true);
+
+            let videos_filter = gtk::FileFilter::new();
+            videos_filter.add_mime_type("video/*");
+            videos_filter.set_name(Some("Video"));
+            dialog.add_filter(&videos_filter);
+
+            let audio_filter = gtk::FileFilter::new();
+            audio_filter.add_mime_type("audio/*");
+            audio_filter.set_name(Some("Audio"));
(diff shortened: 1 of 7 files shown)


examples/src/bin/video_player.rs
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    fn init_actions(&self) {
+        let open = gio::SimpleAction::new("open", None);
+        open.connect_activate(clone!(@weak self as win => move |_, _| {
+            let dialog = gtk::FileChooserDialog::new(Some("Open Video"), Some(&win), gtk::FileChooserAction::Open, &[("Open", gtk::ResponseType::Ok), ("Cancel", gtk::ResponseType::Cancel)]);
+            dialog.set_modal(true);
+
+            let videos_filter = gtk::FileFilter::new();
+            videos_filter.add_mime_type("video/webm");
```
alatiera, line 80: You can use a wildcard with GtkFilter for the subtype.

```suggestion
            videos_filter.add_mime_type("video/*");
```

bilelmoussaoui: Aha, thanks for the hint!

sdroege: Maybe you also want to allow `audio/*`?

examples/src/bin/video_player.rs
```diff
@@ -0,0 +1,116 @@
...
+            videos_filter.set_name(Some("Video"));
+
+            dialog.add_filter(&videos_filter);
+
+            dialog.connect_response(move |d, response| {
+                if response == gtk::ResponseType::Ok {
+                    win.set_video(d.get_file().unwrap());
```
sdroege, line 87: You're moving a strong reference to the window into the dialog's signal handler here. Probably not much of a problem as `d.close()` should get rid of it again, but something to keep in mind here.

sdroege:
> 1 - video player: I decided to use a composite template to avoid having a bunch of Rc

`OnceCell` would've been enough, but this is much nicer indeed :)

examples/src/bin/video_player.rs
```diff
@@ -0,0 +1,116 @@
...
+        glib::Object::new(Self::static_type(), &[("application", app)])
+            .expect("Failed to create VideoPlayerWindow")
+            .downcast::<VideoPlayerWindow>()
+            .expect("Created object is of wrong type")
+    }
+
+    fn init_actions(&self) {
+        let open = gio::SimpleAction::new("open", None);
+        open.connect_activate(clone!(@weak self as win => move |_, _| {
+            let dialog = gtk::FileChooserDialog::new(Some("Open Video"), Some(&win), gtk::FileChooserAction::Open, &[("Open", gtk::ResponseType::Ok), ("Cancel", gtk::ResponseType::Cancel)]);
```
alatiera, line 76: Could we use a FileChooserNative instead?

bilelmoussaoui: I've already tried but somehow the filechoosernative fails silently and instantly, I haven’t got around debugging that yet

examples/src/bin/video_player.rs
```diff
@@ -0,0 +1,116 @@
...
+
+fn main() {
+    let application = gtk::Application::new(
+        Some("com.github.gtk-rs.examples.video_player"),
+        Default::default(),
+    )
+    .expect("Failed to initialize application");
+
+    application.connect_activate(|app| {
+        let win = VideoPlayerWindow::new(app);
```
alatiera, line 111: We create a window here and we dropped it immediately once the scope ends, what keeps the window around after?

sdroege: The application

alatiera: oh right, its an application window not normal window, nevermind me

examples/src/bin/video_player.rs
bilelmoussaoui: I forgot that I have to hold a reference of the FileChooserNative because nothing does from gtk side. Should be fixed

bilelmoussaoui:
@alatiera any other concerns regarding this one? Otherwise I will go ahead and merge it

alatiera:
Looks good!

PR #96, opened 2020-11-30
gtk: fix compute_expand implementation

### LINKED ISSUE #95: Panic when parent's Widget::compute_expand() virtual method is not implemented in subclass
When hexpand is used in sub classed object like ListBoxRow it throws
```
thread 'main' panicked at 'No parent class impl for "compute_expand"', /home/kavan/Documents/App/_build/build/target/cargo-home/git/checkouts/gtk4-rs-e74ad56283dfeb5e/78f7523/gtk4/src/subclass/widget.rs:186:18
```

page.rs
```
use glib::subclass;
use glib::{clone, glib_object_subclass, glib_wrapper};

use gtk::prelude::*;
use gtk::subclass::prelude::*;
use gtk::subclass::{list_box_row::ListBoxRowImpl, widget::WidgetImpl};

use std::cell::RefCell;
use std::error::Error;

use crate::GRESOURCE_ID;

#[derive(Debug, Clone)]
pub struct Page_ {
    id: RefCell<i64>,
    name: RefCell<String>,
    builder: gtk::Builder,
}

impl ObjectSubclass for Page_ {
    const NAME: &'static str = "Page";
    type Type = Page;
    type ParentType = gtk::ListBoxRow;
    type Instance = subclass::simple::InstanceStruct<Self>;
    type Class = subclass::simple::ClassStruct<Self>;

    glib_object_subclass!();

    fn new() -> Self {
        Self {
            id: RefCell::new(0),
            name: RefCell::new(String::new()),
            builder: gtk::Builder::from_resource(&(GRESOURCE_ID.to_owned() + "/page-row.ui")),
        }
    }
}

impl ObjectImpl for Page_ {
    fn constructed(&self, obj: &Self::Type) {
        self.parent_constructed(obj);

        get_widget!(self.builder, gtk::Box, root);
        obj.set_child(Some(&root));
    }
}

glib_wrapper! {
    pub struct Page(ObjectSubclass<Page_>)
    @extends gtk::ListBoxRow, gtk::Widget;
}

impl WidgetImpl for Page_ {}
impl ListBoxRowImpl for Page_ {}

impl Page {
    pub fn new() -> Result<Self, glib::BoolError> {
        glib::Object::new(Self::static_type(), &[]).and_then(|a| {
            a.downcast::<Self>()
                .map_err(|e| glib_bool_error!("Faild to downcast {:?}", e.get_type()))
        })
    }

    pub fn set_id(&self, id: i64) {
        (*Page_::from_instance(self).id.borrow_mut()) = id
    }

    pub fn get_id(&self) -> i64 {
        (*Page_::from_instance(self).id.borrow())
    }

    pub fn set_label(&self, text: String) {
        let priv_ = Page_::from_instance(self);

        get_widget!(priv_.builder, gtk::Label, name);
        name.set_label(&text);

        *priv_.name.borrow_mut() = text
    }

    pub fn get_label(&self) -> String {
        (*Page_::from_instance(self).name.borrow()).to_owned()
    }
}

```

page.ui

```
<interface>
  <requires lib="gtk+" version="3.24"/>
    <object class="GtkBox" id="root">
        <property name="visible">True</property>
        <property name="can_focus">False</property>
        <property name="orientation">horizontal</property>
        <property name="margin-start">8</property>
        <property name="margin-end">8</property>
        <property name="margin-top">8</property>
        <property name="margin-bottom">8</property>
        <property name="spacing">8</property>
        <child>
            <object class="GtkImage">
            <property name="visible">True</property>
            <property name="can_focus">True</property>
            <property name="margin-start">8</property>
            <property name="margin-end">8</property>
            <property name="icon-name">emoji-symbols-symbolic</property>
            </object>
        </child>
        <child>
            <object class="GtkLabel" id="name">
            <property name="visible">True</property>
            <property name="can_focus">True</property>
            <property name="xalign">0.0</property>
            <property name="hexpand">True</property>
            <property name="label">Collection Name</property>
            </object>
        </child>
    </object>
</interface>
```

bilelmoussaoui:
- compute_expand is optional and a parent class might not be implementing it
- compute_expand might return an uninitialized pointer for hexpand/vexpand
  in case the vexpand-set/hexpand-set were set to False

Fixes #95

### DIFF
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -5,13 +5,13 @@ use glib::Cast;
 
 use crate::{
     DirectionType, Orientation, SizeRequestMode, Snapshot, StateFlags, SystemSetting,
-    TextDirection, Tooltip, Widget,
+    TextDirection, Tooltip, Widget, WidgetExt,
 };
 use glib::Object;
 
 pub trait WidgetImpl: WidgetImplExt + ObjectImpl {
-    fn compute_expand(&self, widget: &Self::Type, hexpand_p: &mut bool, vexpand_p: &mut bool) {
-        self.parent_compute_expand(widget, hexpand_p, vexpand_p)
+    fn compute_expand(&self, widget: &Self::Type, hexpand: &mut bool, vexpand: &mut bool) {
+        self.parent_compute_expand(widget, hexpand, vexpand)
     }
 
     fn contains(&self, widget: &Self::Type, x: f64, y: f64) -> bool {
@@ -124,12 +124,7 @@ pub trait WidgetImpl: WidgetImplExt + ObjectImpl {
 }
 
 pub trait WidgetImplExt: ObjectSubclass {
-    fn parent_compute_expand(
-        &self,
-        widget: &Self::Type,
-        hexpand_p: &mut bool,
-        vexpand_p: &mut bool,
-    );
+    fn parent_compute_expand(&self, widget: &Self::Type, hexpand: &mut bool, vexpand: &mut bool);
     fn parent_contains(&self, widget: &Self::Type, x: f64, y: f64) -> bool;
     fn parent_direction_changed(&self, widget: &Self::Type, previous_direction: TextDirection);
     fn parent_focus(&self, widget: &Self::Type, direction_type: DirectionType) -> bool;
@@ -172,27 +167,22 @@ pub trait WidgetImplExt: ObjectSubclass {
 }
 
 impl<T: WidgetImpl> WidgetImplExt for T {
-    fn parent_compute_expand(
-        &self,
-        widget: &Self::Type,
-        hexpand_p: &mut bool,
-        vexpand_p: &mut bool,
-    ) {
+    fn parent_compute_expand(&self, widget: &Self::Type, hexpand: &mut bool, vexpand: &mut bool) {
         unsafe {
             let data = T::type_data();
             let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkWidgetClass;
-            let f = (*parent_class)
-                .compute_expand
-                .expect("No parent class impl for \"compute_expand\"");
-            let mut h: i32 = hexpand_p.to_glib();
-            let mut v: i32 = vexpand_p.to_glib();
-            f(
-                widget.unsafe_cast_ref::<Widget>().to_glib_none().0,
-                &mut h,
-                &mut v,
-            );
-            *hexpand_p = from_glib(h);
-            *vexpand_p = from_glib(v);
+            let widget = widget.unsafe_cast_ref::<Widget>();
+            if let Some(f) = (*parent_class).compute_expand {
+                let mut hexpand_glib = hexpand.to_glib();
+                let mut vexpand_glib = vexpand.to_glib();
+                f(
+                    widget.to_glib_none().0,
+                    &mut hexpand_glib,
+                    &mut vexpand_glib,
+                );
+                *hexpand = from_glib(hexpand_glib);
+                *vexpand = from_glib(vexpand_glib);
+            }
         }
     }
 
@@ -546,10 +536,21 @@ unsafe extern "C" fn widget_compute_expand<T: WidgetImpl>(
     let instance = &*(ptr as *mut T::Instance);
     let imp = instance.get_impl();
     let wrap: Borrowed<Widget> = from_glib_borrow(ptr);
-    let mut hexpand: bool = from_glib(*hexpand_ptr);
-    let mut vexpand: bool = from_glib(*vexpand_ptr);
+
+    let widget = wrap.unsafe_cast_ref::<Widget>();
+    let mut hexpand: bool = if widget.get_hexpand_set() {
+        widget.get_hexpand()
+    } else {
+        from_glib(*hexpand_ptr)
+    };
+    let mut vexpand: bool = if widget.get_vexpand_set() {
+        widget.get_vexpand()
+    } else {
+        from_glib(*vexpand_ptr)
+    };
 
     imp.compute_expand(wrap.unsafe_cast_ref(), &mut hexpand, &mut vexpand);
+
     *hexpand_ptr = hexpand.to_glib();
     *vexpand_ptr = vexpand.to_glib();
 }


bilelmoussaoui:
@kavanmevada mind trying if the changes on this PR fixes your issue? :)

kavanmevada:
Yup, It fixed, now hexpand working fine.

gtk4/src/subclass/widget.rs
```diff
@@ -181,18 +181,25 @@ impl<T: WidgetImpl> WidgetImplExt for T {
...
-            );
-            *hexpand_p = from_glib(h);
-            *vexpand_p = from_glib(v);
+            let widget = widget.unsafe_cast_ref::<Widget>();
+            if let Some(f) = (*parent_class).compute_expand {
+                let mut h: i32 = hexpand_p.to_glib();
+                let mut v: i32 = vexpand_p.to_glib();
+                f(widget.to_glib_none().0, &mut h, &mut v);
+                if !widget.get_hexpand_set() {
+                    *hexpand_p = from_glib(h);
```
sdroege, line 190: You need to do something similar also in the function called from C, otherwise you'll pass around mutable references to uninitialized data if `Xexpand_set == false`.

gtk4/src/subclass/widget.rs
```diff
@@ -181,18 +181,25 @@ impl<T: WidgetImpl> WidgetImplExt for T {
...
-                &mut v,
-            );
-            *hexpand_p = from_glib(h);
-            *vexpand_p = from_glib(v);
+            let widget = widget.unsafe_cast_ref::<Widget>();
+            if let Some(f) = (*parent_class).compute_expand {
+                let mut h: i32 = hexpand_p.to_glib();
+                let mut v: i32 = vexpand_p.to_glib();
+                f(widget.to_glib_none().0, &mut h, &mut v);
+                if !widget.get_hexpand_set() {
```
sdroege, line 189: Ah I missed that these are readable from the outside. That makes it less annoying to handle this correctly as you can know when it's uninitialized and in that case you can just initialize it to `false` or so :)

gtk4/src/subclass/widget.rs
```diff
@@ -550,8 +557,15 @@ unsafe extern "C" fn widget_compute_expand<T: WidgetImpl>(
     let mut vexpand: bool = from_glib(*vexpand_ptr);
 
     imp.compute_expand(wrap.unsafe_cast_ref(), &mut hexpand, &mut vexpand);
-    *hexpand_ptr = hexpand.to_glib();
-    *vexpand_ptr = vexpand.to_glib();
+
+    let widget = wrap.unsafe_cast_ref::<Widget>();
+
+    if !widget.get_hexpand_set() {
+        *hexpand_ptr = hexpand.to_glib();
```
sdroege, line 564: This is not a problem, the C code just ignores it. You can remove this part, but ...

gtk4/src/subclass/widget.rs
```diff
@@ -550,8 +557,15 @@ unsafe extern "C" fn widget_compute_expand<T: WidgetImpl>(
     let mut vexpand: bool = from_glib(*vexpand_ptr);
```
sdroege, line 557: ... these are a problem because you read uninitialized memory here depending on whether `Xexpand_set` is true or not.

gtk4/src/subclass/widget.rs
```diff
@@ -181,18 +181,25 @@ impl<T: WidgetImpl> WidgetImplExt for T {
...
+            if let Some(f) = (*parent_class).compute_expand {
+                let mut h: i32 = hexpand_p.to_glib();
+                let mut v: i32 = vexpand_p.to_glib();
+                f(widget.to_glib_none().0, &mut h, &mut v);
+                if !widget.get_hexpand_set() {
+                    *hexpand_p = from_glib(h);
+                }
+                if !widget.get_vexpand_set() {
+                    *vexpand_p = from_glib(v);
+                }
```
sdroege, line 194: This shouldn't be needed, you can just forward the values as is (pass `v_expand`/`h_expand` directly to `f`).

gtk4/src/subclass/widget.rs
```diff
@@ -181,18 +181,25 @@ impl<T: WidgetImpl> WidgetImplExt for T {
...
-            f(
-                widget.unsafe_cast_ref::<Widget>().to_glib_none().0,
-                &mut h,
-                &mut v,
-            );
-            *hexpand_p = from_glib(h);
-            *vexpand_p = from_glib(v);
+            let widget = widget.unsafe_cast_ref::<Widget>();
+            if let Some(f) = (*parent_class).compute_expand {
+                let mut h: i32 = hexpand_p.to_glib();
```
sdroege, line 186:
```suggestion
                let mut h: i32 = hexpand.to_glib();
```

Maybe?

gtk4/src/subclass/widget.rs
```diff
@@ -181,18 +181,25 @@ impl<T: WidgetImpl> WidgetImplExt for T {
...
+                if !widget.get_vexpand_set() {
+                    *vexpand_p = from_glib(v);
+                }
+            } else {
+                if !widget.get_hexpand_set() {
+                    *hexpand_p = widget.get_hexpand();
+                }
+                if !widget.get_vexpand_set() {
+                    *vexpand_p = widget.get_vexpand();
+                }
```
sdroege, line 201: This is all not needed if you already initialize the values properly in `widget_compute_expand`

gtk4/src/subclass/widget.rs
bilelmoussaoui: Ah my bad. I should have took a coffee before looking at this.

gtk4/src/subclass/widget.rs
bilelmoussaoui: Ah that would simplify things a bit. Although, f takes a gboolean i'm not sure if what I pushed is correct or not

gtk4/src/subclass/widget.rs
```diff
@@ -172,27 +167,18 @@ pub trait WidgetImplExt: ObjectSubclass {
...
-                &mut v,
-            );
-            *hexpand_p = from_glib(h);
-            *vexpand_p = from_glib(v);
+            let widget = widget.unsafe_cast_ref::<Widget>();
+            if let Some(f) = (*parent_class).compute_expand {
+                f(
+                    widget.to_glib_none().0,
+                    &mut hexpand.to_glib(),
+                    &mut vexpand.to_glib(),
```
sdroege, line 179: Ah they're bools, for some reason I thought they're i32s.

Then you need to

```rust
let hexpand_glib = hexpand.to_glib();
[...]
f(..., &mut hexpand_glib, ...);
*hexpand = hexpand_glib;
[...]
```

sdroege: Looks good apart from that now. Can you also send the same PR to gtk3?

sdroege: Otherwise the problem is that you create a temporary local variable, pass that to C, C modifies that, and then it's lost.

sdroege (approved):

PR #97, opened 2020-12-01
WIP: Add #[template(file="FILENAME")] tag to gtk4-macros

felinira:
This automatically generates composite template classes from template files.
Add examples/src/bin/auto_composite_template to demonstrate the behavior.

There is still some cleanup to do, especially in `composite_template_attr.rs`. I'm also not really happy about the fact that you still need to write quite a bit of boilerplate code.


examples/src/bin/auto_composite_template.rs
```diff
@@ -0,0 +1,64 @@
...
+use gtk::subclass::prelude::*;
+use gtk::subclass::widget::*;
+
+mod imp {
+    use super::*;
+
+    /// The private struct, which will hold the automatically generated widgets
+    #[template(file = "examples/src/bin/composite_template.ui")]
+    #[derive(Debug)]
+    pub struct ExApplicationWindow {}
```
sdroege, line 20: I didn't look at the code macro code yet, but is it possible to have other fields in the struct and those would be left alone by the macro?

gtk4-macros/src/lib.rs
```diff
@@ -12,3 +14,12 @@ pub fn composite_template_derive(input: TokenStream) -> TokenStream {
     let gen = composite_template_derive::impl_composite_template(&input);
     gen.into()
 }
+
+#[proc_macro_attribute]
+#[proc_macro_error]
+pub fn template(attrs: TokenStream, item: TokenStream) -> TokenStream {
```
sdroege, line 20: Can you add some documentation for the macro?

examples/src/bin/auto_composite_template.rs
felinira: In theory yes. In practice this does not work because the code currently generates a ::new() function which needs to specify all struct members. Still need to figure out a better solution to this.

gtk4-macros/src/composite_template_attr.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+/// Input is:
+/// struct MyGtkSubclass {
+/// }
+///
+/// We want to generate the struct members and the impl for
+/// * CompositeTemplate
+/// * ObjectSubclass
+/// * ObjectImpl
```
sdroege, line 48: As it generates `ObjectSubclass` and `ObjectImpl`  impls, is there a way to provide them manually or to somehow override (parts of) them? Do you have an idea how that could look like API-wise?

For example `ObjectImpl::get_property` / `set_property` would seem like something that needs to be overridden, and for those I have a couple of ideas how that can also be autogenerated to some degree (that's going to be part of https://github.com/gtk-rs/gtk-rs/pull/31). I'll add those ideas to a comment here tomorrow.

gtk4-macros/src/lib.rs
felinira: I'll definitely do that

gtk4-macros/src/composite_template_attr.rs
sdroege: Similarly `ObjectSubclass::class_init`, `new` and `ObjectImpl::constructed` would be things that will need a way to be overridden.

In short my idea was to not actually derive those impls (yet! that's the next step) but to have an auxiliary struct plus trait that handles this (in which case a derive macro would also be sufficient). Then you would call a single function on that as part of e.g. `ObjectSubclass::class_init` (and wherever else needed) to get it setup up properly.

sdroege:
This looks quite nice overall in any case, thanks :) I think we'll just need to find a way to make this possible in a composable way.

gtk4-macros/src/composite_template_attr.rs
felinira: I see two possibilities: Add `#[attrs]` to modify what is automatically generated and what not or a separate `#[derive]` macro that generates those impls. "Unfortunately" derives are evaluated before attribute macros so I'm not sure if this is possible to do even though it would look less magical and would make more sense.

felinira:
Is there any way the code could automatically generate subclasses for the template class? In theory the <template class="" parent=""> xml token is supposed to do that but I can't figure out a way to implement it because you always need to specify all superclasses to `glib_wrapper!()`

gtk4-macros/src/composite_template_attr.rs
sdroege: The problem with `#[derive]` for exactly those traits however is that that derive macro would also need to compose with anything else in the class hierarchy (upwards and downwards) that might want to extend the impl

sdroege:
> I can't figure out a way to implement it because you always need to specify all superclasses to `glib_wrapper!()`

That's a big open question. With reading the `.gir` file, yes, but that is not really an option as it complicates everything for users (and increases compile times considerably). Related to this topic are https://github.com/gtk-rs/gtk-rs/issues/44 and https://github.com/gtk-rs/gtk-rs/issues/142

gtk4-macros/src/composite_template_attr.rs
felinira: I don't have a problem to remove those impls for now and look for a better solution later. Before the API gets finalized I think this deserves a second look though. Maybe it might be necessary to rename the template method to keep the simple `template` name for something more higher level than this?

felinira:
The alternative would be to at least provide a static field for the template superclass. This would maybe make the API not ideal but still manageable. I just don't know how one would name it to prevent namespace collisions.

sdroege:
> The alternative would be to at least provide a static field for the template superclass. This would maybe make the API not ideal but still manageable. I just don't know how one would name it to prevent namespace collisions.

What do you mean exactly with this?

gtk4-macros/src/composite_template_attr.rs
sdroege: Yes if you don't have an idea right now let's keep it simple for now and a bit more verbose for users. But I agree that we should try finding a way to make this as convenient as the current version of the macro, while somehow still keeping it composable.

Maybe you could make this either
 - An attribute macro that adds the fields to the struct and derives the composite template trait
 - An attribute macro for another struct that adds the fields and derives some trait and then you'd have

```rust
#[template("path/to/ui")]
struct TemplateWidgets;

#[derive(CompositeTemplate)]
struct MyWidget {
    #[templates]
    widgets: TemplateWidgets,
    other_fields,
}
```

This would still require the `ObjectImpl` and `ObjectSubclass` impls as before but would at least autogenerate the struct from the UI file and autogenerate a function to bind them all.

examples/src/bin/auto_composite_template.rs
```diff
@@ -0,0 +1,64 @@
...
+use gtk::prelude::*;
+
+use gtk::subclass::prelude::*;
+use gtk::subclass::widget::*;
+
+mod imp {
+    use super::*;
+
+    /// The private struct, which will hold the automatically generated widgets
+    #[template(file = "examples/src/bin/composite_template.ui")]
```
sdroege, line 18: Would this cause rebuilding if the .ui file changes?

felinira:
> > The alternative would be to at least provide a static field for the template superclass. This would maybe make the API not ideal but still manageable. I just don't know how one would name it to prevent namespace collisions.
> 
> What do you mean exactly with this?

The object that is annotated with the <template class="" parent=""> does not have an id field. We could still implicitly add a struct field for it though and name it something like root_widget or whatever.

I also didn't yet consider templates that don't use <template> but only object tags. Don't know if I should though. If this doesn't provide any impls then it doesn't really require it I guess?

bilelmoussaoui:
> I also didn't yet consider templates that don't use `<template>` but only object tags. Don't know if I should though. If this doesn't provide any impls then it doesn't really require it I guess?

In such case you're not creating a composite template (a subclass of the parent), so none of this would work.

gtk4-macros/src/composite_template_attr.rs
felinira: What would `#[templates]` do in that case? And where would I put the trait? Should probably not be in gtk4-macros but somewhere else right? I also don't know how the widget class would even know its parent class then: As the template is annotated one struct above there is no way to communicate here.

In this case we also can't ever statically check at any point in the future if the struct is actually derived from the parent class because only the widget struct would have access to the widget file.

We could also annotate the module. That might make it more flexible for any future stuff that would be done with the template file as it woukd be in scope and one could easily add an attribute to anything inside later to change the behavior of the macro etc.

```
#[template(file="path/to/ui")]
mod my_widget_imp {
    #[derive(CompositeTemplate)]
    struct MyWidget {
        #[templates]
        widgets: TemplateWidgets,
        other_fields,
    }
}
```

sdroege: I guess that could work too, yes. (annotating the module) Just means that you have to parse much more code and insert your things into possibly existing user code into the right places.

> What would #[templates] do in that case?

Just tell the derive macro that this is the thing that stores all the widgets and implements your new trait, so it can automatically map to register them all via the `CompositeTemplate` trait.

> And where would I put the trait? Should probably not be in gtk4-macros but somewhere else right?

You would have to put it into gtk4. Everything but the actual macro needs to be in normal crates.

>  I also don't know how the widget class would even know its parent class then: As the template is annotated one struct above there is no way to communicate here.

The user would (for now) still manually call into the `CompositeTemplate` trait functions from `class_init` and `constructed`, and that would dispatch to the autogenerated stuff.

> In this case we also can't ever statically check at any point in the future if the struct is actually derived from the parent class because only the widget struct would have access to the widget file.

Can you give an example of what you mean? Like the UI file says that this is a `GtkDialog` but actually the subclass in question is a `GtkApplicationWindow`?

That would indeed not be directly possible anymore, at least statically. At runtime it's as easy to check as otherwise.

felinira:
I reimplemented the macro to annotate the module instead and use the trick with the child struct for the widgets. Now it leaves everything alone and just generates some code around the existing code in the module. It seems clearer to me that way.

Because I couldn't accept not doing anything with the ParentType (It's just too convenient that the template provides it!) I created a CompositeTemplateContext trait that holds the NAME of the current class and the ParentType. That should make it easier to derive macro the ObjectSubclass later.

examples/src/bin/auto_composite_template.rs
felinira: I tested it a few times and it does seem to rerun the macro every time the .ui file changes. I think it's actually because it generates include_bytes! somewhere in the output and include_bytes! seems to trigger a rebuild if the file changes.

gtk4-macros/src/composite_template_attr.rs
BrainBlasted: > An attribute macro for another struct that adds the fields and derives some trait and then you'd have

Changes like these would add an extra layer on the template usage end when accessing the template children :/ Also mind, not all children of a template are widgets - you also have things like gestures and menus.

examples/src/bin/auto_composite_template.rs
BrainBlasted: Hmm - include_bytes would mean this doesn't work with gresources - perhaps there needs to be `file` and `resource` arguments separately

gtk4-macros/src/composite_template_attr.rs
BrainBlasted: E.g. https://github.com/gtk-rs/gtk4-rs/pull/97/files#diff-a24e3fff03262f200f6a7bea3104da8ab03e6aa8234d005bf58b86e14e8829ebR83 feels a bit too long. You'd need to do this everywhere when accessing a template child.

BrainBlasted: One Q: does this work with types defined within the application code?

When working with templates you'll often have something like this:

```xml
<template class="TemplateFoo" parent=GtkBox">
  <child>
    <object class="TemplateBar">
        <!-- properties, children, and gestures on TemplateBar here -->
    </object>
  </child>
</template>
```

where TemplateBar is another object/type defined in the application code.

gtk4-macros/src/util.rs
```diff
@@ -40,43 +43,76 @@ fn parse_attribute(meta: &NestedMeta) -> Result<(String, String)> {
...
+
+pub fn generate_type_from_class_name(class_name: &str) -> Result<Type, syn::Error> {
+    let mut ident_name = class_name.to_string();
 
-    Ident::new(&crate_name, Span::call_site())
+    if let Some(prefix) = class_name.get(0..3) {
+        if let Some(suffix) = class_name.get(3..) {
+            let mut lib = None;
+
+            if prefix.eq("Gtk") {
```
BrainBlasted, line 99: Looks like this answers my question above - hardcoding the libraries here doesn't seem right, and won't work for custom widgets or different libraries.

BrainBlasted (changes requested):

gtk4-macros/src/util.rs
felinira: True. Not happy about it either. Is there a better way?

sdroege: How are Python/Vala/etc handling this? I guess the dynamic languages can just look into the typelibs at runtime to get the C type name mapping but Vala?

gtk4-macros/src/composite_template_attr.rs
felinira: It should work as long as the type is in the namespace.
Concerning the separate widgets class: I could implement Deref. But that really just seems like a hack at that point.

gtk4-macros/src/util.rs
felinira: Looking at vala it seems to build up a type registry at compile time. As in it registers all types in a dict to look up later. But not sure exactly, only glanced at the code.

sdroege: That would require us to read the .gir files at compile-time (slow, fragile) and also create a registry of all types inside the application somehow (I don't think that's possible unless we pre-process all code from build.rs).

felinira: This is really difficult. The only way I could think of would be if all rust crates that provide widgets would alias their types somewhere so that no replacement needs to be done at all. gtk would for example alias `gtk::Box` as `GtkBox`. It could be done in the prelude or it could be in a separate module namespace like
```rust
use gtk::nativetypes;

// autogenerated
struct Widgets {
    label: nativetypes::GtkLabel;
}
```

If you had multiple libraries you would:
```rust
mod nativetypes {
    use gtk::nativetypes::*;
    use libhandy::nativetypes::*;
}

// autogenerated
struct Widgets {
    label: nativetypes::GtkLabel;
    action_row: nativetypes::HdyActionRow;
}

PR #103, opened 2020-12-07
Macro parser improvements

felinira:
Rewrite macro attribute parser to improve error messages

This is a rewrite of the attribute parser in gtk4-macros.
It will now print clear error messages that are annotated at the correct span.

It will also allow to leave out the id field in `#[template_child]` if the id is the same as the identifier.

I really looked around for crates that do this job, but they all don't really seem to support arbitrary attributes inside structs.

This will now catch errors such as:

- multiple attributes for the same field (only recognizes the template attributes, other attributes are fine)
- multiple instances of the same attribute argument (such as `#[template_child(id = "widget", id = "otherwidget")]`)
- wrong attribute arguments
- wrong type of attribute arguments (`ident` or `"literal"` instead of `name = value`)
- wrong type of argument values (number literal instead of string etc.)
- missing attribute arguments should also be easy to add if some new attribute might need that.
- any syn parse errors are correctly handled as well

All with correct span annotation so the compiler will throw nice error messages.


sdroege:
CC @BrainBlasted @gdesmott

sdroege (approved):
Looks good to me. @bilelmoussaoui @BrainBlasted ? :)

Should also be backported to gtk3

bilelmoussaoui:
Thanks for working on this! and yes we should definitely try to keep both templates support in sync between both gtk3/gtk4

BrainBlasted (approved):

sdroege:
See https://github.com/gtk-rs/gtk-rs/pull/206

PR #105, opened 2020-12-07
Rexport dependencies from gtk

### LINKED ISSUE #73: Re-export gdk/gsk/glib/gio
see https://github.com/gtk-rs/gtk-rs/issues/143

bilelmoussaoui:
Fixes #73

### DIFF
--- a/examples/src/bin/basic.rs
+++ b/examples/src/bin/basic.rs
@@ -4,8 +4,7 @@
 //! position, how to add a `button` to this `window` and how to connect signals with
 //! actions.
 
-use gio::prelude::*;
-use gtk::prelude::*;
+use gtk::{gio::prelude::*, prelude::*};
 
 use std::env::args;
 
--- a/examples/src/bin/basic.rs
+++ b/examples/src/bin/basic.rs
@@ -3,8 +3,7 @@
 //! This sample demonstrates how to create a toplevel `window`, set its title, size and
 //! position, how to add a `button` to this `window` and how to connect signals with
 //! actions.
-
-use gtk::{gio::prelude::*, prelude::*};
+use gtk::prelude::*;
 
 use std::env::args;
 
--- a/gtk4/src/lib.rs
+++ b/gtk4/src/lib.rs
@@ -12,6 +12,16 @@
 #![allow(deprecated)]
 
 pub use ffi;
+// Re-export gtk dependencies
+pub use cairo;
+pub use gdk;
+pub use gdk_pixbuf;
+pub use gio;
+pub use glib;
+pub use graphene;
+pub use gsk;
+pub use pango;
+
 #[macro_use]
 #[doc(hidden)]
 #[allow(unused_imports)]
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -5,23 +5,12 @@ authors = ["The Gtk-rs Project Developers"]
 edition = "2018"
 autobins = false
 
-
 [dependencies]
 chrono = "0.4.19"
 
 [dependencies.gtk]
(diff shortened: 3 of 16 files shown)


examples/src/bin/basic.rs
```diff
@@ -4,7 +4,7 @@
 //! position, how to add a `button` to this `window` and how to connect signals with
 //! actions.
 
-use gio::prelude::*;
+use gtk::gio::prelude::*;
```
sdroege, line 7: `gtk::prelude` should simply re-export `gio::prelude::*` (and `glib::prelude::*` etc)

gtk4/src/lib.rs
```diff
@@ -14,6 +14,16 @@
 #![allow(deprecated)]
 
 pub use ffi;
+// Re-export gtk dependencies
```
sdroege, line 17: Only crates that defines types that are used in the public API IMHO, not all dependencies. But I think that's the case here?

sdroege:
Can you do the same for gtk-rs too?

gtk4/src/lib.rs
bilelmoussaoui: Well, yes, those are the gtk dependencies that are part of the public API.

bilelmoussaoui: graphene/cairo are used for snapshot stuff along with gsk. gdk_pixbuf is still used on cell renders with a promise to be gone in gtk5 :)

sdroege: Ok :)

examples/src/bin/basic.rs
bilelmoussaoui: That doesn't cause conflicts in case you have gtk::SettingsExt & gio::SettingsExt? :)

sdroege: It does, that's why it should be `gtk::GtkSettingsExt`. The rename feature exists for that very reason, and that's why it's e.g. `gst::GstObjectExt` :)

bilelmoussaoui: Makes sense, I have updated the PR accordingly

sdroege: No I meant that `gtk::prelude` has a
```rust
pub use gio::prelude::*;
```

In it. So that you only need to `use gtk::prelude::*` and then get everything you need :)

bilelmoussaoui: I have done exactly no?  ^^

bilelmoussaoui:
> Can you do the same for gtk-rs too?

Sure

examples/src/bin/basic.rs
sdroege: But then you wouldn't need `use gtk::{gio::prelude::*, prelude::*};` in the examples but just `use gtk::prelude::*;`

bilelmoussaoui: Ah yes, that's an old commit :P

sdroege:
Shouldn't the same be done for gdk4/gsk4?

bilelmoussaoui:
> Shouldn't the same be done for gdk4/gsk4?

Which makes me wonder, should gdk re-export gdk-x11/gdk-wayland using a feature flag?

sdroege:
> Which makes me wonder, should gdk re-export gdk-x11/gdk-wayland using a feature flag?

That seems potentially useful, but it would make `gdk` depend on those crates... while those crates themselves already depend on `gdk` :)

bilelmoussaoui:
This one should be ready, it makes getting started with gtk-rs much simpler as you don't have to know all the details to display a window. I like it :)

sdroege (approved):

PR #108, opened 2020-12-08
Regen with gtk 3.99.5

### LINKED ISSUE #40: TreeStore: manually implement remove
the param uses `inout` direction which can't be automatically generated by gir for now, this needs to be manually implemented for now.

```
  <method name="remove" c:identifier="gtk_tree_store_remove">
        <doc xml:space="preserve">Removes @iter from @tree_store.  After being removed, @iter is set to the
next valid row at that level, or invalidated if it previously pointed to the
last one.</doc>
        
        <return-value transfer-ownership="none">
          <doc xml:space="preserve">%TRUE if @iter is still valid, %FALSE if not.</doc>
          <type name="gboolean" c:type="gboolean"/>
        </return-value>
        <parameters>
          <instance-parameter name="tree_store" transfer-ownership="none">
            <doc xml:space="preserve">A #GtkTreeStore</doc>
            <type name="TreeStore" c:type="GtkTreeStore*"/>
          </instance-parameter>
          <parameter name="iter" direction="inout" caller-allocates="0" transfer-ownership="full">
            <doc xml:space="preserve">A valid #GtkTreeIter</doc>
            <type name="TreeIter" c:type="GtkTreeIter*"/>
          </parameter>
        </parameters>
      </method>
```

bilelmoussaoui:
Replaces #75 

Closes #40


gtk4/Gir.toml
```diff
@@ -1021,12 +959,6 @@ trust_return_value_nullability = false
     name = "add_choice"
     ignore = true  #strange type of options and option_labels
 
-[[object]]
-name = "Gtk.FileChooserButton"
```
sdroege, line 1025: The file for this also needs to be removed or not? Also check if there are other files :)

sdroege:
Looks ok otherwise, please update gir-files :)

gtk4/Gir.toml
alatiera: Yes, `FileChooserButton` no longer exists iirc

sdroege: Yes I mean the generated rs source file has to be removed from git, if that didn't happen already

bilelmoussaoui: ah yes, i thought i did.

sdroege (approved):

bilelmoussaoui:
CI is happy :yay:

PR #112, opened 2020-12-10
Add a Widget subclass example

YaLTeR:


examples/src/bin/widget_subclass.rs
```diff
@@ -0,0 +1,104 @@
...
+        let gesture = gtk::GestureClick::new();
+        gesture.connect_released(|gesture, _, _, _| {
+            gesture.set_state(gtk::EventSequenceState::Claimed);
+            println!("Button pressed!");
+        });
+        obj.add_controller(&gesture);
+    }
+}
+
+impl WidgetImpl for ButtonPrivate {}
```
bilelmoussaoui, line 69: We tend to put the private struct in a imp module

sdroege: Basically a logical extension of https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#avoid-redundant-prefixes-[rfc-356] . Use module namespaces instead of putting namespaces into type names.

YaLTeR: Updated.

bilelmoussaoui:
Thanks!

examples/src/bin/widget_subclass.rs
```diff
@@ -0,0 +1,109 @@
...
+    use gtk::subclass::prelude::*;
+
+    #[derive(Debug)]
+    pub struct Button {
+        child: RefCell<WeakRef<gtk::Widget>>,
+    }
+
+    impl Drop for Button {
+        fn drop(&mut self) {
+            // Child widgets need to be manually unparented.
```
sdroege, line 20: I think this has to happen from `dispose()` and not `finalize()` actually, which then needs someone to think about how to do that in a safe way.

sdroege: Reason being `gtk_container_destroy()` etc calling `g_object_run_dispose()` and that supposedly cleaning up the whole widget hierarchy. IMHO a hack for not using weak references correctly in GTK but that's just how it is.

YaLTeR: Yeah, I was asking about this yesterday but couldn't find a way to hook into dispose.

sdroege: Needs changes in `glib/src/subclass` and figuring out a way how to safely expose that. At the moment `dispose()` is called the object is not fully existing anymore. It's basically a zombie. Calling any API on it can cause all kinds of interesting effects.

The way how GTK does the `g_object_run_dispose()` is unsafe by design and will probably cause some more headaches in Rust. Something to figure out.

Issue #111, opened 2020-12-10
Add missing WidgetClass methods


bilelmoussaoui:
- [x] `gtk_widget_class_add_binding`
- [x] `gtk_widget_class_add_binding_action`
- [x] `gtk_widget_class_add_binding_signal`
- [x] `gtk_widget_class_add_shortcut`
- [x] ~~`gtk_widget_class_bind_template_callback_full`~~ C only, for language bindings, we should use #24 
- [x] `gtk_widget_class_get_accessible_role`
- [x] `gtk_widget_class_get_activate_signal`
- [x] `gtk_widget_class_get_css_name`
- [x] `gtk_widget_class_get_layout_manager_type`
- [x] `gtk_widget_class_install_action`
- [x] `gtk_widget_class_install_property_action`
- [x] `gtk_widget_class_query_action`
- [x] `gtk_widget_class_set_accessible_role`
- [x] `gtk_widget_class_set_activate_signal`
- [x] `gtk_widget_class_set_activate_signal_from_name`
- [x] `gtk_widget_class_set_template_scope`

The only other widget with class methods we can implement is CellArea

- [ ] `gtk_cell_area_class_find_cell_property`
- [ ] `gtk_cell_area_class_install_cell_property`
- [ ] `gtk_cell_area_class_list_cell_properties`

This one is called on ObjectClass, not sure why it's there...
- [x] `gtk_editable_install_properties`

PR #113, opened 2020-12-10
gtk: add more WidgetClass methods

bilelmoussaoui:
Part of #111

### DIFF
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -849,10 +849,7 @@ unsafe extern "C" fn widget_unroot<T: WidgetImpl>(ptr: *mut ffi::GtkWidget) {
 pub unsafe trait WidgetClassSubclassExt: ClassStruct {
     fn set_template_bytes(&mut self, template: &glib::Bytes) {
         unsafe {
-            let type_class = self as *mut _ as *mut glib::gobject_ffi::GTypeClass;
-            let widget_class =
-                glib::gobject_ffi::g_type_check_class_cast(type_class, ffi::gtk_widget_get_type())
-                    as *mut ffi::GtkWidgetClass;
+            let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
             ffi::gtk_widget_class_set_template(widget_class, template.to_glib_none().0);
         }
     }
@@ -869,10 +866,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
 
     fn set_template_from_resource(&mut self, resource_name: &str) {
         unsafe {
-            let type_class = self as *mut _ as *mut glib::gobject_ffi::GTypeClass;
-            let widget_class =
-                glib::gobject_ffi::g_type_check_class_cast(type_class, ffi::gtk_widget_get_type())
-                    as *mut ffi::GtkWidgetClass;
+            let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
             ffi::gtk_widget_class_set_template_from_resource(
                 widget_class,
                 resource_name.to_glib_none().0,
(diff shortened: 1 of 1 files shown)


gtk4/src/subclass/widget.rs
```diff
@@ -895,6 +895,26 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+        unsafe {
+            let type_class = self as *mut _ as *mut glib::gobject_ffi::GTypeClass;
+            let widget_class =
+                glib::gobject_ffi::g_type_check_class_cast(type_class, ffi::gtk_widget_get_type())
+                    as *mut ffi::GtkWidgetClass;
+            ffi::gtk_widget_class_add_shortcut(widget_class, shortcut.to_glib_none().0);
+        }
+    }
+
+    fn get_layout_manager_type(&mut self) -> glib::Type {
```
sdroege, line 908:
```suggestion
    fn get_layout_manager_type(&self) -> glib::Type {
```

gtk4/src/subclass/widget.rs
```diff
@@ -915,6 +935,39 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    fn get_css_name(&mut self) -> glib::GString {
```
sdroege, line 938:
```suggestion
    fn get_css_name(&self) -> glib::GString {
```

gtk4/src/subclass/widget.rs
```diff
@@ -915,6 +935,39 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+            let type_class = self as *mut _ as *mut glib::gobject_ffi::GTypeClass;
+            let widget_class =
+                glib::gobject_ffi::g_type_check_class_cast(type_class, ffi::gtk_widget_get_type())
+                    as *mut ffi::GtkWidgetClass;
+
+            ffi::gtk_widget_class_set_accessible_role(widget_class, role.to_glib());
+        }
+    }
+
+    fn get_accessible_role(&mut self) -> AccessibleRole {
```
sdroege, line 960:
```suggestion
    fn get_accessible_role(&self) -> AccessibleRole {
```

sdroege:
Needs to be backported to gtk3-rs too

gtk4/src/subclass/widget.rs
bilelmoussaoui: Oups, copy pasting error

bilelmoussaoui: Actually, those methods requires a mutable instance of the class 
```
    pub fn gtk_widget_class_get_accessible_role(
        widget_class: *mut GtkWidgetClass,
    ) -> GtkAccessibleRole;
```

sdroege: No, you just cast it from `*const _` to `*mut _`. That's just C taking const not very serious.

bilelmoussaoui: Aha, I somehow forgot this. Should be properly handled now :)

gtk4/src/subclass/widget.rs
```diff
@@ -895,6 +895,16 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    fn add_shortcut(&mut self, shortcut: &Shortcut) {
+        unsafe {
+            let type_class = self as *mut _ as *mut glib::gobject_ffi::GTypeClass;
+            let widget_class =
+                glib::gobject_ffi::g_type_check_class_cast(type_class, ffi::gtk_widget_get_type())
+                    as *mut ffi::GtkWidgetClass;
```
sdroege, line 903: Now my remaining question: why so complicated? :) We already statically ensured that the type is correct. You can simply cast `self as [...] as *mut ffi::GtkWidgetClass` instead of doing the runtime type check.

bilelmoussaoui: I guess that went unnoticed from the composite template PR :)

sdroege (approved):
Thanks! Should also do all this for gtk3 :)

bilelmoussaoui:
> Thanks! Should also do all this for gtk3 :)

I will try to backport as much as I can :)

PR #114, opened 2020-12-11
gtk: add IMContext subclassing support

bilelmoussaoui:
Part of #56

### DIFF
--- a/gtk4/src/subclass/im_context.rs
+++ b/gtk4/src/subclass/im_context.rs
@@ -1,0 +1,522 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::{IMContext, Widget};
+use glib::subclass::object::ObjectImpl;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, GString, IsA, Object};
+use pango::AttrList;
+
+pub trait IMContextImpl: IMContextImplExt + ObjectImpl {
+    fn commit(&self, im_context: &Self::Type, string: &str) {
+        self.parent_commit(im_context, string)
+    }
+    fn delete_surrounding(&self, im_context: &Self::Type, offset: i32, n_chars: i32) -> bool {
+        self.parent_delete_surrounding(im_context, offset, n_chars)
+    }
+    fn filter_keypress(&self, im_context: &Self::Type, event: &gdk::Event) -> bool {
+        self.parent_filter_keypress(im_context, event)
+    }
+    fn focus_in(&self, im_context: &Self::Type) {
+        self.parent_focus_in(im_context)
+    }
+    fn focus_out(&self, im_context: &Self::Type) {
+        self.parent_focus_out(im_context)
+    }
+    fn get_preedit_string(&self, im_context: &Self::Type) -> (GString, AttrList, i32) {
+        self.parent_get_preedit_string(im_context)
+    }
+    fn get_surrounding(&self, im_context: &Self::Type) -> Option<(GString, i32)> {
+        self.parent_get_surrounding(im_context)
+    }
+    fn preedit_changed(&self, im_context: &Self::Type) {
+        self.parent_preedit_changed(im_context)
+    }
+    fn preedit_end(&self, im_context: &Self::Type) {
+        self.parent_preedit_end(im_context)
+    }
+    fn preedit_start(&self, im_context: &Self::Type) {
+        self.parent_preedit_start(im_context)
+    }
+    fn reset(&self, im_context: &Self::Type) {
+        self.parent_reset(im_context)
+    }
+    fn retrieve_surrounding(&self, im_context: &Self::Type) -> bool {
+        self.parent_retrieve_surrounding(im_context)
+    }
+    fn set_client_widget<P: IsA<Widget>>(&self, im_context: &Self::Type, widget: Option<&P>) {
+        self.parent_set_client_widget(im_context, widget)
+    }
+    fn set_cursor_location(&self, im_context: &Self::Type, area: &gdk::Rectangle) {
+        self.parent_set_cursor_location(im_context, area)
+    }
+    fn set_surrounding(&self, im_context: &Self::Type, text: &str, cursor_index: i32) {
+        self.parent_set_surrounding(im_context, text, cursor_index)
+    }
+    fn set_use_preedit(&self, im_context: &Self::Type, use_preedit: bool) {
+        self.parent_set_use_preedit(im_context, use_preedit)
+    }
+}
+
+pub trait IMContextImplExt: ObjectSubclass {
+    fn parent_commit(&self, im_context: &Self::Type, string: &str);
+    fn parent_delete_surrounding(&self, im_context: &Self::Type, offset: i32, n_chars: i32)
+        -> bool;
+    fn parent_filter_keypress(&self, im_context: &Self::Type, event: &gdk::Event) -> bool;
+    fn parent_focus_in(&self, im_context: &Self::Type);
+    fn parent_focus_out(&self, im_context: &Self::Type);
+    fn parent_get_preedit_string(&self, im_context: &Self::Type) -> (GString, AttrList, i32);
+    fn parent_get_surrounding(&self, im_context: &Self::Type) -> Option<(GString, i32)>;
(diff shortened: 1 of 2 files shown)


gtk4/src/subclass/im_context.rs
```diff
@@ -0,0 +1,510 @@
...
+    }
+    fn retrieve_surrounding(&self, im_context: &Self::Type) -> bool {
+        self.parent_retrieve_surrounding(im_context)
+    }
+    fn set_client_widget<P: IsA<Widget>>(&self, im_context: &Self::Type, widget: Option<&P>) {
+        self.parent_set_client_widget(im_context, widget)
+    }
+    fn set_cursor_location(&self, im_context: &Self::Type) -> Option<gdk::Rectangle> {
+        self.parent_set_cursor_location(im_context)
+    }
```
bilelmoussaoui, line 50: I'm not sure whether this method should return the area or not...

sdroege: No, it's the other way around. The caller passes a rectangle (not nullable!) in. See `gtk_im_context_set_cursor_location()` docs.

gtk4/src/subclass/im_context.rs
```diff
@@ -0,0 +1,510 @@
...
+    let wrap: Borrowed<IMContext> = from_glib_borrow(ptr);
+
+    let area = imp.set_cursor_location(wrap.unsafe_cast_ref());
+    *areaptr = *area.to_glib_none().0;
+}
+
+unsafe extern "C" fn im_context_set_surrounding<T: IMContextImpl>(
+    ptr: *mut ffi::GtkIMContext,
+    textptr: *const libc::c_char,
+    _length: i32,
```
sdroege, line 490: You probably want to make use of this

bilelmoussaoui: Use it as in creating a string from textptr for only the length of bytes passed? that makes sense i guess

gtk4/src/subclass/im_context.rs
```diff
@@ -0,0 +1,508 @@
...
+unsafe extern "C" fn im_context_set_surrounding<T: IMContextImpl>(
+    ptr: *mut ffi::GtkIMContext,
+    textptr: *const libc::c_char,
+    length: i32,
+    cursor_index: i32,
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<IMContext> = from_glib_borrow(ptr);
+    let text: GString = FromGlibContainer::from_glib_none_num(textptr, length as usize);
```
bilelmoussaoui, line 495: The C code does uses `g_strndup (text, len);` to only get the passed length, does it make sense to a transfer: none here?

sdroege: What C code? And why not use `from_glib_borrow()` here?

bilelmoussaoui: But then I would have to convert the gstring to bytes to only take `length` bytes & convert it back to a gstring?

sdroege: Pff... hmm.

```rust
let s = GString::from_glib_borrow(textptr);
f(&s[..length]);
```

or something like that maybe? The `from_glib_none()` there does a full copy of the string :)

bilelmoussaoui: That worked fine :) thanks!

gtk4/src/subclass/im_context.rs
```diff
@@ -0,0 +1,512 @@
...
+    fn reset(&self, im_context: &Self::Type) {
+        self.parent_reset(im_context)
+    }
+    fn retrieve_surrounding(&self, im_context: &Self::Type) -> bool {
+        self.parent_retrieve_surrounding(im_context)
+    }
+    fn set_client_widget<P: IsA<Widget>>(&self, im_context: &Self::Type, widget: Option<&P>) {
+        self.parent_set_client_widget(im_context, widget)
+    }
+    fn set_cursor_location(&self, im_context: &Self::Type, area: &mut gdk::Rectangle) {
```
sdroege, line 48:
```suggestion
    fn set_cursor_location(&self, im_context: &Self::Type, area: &gdk::Rectangle) {
```

etc according to the docs

gtk4/src/subclass/im_context.rs
```diff
@@ -0,0 +1,512 @@
...
+}
+
+impl<T: IMContextImpl> IMContextImplExt for T {
+    fn parent_commit(&self, im_context: &Self::Type, string: &str) {
+        unsafe {
+            let data = T::type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut ffi::GtkIMContextClass;
+            let f = (*parent_class)
+                .commit
+                .expect("No parent class impl for \"commit\"");
```
sdroege, line 86: This is a signal. That's usually a good sign for a vfunc to be optional :)

gtk4/src/subclass/im_context.rs
```diff
@@ -0,0 +1,512 @@
...
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<IMContext> = from_glib_borrow(ptr);
+
+    let (text, attrs, cursor_idx) = imp.get_preedit_string(wrap.unsafe_cast_ref());
+
+    *text_ptr = text.to_glib_none().0;
+    *cursor_index_ptr = cursor_idx;
+    *attrs_ptr = attrs.to_glib_none().0;
```
sdroege, line 398: The docs say that these are all `transfer full`

gtk4/src/subclass/im_context.rs
```diff
@@ -0,0 +1,512 @@
...
+    ptr: *mut ffi::GtkIMContext,
+    text_ptr: *mut *mut libc::c_char,
+    cursor_index_ptr: *mut libc::c_int,
+) -> glib::ffi::gboolean {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<IMContext> = from_glib_borrow(ptr);
+
+    if let Some((text, cursor_idx)) = imp.get_surrounding(wrap.unsafe_cast_ref()) {
+        *text_ptr = text.to_glib_none().0;
```
sdroege, line 411: `transfer full`

gtk4/src/subclass/im_context.rs
```diff
@@ -0,0 +1,512 @@
...
+        *cursor_index_ptr = cursor_idx;
+        return true.to_glib();
+    }
+    false.to_glib()
```
sdroege, line 415: Maybe to be friendly you want to set `text_ptr` and `cursor_index_ptr` to `NULL` / `0` in this case

gtk4/src/subclass/im_context.rs
```diff
@@ -0,0 +1,512 @@
...
+}
+
+unsafe extern "C" fn im_context_set_cursor_location<T: IMContextImpl>(
+    ptr: *mut ffi::GtkIMContext,
+    areaptr: *mut gdk::ffi::GdkRectangle,
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<IMContext> = from_glib_borrow(ptr);
+    let mut area = gdk::Rectangle::uninitialized();
```
sdroege, line 479: The caller provides the area rect and you just pass a read-only reference of it onwards

gtk4/src/subclass/im_context.rs
```diff
@@ -0,0 +1,512 @@
...
+    cursor_index: i32,
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<IMContext> = from_glib_borrow(ptr);
+    let text: Borrowed<GString> = from_glib_borrow(textptr);
+
+    imp.set_surrounding(
+        wrap.unsafe_cast_ref(),
+        &text[0..(length as usize)],
```
sdroege, line 498: > the length of text , or -1 if text is nul-terminated

So you need to handle `-1` here differently. `-1 == usize::MAX` would panic here now :)

sdroege (changes requested):
Generally you have to compare things to the docs a bit more :)

sdroege (approved):

PR #124, opened 2020-12-17
Regenerate with gtk 4.0.0

bilelmoussaoui:


gdk4/src/auto/toplevel_layout.rs
```diff
@@ -39,11 +40,19 @@ impl ToplevelLayout {
     }
 
     #[doc(alias = "gdk_toplevel_layout_get_fullscreen")]
-    pub fn get_fullscreen(&self) -> bool {
+    pub fn get_fullscreen(&self) -> Option<bool> {
```
bilelmoussaoui, line 43: Hmm, what makes gir mis-regenerates this one? it should only return a bool per the docs https://developer.gnome.org/gdk4/stable/GdkToplevelLayout.html#gdk-toplevel-layout-get-fullscreen

gdk4/src/auto/toplevel_layout.rs
```diff
@@ -57,10 +66,11 @@ impl ToplevelLayout {
     }
 
     #[doc(alias = "gdk_toplevel_layout_get_maximized")]
-    pub fn get_maximized(&self) -> bool {
+    pub fn get_maximized(&self, maximize: bool) -> bool {
```
sdroege, line 69: This is wrong. `maximize` in the C API is a pointer to a bool, it's an out parameter. Missing annotation

gdk4/src/auto/toplevel_layout.rs
sdroege: No, this is completely correct and the docs are outdated: https://gitlab.gnome.org/GNOME/gtk/-/blob/master/gdk/gdktoplevellayout.c#L261

gdk4/src/auto/toplevel_layout.rs
sdroege: Should work the same as `get_fullscreen()` basically

gdk4/src/auto/toplevel_layout.rs
bilelmoussaoui: Ah yes my bad. I somehow missed it's an out param.

gdk4/src/auto/toplevel_layout.rs
bilelmoussaoui: Seems to be fixed already in-tree, I have re-updated the gir-files in https://github.com/gtk-rs/gir-files/pull/71

sdroege (approved):

PR #125, opened 2020-12-17
Use manual gdk::Rectangle from gdk3-rs bindings

BrainBlasted:
This is what we need to do instead of generating bindings
for GdkRectangle.


gdk4/src/rectangle.rs
```diff
@@ -0,0 +1,167 @@
...
+#[repr(C)]
+pub struct Rectangle {
+    pub x: i32,
+    pub y: i32,
+    pub width: i32,
+    pub height: i32,
+}
+
+impl Rectangle {
+    pub fn intersect(&self, other: &Rectangle) -> Option<Rectangle> {
```
sdroege, line 19: You forgot the `contains_point()` function

BrainBlasted: Added contains_point and equal

gdk4/src/rectangle.rs
```diff
@@ -0,0 +1,190 @@
...
+            from_glib(ffi::gdk_rectangle_contains_point(
+                self.to_glib_none().0,
+                x,
+                y,
+            ))
+        }
+    }
+
+    #[doc(alias = "gdk_rectangle_equal")]
+    pub fn equal(&self, rect2: &Rectangle) -> bool {
```
sdroege, line 31: There's a reason why I didn't mention equal :) That's equivalent to the `#[derive(PartialEq)]` you already have, and having a separate `equal()` function (in addition to the `eq()` from `PartialEq`) is just confusing

sdroege (approved):

PR #132, opened 2020-12-19
gtk: add SelectionModel subclassing support

kavanmevada:


gtk4/src/subclass/selection_model.rs
```diff
@@ -0,0 +1,173 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::Bitset;
+use crate::SelectionModel;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::Cast;
+
+pub trait SelectionModelImpl: ObjectImpl {
```
sdroege, line 9:
```suggestion
pub trait SelectionModelImpl: ObjectImpl + gio::ListModelImpl {
```

See docs:
> GtkSelectionModel requires GListModel and GObject.

gtk4/src/subclass/selection_model.rs
```diff
@@ -0,0 +1,173 @@
...
+    fn get_selection_in_range(&self, model: &Self::Type, position: u32, n_items: u32) -> Bitset;
```
sdroege, line 10: This one has a default impl, as have many others. See https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/8549046994146118850bae111bf1c7e942d55a1b/gstreamer/src/subclass/child_proxy.rs#L10-26 how to handle that

sdroege (commented):
Looks good otherwise

kavanmevada:
Thanks, @sdroege

gtk4/src/subclass/selection_model.rs
sdroege: Other functions here also have default implementations, you only added it for this one (but that looks correct) :)

sdroege:
@kavanmevada Why did you close this PR? Seems like not much is missing here to get it merged :)

kavanmevada:
Sorry I can't keep doing things like this not good for me.

I truly appreciate your all reviews and your valuable time in project like
this, and thanks again for reviewing all my commits so fast, thanks a lot ❤️

There are some people who can finish all my half done work.

sdroege:
@kavanmevada Thanks for all your work and sorry to hear that you won't continue being involved here. Good luck with whatever else you're going to spend your time on :)

Let's re-open this then. @bilelmoussaoui, do you want to finish it?

bilelmoussaoui:
Sure thing

PR #136, opened 2020-12-20
update bitflags to 1.2

bvinc:
PR gtk-rs/gtk-rs#230 exists to update bitflags to 1.2.  This should be applied at the same time so cargo can pick a proper version of bitflags.


bilelmoussaoui:
Thanks! Never mind the build failure, I will fix it ;)

bvinc:
I don't think this is necessary anymore.  1.2 should be automatically selected by cargo.

PR #137, opened 2020-12-21
gtk: add more WidgetClass methods

bilelmoussaoui:
part of #111

### DIFF
--- a/examples/src/bin/video_player.rs
+++ b/examples/src/bin/video_player.rs
@@ -52,6 +52,24 @@ mod imp {
 
         fn class_init(klass: &mut Self::Class) {
             Self::bind_template(klass);
+            klass.install_action(
+                "win.open",
+                None,
+                move |win, _action_name, _action_target| {
+                    let self_ = imp::VideoPlayerWindow::from_instance(&win);
+                    self_.dialog.set_transient_for(Some(win));
+                    self_
+                        .dialog
+                        .connect_response(clone!(@weak win => move |d, response| {
+                            if response == gtk::ResponseType::Accept {
+                                win.set_video(d.get_file().unwrap());
+                            }
+                            d.destroy();
+                        }));
+
+                    self_.dialog.show();
+                },
+            );
         }
 
         fn instance_init(obj: &glib::subclass::InitializingObject<Self>) {
@@ -59,13 +77,7 @@ mod imp {
         }
     }
 
-    impl ObjectImpl for VideoPlayerWindow {
-        fn constructed(&self, obj: &Self::Type) {
-            obj.init_actions();
-            self.parent_constructed(obj);
-        }
-    }
-
+    impl ObjectImpl for VideoPlayerWindow {}
     impl WidgetImpl for VideoPlayerWindow {}
     impl WindowImpl for VideoPlayerWindow {}
     impl ApplicationWindowImpl for VideoPlayerWindow {}
@@ -79,24 +91,6 @@ glib::wrapper! {
 impl VideoPlayerWindow {
     pub fn new<P: glib::IsA<gtk::Application>>(app: &P) -> Self {
         glib::Object::new(&[("application", app)]).expect("Failed to create VideoPlayerWindow")
-    }
-
-    fn init_actions(&self) {
-        let open = gio::SimpleAction::new("open", None);
-
-        open.connect_activate(clone!(@weak self as win => move |_, _| {
-            let self_ = imp::VideoPlayerWindow::from_instance(&win);
-            self_.dialog.set_transient_for(Some(&win));
-            self_.dialog.connect_response(clone!(@weak win => move |d, response| {
-                if response == gtk::ResponseType::Accept {
-                    win.set_video(d.get_file().unwrap());
-                }
-                d.destroy();
-            }));
-
-            self_.dialog.show();
-        }));
-        self.add_action(&open);
     }
 
     fn set_video(&self, video: gio::File) {
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -865,6 +865,7 @@ unsafe extern "C" fn widget_unroot<T: WidgetImpl>(ptr: *mut ffi::GtkWidget) {
 }
 
 pub unsafe trait WidgetClassSubclassExt: ClassStruct {
+    #[doc(alias = "gtk_widget_class_set_template")]
     fn set_template_bytes(&mut self, template: &glib::Bytes) {
         unsafe {
             let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
@@ -882,6 +883,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         self.set_template_bytes(&template_bytes);
     }
 
+    #[doc(alias = "gtk_widget_class_set_template_from_resource")]
     fn set_template_from_resource(&mut self, resource_name: &str) {
         unsafe {
             let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
@@ -960,6 +962,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_set_template_scope")]
     fn set_template_scope<S: IsA<BuilderScope>>(&mut self, scope: &S) {
         unsafe {
             let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
@@ -967,6 +970,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_bind_template_child_full")]
     fn bind_template_child(&mut self, name: &str) {
         unsafe {
             let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
@@ -979,6 +983,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_add_shortcut")]
     fn add_shortcut(&mut self, shortcut: &Shortcut) {
         unsafe {
             let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
@@ -986,6 +991,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_install_property_action")]
     fn install_property_action(&mut self, action_name: &str, property_name: &str) {
         unsafe {
             let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
@@ -1010,6 +1016,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_set_activate_signal")]
     fn set_activate_signal(&mut self, signal_id: SignalId) {
         unsafe {
             let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
@@ -1017,6 +1024,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_set_activate_signal_from_name")]
     fn set_activate_signal_from_name(&mut self, signal_name: &str) {
         unsafe {
             let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
@@ -1027,6 +1035,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_set_layout_manager_type")]
     fn set_layout_manager_type<T: IsA<LayoutManager>>(&mut self) {
         unsafe {
             let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
@@ -1034,6 +1043,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_get_layout_manager_type")]
     fn get_layout_manager_type(&self) -> glib::Type {
         unsafe {
             let widget_class = self as *const _ as *mut ffi::GtkWidgetClass;
@@ -1041,6 +1051,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_set_css_name")]
     fn set_css_name(&mut self, name: &str) {
         unsafe {
             let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
@@ -1048,6 +1059,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_get_css_name")]
     fn get_css_name(&self) -> glib::GString {
         unsafe {
             let widget_class = self as *const _ as *mut ffi::GtkWidgetClass;
@@ -1055,6 +1067,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_set_accessible_role")]
     fn set_accessible_role(&mut self, role: AccessibleRole) {
         unsafe {
             let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
@@ -1062,6 +1075,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_get_accessible_role")]
     fn get_accessible_role(&self) -> AccessibleRole {
         unsafe {
             let widget_class = self as *const _ as *mut ffi::GtkWidgetClass;
@@ -1070,6 +1084,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
     }
 
     #[allow(clippy::missing_safety_doc)]
+    #[doc(alias = "gtk_widget_class_bind_template_child_full")]
     unsafe fn bind_template_child_with_offset<T>(
         &mut self,
         name: &str,
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -1,15 +1,21 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use glib::prelude::*;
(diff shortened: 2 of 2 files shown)


gtk4/src/subclass/widget.rs
```diff
@@ -895,6 +896,118 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+    ) {
+        unsafe {
+            unsafe extern "C" fn activate_trampoline<
+                F: Fn(&Widget, &str, Option<&glib::Variant>) + 'static,
+            >(
+                this: *mut ffi::GtkWidget,
+                action: *const libc::c_char,
+                parameter: *mut glib::ffi::GVariant,
+            ) {
+                f(
```
bilelmoussaoui, line 914: This is almost working, but I have no idea how can i pass the callback to here as there's no way to pass user_data to the callback, any ideas?

gtk4/src/subclass/widget.rs
```diff
@@ -895,6 +896,118 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+            let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
+            ffi::gtk_widget_class_install_property_action(
+                widget_class,
+                action_name.to_glib_none().0,
+                property_name.to_glib_none().0,
+            );
+        }
+    }
+
+    fn set_activate_signal(&mut self, signal: SignalHandlerId) {
```
sdroege, line 982: This is a signal id, not a signal **handler** id :) The return value of `g_signal_new()` basically.

gtk4/src/subclass/widget.rs
```diff
@@ -895,6 +896,118 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+        unsafe {
+            let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
+            ffi::gtk_widget_class_set_activate_signal(widget_class, signal.to_glib() as u32);
+        }
+    }
+
+    fn get_activate_signal(&self) -> Option<SignalHandlerId> {
```
sdroege, line 989: Same here

gtk4/src/subclass/widget.rs
sdroege: We don't return those from `ObjectClassSubclassExt::add_signal()` currently, but my plan from https://github.com/gtk-rs/gtk-rs/pull/31 would also involve that

sdroege: So for now don't add these, I guess

gtk4/src/subclass/widget.rs
```diff
@@ -895,6 +896,118 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+            let handler_id = ffi::gtk_widget_class_get_activate_signal(widget_class);
+            if handler_id == 0 {
+                None
+            } else {
+                Some(from_glib(handler_id as u64))
+            }
+        }
+    }
+
+    fn set_activate_signal_from_name(&mut self, signal_name: &str) {
```
sdroege, line 1001: This one should be fine

gtk4/src/subclass/widget.rs
```diff
@@ -895,6 +896,118 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+                activate_func,
+            );
+        }
+    }
+
+
+    fn query_action(
+        &self,
+        index: u32,
+    ) -> Option<(glib::Type, glib::GString, Option<glib::VariantType>, glib::GString)> {
```
sdroege, line 939: `property_name` (the last string) can be `NULL` according to the code. `parameter_type` (the `VariantType`) maybe not.

This function needs better documentation on the C side and annotations.

Also this function is basically an iterator (you might want to add one)

gtk4/src/subclass/widget.rs
sdroege: Well this sucks, it's broken API. The only thing you can do here is to install a `fn(*mut gtk::ffi::GtkWidget, *const c_char, *mut glib::ffi::GVariant)` here.

This needs some new API added to GTK that takes a `user_data`, and if they care about dynamic bindings also a `GDestroyNotify` (that would never be called).

The annotation of the C function is also completely wrong: ` * @activate: (scope call): callback to use when the action is activated`. This is not `scope call`, it's a callback that has to be forever valid.

tl;dr: Fix it first in C by adding some new API there. This function as-is is completely useless for bindings.

sdroege: See also https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/1732#note_989762 . I'm surprised that this wasn't noticed as part of the review of that MR

sdroege: You could probably work around it by storing a hash table in the `interface_data` in `ObjectSubclass::type_data()` here that maps from the `action_name` (second parameter here and in the callback) to the `user_data` (the `Box<dyn Fn>` closure).

See [here](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/3428e5c0c753f7eae8a60dcae7cbf30c9828cd88/gstreamer/src/subclass/uri_handler.rs#L26-34) and [here](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/3428e5c0c753f7eae8a60dcae7cbf30c9828cd88/gstreamer/src/subclass/uri_handler.rs#L53-54) for some usage of that.

It's all rather awkward though.

@ids1024 FYI, another usage of the `type_data()` that is potentially accessed a lot

gtk4/src/subclass/widget.rs
bilelmoussaoui: Aha, my bad. I will disable those for now

gtk4/src/subclass/widget.rs
bilelmoussaoui: I have went ahead and opened an issue at https://gitlab.gnome.org/GNOME/gtk/-/issues/3498 for now

bilelmoussaoui: I went ahead and disabled this for now; i will try adding a gclosure alternative to gtk first & use it here in the future

gtk4/src/subclass/widget.rs
bilelmoussaoui: I have disabled this as well as I don't really need it for now

gtk4/src/subclass/widget.rs
sdroege: Can you create an issue about that? Seems rather fundamental :)

gtk4/src/subclass/widget.rs
sdroege: Can you create issues for that? On gtk (annotations) and here?

gtk4/src/subclass/widget.rs
bilelmoussaoui: There's an issue already to track all the widget class methods, see #111

sdroege (approved):
Looks good to me otherwise, but it's very small now :)

Shouldn't at least `gtk_widget_class_install_action()` be added?

bilelmoussaoui:
> Looks good to me otherwise, but it's very small now :)
> 
> Shouldn't at least `gtk_widget_class_install_action()` be added?

Indeed it's, let me try to add that one tonight :)

sdroege:
Let me know when this is ready for another review :)

gtk4/src/subclass/widget.rs
bilelmoussaoui: Going back to this, I'm a bit stuck on how to properly do thing. Here's what I have currently more or less 
```rust
fn install_action<F: Fn(&Widget, &str, Option<&Variant>) + 'static>(
        &mut self,
        action_name: &str,
        parameter_type: Option<&str>,
        activate: F,
    ) {
        // We store the activate callbacks in a HashMap<action_name, activate>
        // so that we can retrieve f later on the activate_trampoline call
        let mut data = <Self::Type as ObjectSubclass>::type_data();
        let data = data.as_mut();

        #[derive(Debug)]
        struct Actions(pub(crate) HashMap<String, glib::ffi::gpointer>);
        unsafe impl Sync for Actions {}
        unsafe impl Send for Actions {}

        let f: Box_<F> = Box_::new(activate);

        let actions = data
            .get_class_data_mut::<Actions>(Widget::static_type())
            .unwrap();
        actions.0.insert(action_name.to_string(), Box::into_raw(f));

        unsafe extern "C" fn activate_trampoline<F>(
            this: *mut ffi::GtkWidget,
            action_name: *const libc::c_char,
            parameter: *mut glib::ffi::GVariant,
        ) {
            let action_name = GString::from_glib_borrow(action_name);

            let mut data = <Self::Type as ObjectSubclass>::type_data();
            let data = data.as_mut();

            let actions = data
                .get_class_data::<Actions>(Widget::static_type())
                .unwrap();
            let activate_callback = actions.0.get(&action_name.to_string()).unwrap();

            let f: &F = &*(activate_callback as *const F);
            f(
                &from_glib_borrow(this),
                &action_name,
                Option::<Variant>::from_glib_borrow(parameter)
                    .as_ref()
                    .as_ref(),
            )
        }

        unsafe {
            let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
            ffi::gtk_widget_class_install_action(
                widget_class,
                action_name.to_glib_none().0,
                parameter_type.to_glib_none().0,
                Some(transmute::<
                    _,
                    unsafe extern "C" fn(*mut ffi::GtkWidget, *const i8, *mut glib::ffi::GVariant),
                >(activate_trampoline::<F> as *const ())),
            );
        }
    }
```

The issue I'm having currently is how can I access the TypeData inside the activate_tramploline as `Self` isn't available in that scope
```rust
            let mut data = <Self::Type as ObjectSubclass>::type_data();
            let data = data.as_mut();
```
Any ideas? :)

bilelmoussaoui: Oh, maybe I just need to cast `this`? :)

gtk4/src/subclass/widget.rs
```diff
@@ -895,6 +967,43 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+            );
+        }
+    }
+
+    fn get_activate_signal(&self) -> SignalId {
+        unsafe {
+            let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
+            let signal_id = ffi::gtk_widget_class_get_activate_signal(widget_class);
+            let type_ = Self::Type;
+            SignalId(type_, signal_id)
```
bilelmoussaoui, line 986: We can't construct the SignalId as the constructor is private, I wonder if it makes sense to make it public?

sdroege: Yes, feel free to add a public, unsafe constructor for this. It probably also makes sense to have public safe variants of `g_signal_query`, `g_signal_lookup`, `g_signal_name`.

Also the type is only really in the `SignalId` for checking during `emit`. The signal ids are unique per process, not just unique per type.

gtk4/src/subclass/widget.rs
sdroege: Pass `Self` as a second type parameter to the trampoline. `activate_trampoline<F, S>` :)

That seems like a good idea anyway, then you can use `Self::Type` instead of `gtk::Widget` as first parameter of the callback.

gtk4/src/subclass/widget.rs
```diff
@@ -876,6 +880,82 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_install_action")]
+    fn install_action<
+        F: Fn(&Widget, &str, Option<&Variant>) + 'static,
+        S: WidgetImpl + ObjectSubclass,
```
bilelmoussaoui, line 886: I'm not sure whether i need something similar here, or if there's a better way to pass Self as S to trampoline's scope?

gtk4/src/subclass/widget.rs
```diff
@@ -876,6 +880,82 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+
+        let f: Box_<F> = Box_::new(activate);
+
+        let actions = data
+            .get_class_data_mut::<Actions>(Widget::static_type())
+            .unwrap();
+        actions.0.insert(action_name.to_string(), Box::into_raw(f));
+
+        unsafe extern "C" fn activate_trampoline<F, S>(
+            this: S::Type,
```
bilelmoussaoui, line 911: This currently fails because `the associated type 'Type' is not found` for S, which is defined as `ObjectSubclass`

gtk4/src/subclass/widget.rs
sdroege: You could require that `Self: ClassStruct` and then go via that. See the impl line further below `unsafe impl<T: ClassStruct> WidgetClassSubclassExt for T where T::Type: WidgetImpl {}`.

gtk4/src/subclass/widget.rs
sdroege: You need to add a bound for `ObjectSubclass` to the function for that to work. Type parameters on nested functions need to repeat all bounds they need, you don't get the type parameters of the surrounding scope for free :)

Nested functions are just like normal functions elsewhere, the only difference is the scoping of the function itself.

bilelmoussaoui:
I managed to get the `install_action` working :tada: This should be ready for another round of review

gtk4/src/subclass/widget.rs
sdroege: This one is still pending an issue or two :)

examples/src/bin/video_player.rs
```diff
@@ -52,20 +52,32 @@ mod imp {
 
         fn class_init(klass: &mut Self::Class) {
             Self::bind_template(klass);
+            klass.install_action(
```
sdroege, line 55: We could also consider doing this like for signals with a builder, etc. But that can be done at another time

Actions are in the end very similar to signals.

examples/src/bin/video_player.rs
```diff
@@ -52,20 +52,32 @@ mod imp {
...
+                move |win, _action_name, _action_target| {
+                    let self_ = imp::VideoPlayerWindow::from_instance(&win);
+                    self_.dialog.set_transient_for(Some(win));
+                    self_
+                        .dialog
+                        .connect_response(clone!(@weak win => move |d, response| {
+                            if response == gtk::ResponseType::Accept {
+                                win.set_video(d.get_file().unwrap());
+                            }
+                            d.destroy();
```
sdroege, line 67:
```suggestion
                            d.close();
```

Maybe? `destroy()` seems so destructive

gtk4/src/subclass/widget.rs
```diff
@@ -880,6 +894,84 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+
+                let mut data = <S::Type as ObjectSubclassType>::type_data();
+                let data = data.as_mut();
+
+                let actions = data
+                    .get_class_data::<Actions>(Widget::static_type())
+                    .unwrap();
+                let activate_callback =
+                    *actions.0.get(&action_name.to_string()).unwrap_or_else(|| {
+                        panic!("Action name '{}' was not found", action_name.to_string());
```
sdroege, line 941:
```suggestion
                        panic!("Action name '{}' was not found", &action_name);
```

Or `as_str()` or so

gtk4/src/subclass/widget.rs
```diff
@@ -880,6 +894,84 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+            ) where
+                S: ClassStruct,
+                <S as ClassStruct>::Type: ObjectSubclass,
+                F: Fn(&<<S as ClassStruct>::Type as ObjectSubclass>::Type, &str, Option<&Variant>)
+                    + 'static,
+            {
+                let action_name = GString::from_glib_borrow(action_name);
+
+                let mut data = <S::Type as ObjectSubclassType>::type_data();
+                let data = data.as_mut();
```
sdroege, line 934: mutable shouldn't be needed here

gtk4/src/subclass/widget.rs
```diff
@@ -892,48 +984,95 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+                property_name.to_glib_none().0,
+            );
+        }
+    }
+
+    #[doc(alias = "gtk_widget_class_get_activate_signal")]
+    fn get_activate_signal(&self) -> SignalId {
+        unsafe {
+            let widget_class = self as *const _ as *mut ffi::GtkWidgetClass;
+            let signal_id = ffi::gtk_widget_class_get_activate_signal(widget_class);
```
sdroege, line 1011: Can this return `0`, i.e. an invalid signal ID if none was set before?

gtk4/src/subclass/widget.rs
```diff
@@ -892,48 +984,95 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+            from_glib(signal_id)
+        }
+    }
+
+    #[doc(alias = "gtk_widget_class_set_activate_signal")]
+    fn set_activate_signal(&mut self, signal_id: SignalId) {
+        unsafe {
+            let widget_class = self as *mut _ as *mut ffi::GtkWidgetClass;
+            ffi::gtk_widget_class_set_activate_signal(widget_class, signal_id.to_glib())
```
sdroege, line 1020: I have no idea how this is supposed to be used :) An example for that would be good at some point, and maybe we can invent an easier API for it in Rust.

sdroege (commented):
Looks good otherwise

examples/src/bin/video_player.rs
bilelmoussaoui: There's no close method for Native dialogs

gtk4/src/subclass/widget.rs
bilelmoussaoui: Some widgets can be activated like buttons, check buttons, toggle buttons, switches. Using this function allows you to change the signal to be used for activation instead of the typical "activate". 

I think it's good as is for now

examples/src/bin/video_player.rs
bilelmoussaoui: Ah, yes, that can actually work. I will leave that to someone interested in improving how signals can be created/used today.

gtk4/src/subclass/widget.rs
bilelmoussaoui: Ah, indeed, it can return 0 if the widget doesn't have an activation signal set.

gtk4/src/subclass/widget.rs
bilelmoussaoui: The first issue is #111, the upstream one https://gitlab.gnome.org/GNOME/gtk/-/issues/3757

examples/src/bin/video_player.rs
sdroege: Can you create an issue?

PR #139, opened 2020-12-21
gdk: add Paintable subclassing support

bilelmoussaoui:
part of #56

### DIFF
--- a/examples/src/bin/custom_paintable.rs
+++ b/examples/src/bin/custom_paintable.rs
@@ -1,0 +1,147 @@
+//! # Custom GdkPaintable
+//!
+//! This example shows how to create a custom gdk::Paintable
+//! that renders a square with a linear color gradient
+
+use gtk::prelude::*;
+use gtk::subclass::prelude::*;
+use gtk::{gdk, glib, graphene, gsk};
+use std::env::args;
+
+mod imp {
+    use super::*;
+    use glib::subclass;
+    pub struct CustomPaintable {}
+
+    impl ObjectSubclass for CustomPaintable {
+        const NAME: &'static str = "CustomPaintable";
+        type Type = super::CustomPaintable;
+        type ParentType = glib::Object;
+        type Instance = subclass::simple::InstanceStruct<Self>;
+        type Class = subclass::simple::ClassStruct<Self>;
+
+        glib::object_subclass!();
+
+        fn type_init(type_: &mut subclass::InitializingType<Self>) {
+            type_.add_interface::<gdk::Paintable>();
+        }
+
+        fn new() -> Self {
+            Self {}
+        }
+    }
+
+    impl ObjectImpl for CustomPaintable {}
+
+    impl PaintableImpl for CustomPaintable {
+        fn get_flags(&self, _paintable: &Self::Type) -> gdk::PaintableFlags {
+            // Fixed size
+            gdk::PaintableFlags::SIZE
+        }
+
+        fn get_intrinsic_width(&self, _paintable: &Self::Type) -> i32 {
+            200
+        }
+
+        fn get_intrinsic_height(&self, _paintable: &Self::Type) -> i32 {
+            200
+        }
+
+        fn snapshot(
+            &self,
+            _paintable: &Self::Type,
+            snapshot: &gdk::Snapshot,
+            width: f64,
+            height: f64,
+        ) {
+            let snapshot = snapshot.downcast_ref::<gtk::Snapshot>().unwrap();
+            // Draw a linear gradient
+            snapshot.append_linear_gradient(
+                &graphene::Rect::new(0_f32, 0_f32, width as f32, height as f32),
+                &graphene::Point::new(0f32, 0f32),
+                &graphene::Point::new(width as f32, height as f32),
+                &[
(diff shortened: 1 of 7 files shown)


bilelmoussaoui:
I can't seem to get the paintable to render what it should... I will have an other look tomorrow

examples/src/bin/custom_paintable.rs
```diff
@@ -0,0 +1,100 @@
...
+        type Type = super::CustomPaintable;
+        type ParentType = glib::Object;
+        type Instance = subclass::simple::InstanceStruct<Self>;
+        type Class = subclass::simple::ClassStruct<Self>;
+
+        glib::object_subclass!();
+
+        fn type_init(type_: &mut subclass::InitializingType<Self>) {
+            type_.add_interface::<gdk::Paintable>();
+        }
```
sdroege, line 27:
```suggestion
        }

```

examples/src/bin/custom_paintable.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    impl ObjectImpl for CustomPaintable {}
+
+    impl PaintableImpl for CustomPaintable {
+        fn get_flags(&self, _paintable: &Self::Type) -> gdk::PaintableFlags {
+            // Fixed size
+            gdk::PaintableFlags::SIZE
+        }
```
sdroege, line 39: Some more space between functions please :)

examples/src/bin/custom_paintable.rs
```diff
@@ -0,0 +1,100 @@
...
+            200
+        }
+        fn snapshot(
+            &self,
+            _paintable: &Self::Type,
+            snapshot: &gdk::Snapshot,
+            width: f64,
+            height: f64,
+        ) {
+            let gtk_snapshot = snapshot.downcast_ref::<gtk::Snapshot>().unwrap();
```
sdroege, line 53: I usually just shadow old variables if it's a downcast. You won't need the original one anymore

```suggestion
            let snapshot = snapshot.downcast_ref::<gtk::Snapshot>().unwrap();
```

sdroege: Also it's a bit strange: why do you assume that this is a `gtk::Snapshot`? Is this somehow guaranteed and nothing else can be passed here for some reason, so failure to cast does not have to be handled?

gdk4/src/subclass/paintable.rs
```diff
@@ -0,0 +1,173 @@
...
+
+pub trait PaintableImpl: ObjectImpl {
+    fn get_current_image(&self, paintable: &Self::Type) -> Paintable {
+        unsafe {
+            let type_ = ffi::gdk_paintable_get_type();
+            let iface = glib::gobject_ffi::g_type_default_interface_ref(type_)
+                as *mut ffi::GdkPaintableInterface;
+            assert!(!iface.is_null());
+
+            let ret = ((*iface).get_current_image.as_ref().unwrap())(
```
sdroege, line 16: The default implementation is not too useful:

```cpp
static GdkPaintable *
gdk_paintable_default_get_current_image (GdkPaintable *paintable)
{
  g_warning ("FIXME: implement by snapshotting at default size and returning a GskRendererNodePaintable");

  return paintable;
}
```

AFAIU this is a lazy way to say that this virtual method is actually mandatory, or not?

sdroege:
> I can't seem to get the paintable to render what it should... I will have an other look tomorrow

I don't see anything wrong, must be wrong API usage :) Best to check with the GTK people.

gdk4/src/subclass/paintable.rs
bilelmoussaoui: If your paintable's content is constant & doesn't change, your return self which is the default implementation, I'm not sure why that warning's there

examples/src/bin/custom_paintable.rs
bilelmoussaoui: I think it's mostly because Snapshot needs to use Gsk types which are not defined in gdk, hence the use of a new type on gtk for that, see https://github.com/aeldemery/gtk4_list_colors/blob/master/sources/ColorWidget.vala#L57 which is a rewrite of a gtk demo in Vala

gdk4/src/subclass/paintable.rs
sdroege: Worth a GTK bug report IMHO. There should never be warnings in normal circumstances.

examples/src/bin/custom_paintable.rs
sdroege: Seems like a strange API choice but whatever :)

bilelmoussaoui:
It was a gdk::RGBA change that we forgot, it works fine now 
image

sdroege:
Maybe make it a rainbox or so instead of a depressing black/white gradient :D

bilelmoussaoui:
Replaced it with something better :p 
image

sdroege:
That's better indeed. Let me know when the whole thing is ready for another review!

bilelmoussaoui:
It's ready, at least from my side :) The next step is figuring out how all the fancy shaders can be used

PR #140, opened 2020-12-22
Fix weird indents in toml files

GuillaumeGomez:
This is now checked by `checker` so unless this PR is merged, CI will fail. ;)


bilelmoussaoui:
almost good :p thanks !

Issue #23, opened 2020-10-21
Implement Gtk.CustomLayout


bilelmoussaoui:
The `gtk_custom_layout_new` needs some manual bindings

ebassi:
I don't recommend implementing CustomLayout. It's a C API for transitioning existing widgets from the `GtkWidget` virtual functions of GTK3 to the virtual functions in `GtkLayoutManager`. Writing subclasses, even private ones, in Rust is definitely easier than doing so in C.

bilelmoussaoui:
Thanks @ebassi for confirming that's something we shouldn't include. It's already possible to subclass `GtkLayoutManager` which should cover the custom layout manager use case

PR #145, opened 2021-01-03
Ignore CustomLayout

A6GibKm:
Fixes: https://github.com/gtk-rs/gtk4-rs/issues/23


gtk4/src/custom_layout.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+unsafe extern "C" fn measure_trampoline<G>(
+    widget: *mut ffi::GtkWidget,
+    orientation: ffi::GtkOrientation,
+    for_size: *mut i32,
+    minimum: *mut i32,
+    natural: *mut i32,
+    minimum_baseline: *mut i32,
+    natural_baseline: *mut i32,
```
sdroege, line 43: These are all `out` parameters, or maybe even `inout`. You need to write the result to them

gtk4/src/custom_layout.rs
```diff
@@ -0,0 +1,73 @@
...
+                Some(measure_trampoline::<G>),
+                Some(allocate_trampoline::<H>),
+            ))
+        }
+    }
+}
+
+unsafe extern "C" fn request_mode_trampoline<F>(
+    widget: *mut ffi::GtkWidget,
+    f: glib::ffi::gpointer,
```
sdroege, line 27: All these functions don't take a `user_data` or similar, so you can only really pass `fn()` pointers into them so binding this is rather useless as is. Looks like a bad oversight in the C API, or maybe bindings are instead supposed to properly subclass `gtk::LayoutManager` directly. I guess the latter

bilelmoussaoui:
Per https://github.com/gtk-rs/gtk4-rs/issues/23#issuecomment-766743693, I think we should just add the CustomLayout to the `ignore` list. Do you mind updating the PR to do so? thanks

gtk4/Gir.toml
```diff
@@ -186,6 +185,7 @@ generate = [
 
 # TODO
 ignore = [
+    "Gtk.CustomLayout",
```
bilelmoussaoui, line 188: This needs a comment to that tells why it's ignored, so something like ` # subclass LayoutManager if you need a custom layout`

bilelmoussaoui: also a rebase

A6GibKm: Done.

gtk4/src/auto/custom_layout.rs
```diff
@@ -14,12 +14,7 @@ glib::wrapper! {
     }
 }
```
bilelmoussaoui, line 16: The whole file needs to be removed

Issue #147, opened 2021-01-04
Add Gtk::Orientable interface subclassing support


bilelmoussaoui:
Should be pretty easy

PR #148, opened 2021-01-04
gtk: Add Orientable subclassing support

A6GibKm:
Fixes: https://github.com/gtk-rs/gtk4-rs/issues/147

### DIFF
--- a/examples/src/bin/orientable_subclass.rs
+++ b/examples/src/bin/orientable_subclass.rs
@@ -1,0 +1,176 @@
+use std::cell::RefCell;
+use std::env;
+
+use gtk::glib;
+use gtk::prelude::*;
+use gtk::subclass::prelude::ObjectSubclass;
+
+mod imp {
+    use super::*;
+    use gtk::{glib::translate::ToGlib, subclass::prelude::*};
+
+    #[derive(Debug)]
+    pub struct CustomOrientable {
+        first_label: RefCell<Option<gtk::Widget>>,
+        second_label: RefCell<Option<gtk::Widget>>,
+        orientation: RefCell<gtk::Orientation>,
+    }
+
+    // Every widget that implements Orientable has to define a "orientation"
+    // property like below, gtk::Orientation::Horizontal is a placeholder
+    // for the initial value.
+    //
+    // glib::ParamFlags::CONSTRUCT allows us to set that property the moment
+    // we create a new instance of the widget
+    static PROPERTIES: [glib::subclass::Property; 1] =
+        [glib::subclass::Property("orientation", |name| {
+            glib::ParamSpec::enum_(
+                name,
+                "orientation",
+                "Orientation",
+                gtk::Orientation::static_type(),
+                gtk::Orientation::Horizontal.to_glib(),
+                glib::ParamFlags::READWRITE | glib::ParamFlags::CONSTRUCT,
+            )
+        })];
+
+    impl ObjectSubclass for CustomOrientable {
+        const NAME: &'static str = "ExCustomOrientable";
+        type Type = super::CustomOrientable;
+        type ParentType = gtk::Widget;
+        type Instance = glib::subclass::simple::InstanceStruct<Self>;
+        type Class = glib::subclass::simple::ClassStruct<Self>;
+
+        glib::object_subclass!();
+
+        fn type_init(type_: &mut glib::subclass::InitializingType<Self>) {
+            type_.add_interface::<gtk::Orientable>();
+        }
+
+        fn class_init(klass: &mut Self::Class) {
+            // The layout manager determines how child widgets are laid out.
+            klass.set_layout_manager_type::<gtk::BoxLayout>();
+            klass.install_properties(&PROPERTIES);
+        }
+
+        fn new() -> Self {
+            // Here we set the default orientation.
+            Self {
+                first_label: RefCell::new(None),
+                second_label: RefCell::new(None),
+                orientation: RefCell::new(gtk::Orientation::Horizontal),
+            }
+        }
+    }
+
+    impl ObjectImpl for CustomOrientable {
+        fn constructed(&self, obj: &Self::Type) {
+            self.parent_constructed(obj);
+
+            // Create the children labels.
+            let first_label = gtk::Label::new(Some("Hello"));
+            let second_label = gtk::Label::new(Some("World!"));
+            let layout_manager = obj
+                .get_layout_manager()
+                .unwrap()
+                .downcast::<gtk::BoxLayout>()
+                .unwrap();
+            layout_manager.set_spacing(6);
+            first_label.set_parent(obj);
+            second_label.set_parent(obj);
+            self.first_label
+                .replace(Some(first_label.upcast::<gtk::Widget>()));
+            self.second_label
+                .replace(Some(second_label.upcast::<gtk::Widget>()));
+        }
+
+        fn dispose(&self, _obj: &Self::Type) {
+            // Child widgets need to be manually unparented in `dispose()`.
+            if let Some(child) = self.first_label.borrow_mut().take() {
+                child.unparent();
+            }
+
+            if let Some(child) = self.second_label.borrow_mut().take() {
+                child.unparent();
+            }
+        }
+
+        fn set_property(&self, obj: &Self::Type, id: usize, value: &glib::Value) {
+            let prop = &PROPERTIES[id];
+
+            match *prop {
+                glib::subclass::Property("orientation", ..) => {
+                    let orientation = value.get().unwrap().unwrap();
+                    self.orientation.replace(orientation);
+                    // We have to set the value in our layout manager as well.
+                    let layout_manager = obj
+                        .get_layout_manager()
+                        .unwrap()
+                        .downcast::<gtk::BoxLayout>()
+                        .unwrap();
+                    layout_manager.set_orientation(orientation);
+                }
+                _ => unimplemented!(),
+            }
+        }
+
+        fn get_property(&self, _obj: &Self::Type, id: usize) -> glib::Value {
+            let prop = &PROPERTIES[id];
+
+            match *prop {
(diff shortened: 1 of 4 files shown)


gtk4/src/subclass/orientable.rs
```diff
@@ -0,0 +1,41 @@
...
+            let type_ = ffi::gtk_orientable_get_type();
+            let iface = glib::gobject_ffi::g_type_default_interface_ref(type_)
+                as *mut ffi::GtkOrientableIface;
+            assert!(!iface.is_null());
+
+            let orientation = ffi::gtk_orientable_get_orientation(
+                orientable.unsafe_cast_ref::<Orientable>().to_glib_none().0,
+            );
+
+            glib::gobject_ffi::g_type_default_interface_unref(iface as glib::ffi::gpointer);
```
sdroege, line 20: There are no default implementations for any of the functions in `GtkOrientable`

sdroege: Also there are no methods at all in `GtkOrientable` :) It works by having your subclass provide a (appropriately typed) "orientation" property.

The trait should be empty

gtk4/src/subclass/orientable.rs
```diff
@@ -0,0 +1,41 @@
...
+
+            ffi::gtk_orientable_set_orientation(
+                orientable.unsafe_cast_ref::<Orientable>().to_glib_none().0,
+                orientation.to_glib(),
+            );
+
+            glib::gobject_ffi::g_type_default_interface_unref(iface as glib::ffi::gpointer);
+        }
+    }
+}
```
sdroege, line 41: You need to provide an `IsImplementable` impl for this

gtk4/src/subclass/orientable.rs
A6GibKm: Done.

gtk4/src/subclass/orientable.rs
A6GibKm: Done.

A6GibKm:
It was not clear to me how to implement this for the Button example. See https://paste.centos.org/view/22d00849

sdroege:
> It was not clear to me how to implement this for the Button example. See https://paste.centos.org/view/22d00849

You need to install an `orientation` property of type `gtk::Orientation` on your subclass.

Let's maybe also get that example once it works into the examples subdirectory.

sdroege (approved):

A6GibKm:
Added an example, which of course, does not not work hehe.

sdroege:
How does it not work, what happens?

A6GibKm:
I am sorry for the abysmal lack of precision. The two labels appear horizontal.

A6GibKm:
Fixed.

examples/src/bin/orientable_subclass.rs
```diff
@@ -0,0 +1,168 @@
...
+
+impl CustomOrientable {
+    pub fn new() -> Self {
+        glib::Object::new(&[]).expect("Failed to create CustomOrientable")
+    }
+}
+
+fn main() {
+    let application = gtk::Application::new(
+        Some("com.github.gtk-rs.examples.widget_subclass"),
```
bilelmoussaoui, line 136:
```suggestion
        Some("com.github.gtk-rs.examples.orientable_subclass"),
```

bilelmoussaoui:
Looks good to me overall, @sdroege mind having a quick look at the various comments written to see if something doesn't really make sense? :)

examples/src/bin/orientable_subclass.rs
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        fn set_property(&self, obj: &Self::Type, id: usize, value: &glib::Value) {
+            let prop = &PROPERTIES[id];
+
+            match *prop {
+                glib::subclass::Property("orientation", ..) => {
+                    let orientation = value.get().unwrap().unwrap();
+                    self.orientation.replace(orientation);
+                    // We have to set the value in our layout manager as well.
```
sdroege, line 105: Why store it in both places? You could also just use the layout manager or not?

sdroege:
> Looks good to me overall, @sdroege mind having a quick look at the various comments written to see if something doesn't really make sense? :)

Which comments? Does it work now? :) I have no idea about all the GTK specifics here, that's something for you to check but I guess it's ok if it works

examples/src/bin/orientable_subclass.rs
bilelmoussaoui: Actually that's what gtkbox does, not sure why though..

A6GibKm:
Comments in the example. The example works just fine, it is just a custom widget consisting of a pair of labels, and a button which changes the orientation of the former.

examples/src/bin/orientable_subclass.rs
```diff
@@ -0,0 +1,176 @@
...
+    #[derive(Debug)]
+    pub struct CustomOrientable {
+        first_label: RefCell<Option<gtk::Widget>>,
+        second_label: RefCell<Option<gtk::Widget>>,
+        orientation: RefCell<gtk::Orientation>,
+    }
+
+    // Every widget that implements Orientable has to define a "orientation"
+    // property like below, gtk::Orientation::Horizontal is a placeholder
+    // for the initial value.
```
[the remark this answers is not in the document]
A6GibKm: The interesting bit is that it is not the default. As you have to specify the value in `new()` anyway.

By the way, I noted two nitpicks with Orientation being automagically generated. It is not an enum consisting only of two values and it does not implement default.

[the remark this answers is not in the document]
sdroege: > The interesting bit is that it is not the default. As you have to specify the value in new() anyway.

It's for documentation purposes the default, and also if you introspect the property. If you specify `ParamFlags::CONSTRUCT` for the property, object construction will also call `set_property()` with that default value.

> By the way, I noted two nitpicks with Orientation being automagically generated. It is not an enum consisting only of two values

There's also the `__Unknown` variant as we can't know if/when the C code introduces new values and you wouldn't want to crash if that happens.

> and it does not implement default.

That can be solved but what should be the default? Both horizontal and vertical seem equally fine as a default

examples/src/bin/orientable_subclass.rs
sdroege: :man_shrugging: More state to keep in sync, I don't like it

examples/src/bin/orientable_subclass.rs
```diff
@@ -0,0 +1,176 @@
...
+    //
+    // glib::ParamFlags::CONSTRUCT allows us to set that property the moment
+    // we create a new instance of the widget
```
sdroege, line 24: That's always possible, but `CONSTRUCT` properties are set before all the others. It rarely makes a difference.

But `CONSTRUCT` means that it will even be passed to `set_property()` on construction if it was not specified as part of `glib::Object::new()`. In that case the default property value would be used.

sdroege (commented):
Looks good to me apart from the comments I just added

examples/src/bin/orientable_subclass.rs
A6GibKm: The thing is that if you set Vertical everywhere, without that flag, it would start as Horizontal anyways. The setter only sends the value to the layout manager on the first set.

sdroege: I don't understand what you mean, sorry. Can you explain in more detail what happens and what problem you observe?

A6GibKm: We are currently setting horizontal as the default value of the PROPERTY, and in the RefCell in new(). I don't know if this is a problem, but if you switch them both to vertical, the initial (visible) state will still be horizontal, my guess is that the setter (which changes the layout manager) has not kicked yet.

examples/src/bin/orientable_subclass.rs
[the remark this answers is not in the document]
A6GibKm: > That can be solved but what should be the default? Both horizontal and vertical seem equally fine as a default

In practice I have found Horizontal to be the gtk default.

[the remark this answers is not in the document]
sdroege: @bilelmoussaoui what do you think? Should such a `Default` impl be added?

examples/src/bin/orientable_subclass.rs
sdroege: @bilelmoussaoui is that how layout managers work?

From the Rust side, you should get things set correctly at least.

examples/src/bin/orientable_subclass.rs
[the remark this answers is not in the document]
bilelmoussaoui: It can be added as a default, not sure if it's worth it though :)

examples/src/bin/orientable_subclass.rs
bilelmoussaoui: Ah, that's only because GtkBox exposes the "orientation" property, so we could technically just bind both properties here. In general you would just set the proper orientation only once without having to implement orientable at all. 

So this more of a copy of the GtkBox I guess

examples/src/bin/orientable_subclass.rs
bilelmoussaoui: Layout managers are currently a bit weird to use. And the current example is trying to implement a GtkBox without really implementing one. I would say it's okay as is, and yes the property has to be set at least once, so that the layout manager's property here gets updated. That's why passing a value when we construct the widget is required in such case

sdroege: You don't have to provide it to the `new()` call if you use the `CONSTRUCT` flag for the property.

bilelmoussaoui: Ah; TIL

sdroege: So how should we proceed here? :)

sdroege: Works for me

bilelmoussaoui:
Thanks @A6GibKm

PR #149, opened 2021-01-05
Add Dialog::run_future()

### LINKED ISSUE #144: Keep gtk::DialogExt::run() as async variant
When looking at my usage of `dialog.run()` I decided to just implement an `async run()`.

The only solution I see is to connect to the "response" event and send the result through a channel. But I can also check how gtk3 implemented this.

Any thoughts?

sophie-h:
Closes #144

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -34,6 +34,9 @@ name = "css"
 name = "custom_paintable"
 
 [[bin]]
+name = "dialog"
+
+[[bin]]
 name = "entry_completion"
 
 [[bin]]
--- a/examples/src/bin/dialog.rs
+++ b/examples/src/bin/dialog.rs
@@ -1,0 +1,77 @@
+//! # Dialog Example
+//!
+//! Example of how to obtain the response to a dialog as a future
+
+use gtk::glib::clone;
+use gtk::glib::signal::Inhibit;
+use gtk::prelude::*;
+
+use std::env::args;
+use std::rc::Rc;
+
+async fn dialog<W: IsA<gtk::Window>>(window: Rc<W>) {
+    let question_dialog = gtk::MessageDialogBuilder::new()
+        .transient_for(&*window)
+        .modal(true)
+        .buttons(gtk::ButtonsType::OkCancel)
+        .text("What is your answer?")
+        .build();
+
+    let answer = question_dialog.run_future().await;
+    question_dialog.close();
+
+    let info_dialog = gtk::MessageDialogBuilder::new()
+        .transient_for(&*window)
+        .modal(true)
+        .buttons(gtk::ButtonsType::Close)
+        .text("You answered")
+        .secondary_text(&format!("Your answer: {:?}", answer))
+        .build();
+
+    info_dialog.run_future().await;
+    info_dialog.close();
+}
+
+fn build_ui(application: &gtk::Application) {
+    let button = gtk::ButtonBuilder::new()
+        .label("Open Dialog")
+        .halign(gtk::Align::Center)
+        .valign(gtk::Align::Center)
+        .build();
+
+    let window = Rc::new(
+        gtk::ApplicationWindowBuilder::new()
+            .application(application)
+            .title("Dialog Example")
+            .default_width(350)
+            .default_height(70)
+            .child(&button)
+            .visible(true)
+            .build(),
+    );
+
+    button.connect_clicked(clone!(@strong window =>
+        move |_| {
+            gtk::glib::MainContext::default().spawn_local(dialog(Rc::clone(&window)));
+        }
+    ));
+
+    window.connect_close_request(move |window| {
+        if let Some(application) = window.get_application() {
+            application.remove_window(window);
+        }
+        Inhibit(false)
+    });
+}
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let application = gtk::ApplicationBuilder::new()
+        .application_id("com.github.gtk-rs.examples.dialog")
+        .build()?;
+
+    application.connect_activate(build_ui);
+
+    application.run(&args().collect::<Vec<_>>());
+
+    Ok(())
+}
--- a/gtk4/Cargo.toml
+++ b/gtk4/Cargo.toml
@@ -36,6 +36,7 @@ git = "https://github.com/gtk-rs/lgpl-docs"
 libc = "0.2"
 bitflags = "1.0"
 field-offset = "0.3"
+futures-channel = "0.3"
 once_cell = "1.0"
 ffi =  { package = "gtk4-sys", path = "./sys" }
 gtk4-macros =  { path = "../gtk4-macros" }
--- a/gtk4/src/dialog.rs
+++ b/gtk4/src/dialog.rs
@@ -1,9 +1,12 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use crate::{Dialog, DialogExt, DialogFlags, ResponseType, Widget, Window};
+use crate::{Dialog, DialogExt, DialogFlags, ResponseType, Widget, WidgetExt, Window};
 use glib::object::Cast;
 use glib::translate::*;
-use glib::IsA;
+use glib::{IsA, ObjectExt};
+use std::cell::Cell;
+use std::future::Future;
+use std::pin::Pin;
 use std::ptr;
 
 impl Dialog {
@@ -36,9 +39,25 @@ pub trait DialogExtManual: 'static {
 
     #[doc(alias = "gtk_dialog_get_response_for_widget")]
     fn get_response_for_widget<P: IsA<Widget>>(&self, widget: &P) -> ResponseType;
+
+    // rustdoc-stripper-ignore-next
+    /// Shows the dialog and returns a `Future` that resolves to the
+    /// `ResponseType` on response.
+    ///
+    /// ```ignore
+    /// let dialog = gtk::MessageDialogBuilder::new()
+    ///    .buttons(gtk::ButtonsType::OkCancel)
+    ///    .text("What is your answer?")
+    ///    .build();
+    ///
+    /// let answer = dialog.run_future().await;
+    /// dialog.close();
+    /// println!("Answer: {:?}", answer);
+    /// ```
+    fn run_future<'a>(&'a self) -> Pin<Box<dyn Future<Output = ResponseType> + 'a>>;
 }
 
-impl<O: IsA<Dialog>> DialogExtManual for O {
+impl<O: IsA<Dialog> + IsA<Widget>> DialogExtManual for O {
     fn add_buttons(&self, buttons: &[(&str, ResponseType)]) {
         for &(text, id) in buttons {
             //FIXME: self.add_button don't work on 1.8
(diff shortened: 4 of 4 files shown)


examples/src/bin/dialog.rs
```diff
@@ -0,0 +1,55 @@
...
+
+async fn dialog<W: IsA<gtk::Window>>(window: Rc<W>) {
+    let dialog = gtk::MessageDialog::new(
+        Some(&*window),
+        gtk::DialogFlags::empty(),
+        gtk::MessageType::Question,
+        gtk::ButtonsType::OkCancel,
+        "Hello",
+    );
+    println!("{:#?}", dialog.run_future().await);
```
sophie-h, line 22: Replace with reaction in interface

examples/Cargo.toml
```diff
@@ -3,47 +3,10 @@ name = "gtk4-rs-examples"
...
 
 [dependencies]
 chrono = "0.4.19"
 
 [dependencies.gtk]
 path = "../gtk4"
 package = "gtk4"
-
-[[bin]]
-name = "basic"
```
sdroege, line 16: Can you put this into a separate commit?

gtk4/Cargo.toml
```diff
@@ -37,9 +37,11 @@ optional = true
 git = "https://github.com/gtk-rs/lgpl-docs"
 
 [dependencies]
+async-trait = "0.1"
```
sdroege, line 40: Let's do this manually for now instead of adding another dependency. It's easy enough

gtk4/src/dialog.rs
```diff
@@ -30,15 +33,19 @@ impl Dialog {
...
 
+#[async_trait(?Send)]
 pub trait DialogExtManual: 'static {
     #[doc(alias = "gtk_dialog_add_buttons")]
     fn add_buttons(&self, buttons: &[(&str, ResponseType)]);
 
     #[doc(alias = "gtk_dialog_get_response_for_widget")]
     fn get_response_for_widget<P: IsA<Widget>>(&self, widget: &P) -> ResponseType;
+
+    async fn run_future(&self) -> ResponseType;
```
sdroege, line 44:
```suggestion
    fn run_future(&self) -> impl Future<Output = ResponseType>;
```

gtk4/src/dialog.rs
```diff
@@ -49,9 +56,31 @@ impl<O: IsA<Dialog>> DialogExtManual for O {
...
+        let sender = std::cell::Cell::new(Some(sender));
+
+        let response_handler = self.connect_response(move |_, response_type| {
+            if let Some(m) = sender.replace(None) {
+                let _result = m.send(response_type);
+            }
+        });
+
+        let modal = self.get_modal();
+        self.set_modal(true);
```
sdroege, line 77: Why set it to modal?

gtk4/src/dialog.rs
```diff
@@ -49,9 +56,31 @@ impl<O: IsA<Dialog>> DialogExtManual for O {
...
         unsafe {
             from_glib(ffi::gtk_dialog_get_response_for_widget(
-                self.as_ref().to_glib_none().0,
+                AsRef::<Dialog>::as_ref(self).to_glib_none().0,
                 widget.as_ref().to_glib_none().0,
             ))
         }
     }
+
+    async fn run_future(&self) -> ResponseType {
```
sdroege, line 65:
```suggestion
    fn run_future(&self) -> Pin<Box<Future<Output = ResponseType>>> {
```

sdroege: And the body would basically be

```
async {
    // What you have now
}
```

gtk4/Cargo.toml
sophie-h: Done

gtk4/src/dialog.rs
```diff
@@ -49,9 +53,30 @@ impl<O: IsA<Dialog>> DialogExtManual for O {
...
+
+            let sender = std::cell::Cell::new(Some(sender));
+
+            let response_handler = self.connect_response(move |_, response_type| {
+                if let Some(m) = sender.replace(None) {
+                    let _result = m.send(response_type);
+                }
+            });
+
+            self.show();
```
sophie-h, line 74: Do we want to keep this? It doesn't seem too unreasonable.

sdroege: I'd make it behave like `gtk_dialog_run()` in GTK3. Least amount of surprises :) I don't know what exactly it did back then

gtk4/src/dialog.rs
```diff
@@ -49,9 +53,30 @@ impl<O: IsA<Dialog>> DialogExtManual for O {
...
+
+            let result = receiver.await.unwrap_or(ResponseType::None);
+            self.disconnect(response_handler);
```
sdroege, line 77: If `receiver.await` returns an `Err`, then the signal handler was freed before the dialog got a response. In that case, `disconnect()` will probably give a critical warning so you probably only want to call it in the other case

sdroege (commented):
Looks good to me otherwise, seems like a good usability improvement to me :)

sdroege:
@bilelmoussaoui (doesn't show you in the reviewers list for some reason) Also seems like a good idea to you?

GuillaumeGomez (approved):

gtk4/src/dialog.rs
sophie-h: Fixed.

bilelmoussaoui:
> @bilelmoussaoui (doesn't show you in the reviewers list for some reason) Also seems like a good idea to you?

It feels a bit odd to me to start providing async variants on top of the basic bindings like this :)

sdroege:
> It feels a bit odd to me to start providing async variants on top of the basic bindings like this :)

We also have those in glib/gio already though.

bilelmoussaoui:
> > It feels a bit odd to me to start providing async variants on top of the basic bindings like this :)
> 
> We also have those in glib/gio already

sophie-h:
> It feels a bit odd to me to start providing async variants on top of the basic bindings like this :)

Sounds like an argument that would also apply to all the *Builder traits that are also rusty variants on basic bindings. Also, I think that `run()` would still be part of GTK4 if it wasn't for the blocking behavior.

bilelmoussaoui:
> > It feels a bit odd to me to start providing async variants on top of the basic bindings like this :)
> 
> Sounds like an argument that would also apply to all the *Builder traits that are also rusty variants on basic bindings. Also, I think that `run()` would still be part of GTK4 if it wasn't for the blocking behavior.

Not really, they are just a way of calling `glib::Object::new(&[])` =) I'm not really against such change, I was just wondering if/where we should draw a line between providing just bare-bone bindings or more than that.

sdroege:
IMHO this is still fine, especially as it's not a lot of code and improves usability considerably. And it's not very different from the various futures API we provide in glib and gio already.

A function like this for gtk3 would also make sense.

@GuillaumeGomez What do you think?

GuillaumeGomez:
My issue with this is a bit different: could we make this change automatic through gir? Could we detect the blocking `run` functions so we don't have generate them by hand?

bilelmoussaoui:
> My issue with this is a bit different: could we make this change automatic through gir? Could we detect the blocking `run` functions so we don't have generate them by hand?

Gtk4 doesn't have any run functions anymore

sdroege:
> Gtk4 doesn't have any run functions anymore

And for gtk3 we should really get rid of them. Re-entrant mainloops are a disaster :)

GuillaumeGomez:
> And for gtk3 we should really get rid of them. Re-entrant mainloops are a disaster :)

Do we have an issue for this?

> Gtk4 doesn't have any run functions anymore

Then how can we add a `run_future`? I'm lost here...

sdroege:
> Do we have an issue for this?

I don't think so but I changed all examples a long time ago.

> Then how can we add a `run_future`? I'm lost here...

By... implementing something that behaves like `run()` before :P Just look at the code

examples/src/bin/dialog.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    let info_dialog = gtk::MessageDialogBuilder::new()
+        .transient_for(&*window)
+        .modal(true)
+        .buttons(gtk::ButtonsType::Close)
+        .text("You answered")
+        .secondary_text(&format!("Your answer: {:?}", answer))
+        .build();
+
+    info_dialog.run_future().await;
```
sdroege, line 31: Might make sense to do something with the return value here for example purposes

sdroege (approved):

sophie-h:
> And for gtk3 we should really get rid of them. Re-entrant mainloops are a disaster :)

We could deprecate `run()` in gtk3-rs and add the future version (only requires extra handling of the delete event.)

examples/src/bin/dialog.rs
sophie-h: The answer from the previous dialog is shown. I think this is a good example why you might wan't to use async/await even if it's only for closing the dialog afterwards.

sdroege: ok :)

sdroege:
Now just need to fix the CI. Something needs to be done to allow `await` inside documentation. @GuillaumeGomez ?

GuillaumeGomez:
After thinking about this for a while, I think it'd be nice to add it. **However**: please add the non-future equivalent (with a callback).

GuillaumeGomez:
> Something needs to be done to allow `await` inside documentation. @GuillaumeGomez ?

Just wrap it inside an async function?

sdroege:
> After thinking about this for a while, I think it'd be nice to add it. **However**: please add the non-future equivalent (with a callback).

Why? That already exists: connect to the `response` signal.

GuillaumeGomez:
Then I can answer the same for `run_future`. :)

sdroege:
> Then I can answer the same for `run_future`. :)

Signals can't generally be used as futures, that's why such a helper API here is useful for this specific signal.

GuillaumeGomez:
Maybe, but it can be done as proved by this PR. So the same can be done for `run`. It wouldn't make sense to only have future version of `run`.

sdroege:
> Maybe, but it can be done as proved by this PR. So the same can be done for `run`. It wouldn't make sense to only have future version of `run`.

Can you implement `run()` with a closure then? I don't know how you imagine that to look like

GuillaumeGomez:
Sure.

gtk4/src/dialog.rs
```diff
@@ -36,9 +38,24 @@ pub trait DialogExtManual: 'static {
 
     #[doc(alias = "gtk_dialog_get_response_for_widget")]
     fn get_response_for_widget<P: IsA<Widget>>(&self, widget: &P) -> ResponseType;
+
+    /// Shows the dialog and returns a `Future` that resolves to the
+    /// `ResponseType` on response.
+    ///
+    /// ```
```
GuillaumeGomez, line 45: Please add `no_run`.

gtk4/src/dialog.rs
```diff
@@ -36,9 +38,24 @@ pub trait DialogExtManual: 'static {
 
     #[doc(alias = "gtk_dialog_get_response_for_widget")]
     fn get_response_for_widget<P: IsA<Widget>>(&self, widget: &P) -> ResponseType;
+
+    /// Shows the dialog and returns a `Future` that resolves to the
```
GuillaumeGomez, line 42: Please add before this doc comment:

```rust
```

gtk4/src/dialog.rs
sophie-h: Thanks! Fixed.

gtk4/src/dialog.rs
sophie-h: Thanks! Fixed.

sophie-h:
All green

PR #152, opened 2021-01-07
examples: add a gif paintable

bilelmoussaoui:
adds a sample paintable that uses the image crate to convert it to frames
and then render them

It doesn't work as expected currently

### DIFF
--- a/examples/src/bin/gif_paintable.rs
+++ b/examples/src/bin/gif_paintable.rs
@@ -1,0 +1,305 @@
+// A custom GDK paintable capable of rendering a GIF
+// The paintable makes uses of the awesome image
+// crate to read a gif file and transform it to a Vec<Frame>
+// Which are then rendered by the paintable at snapshot
+// A tick callback that calls invalidate_contents on the paintable
+// to render a new frame
+use glib::clone;
+use gtk::prelude::*;
+use gtk::subclass::prelude::*;
+use gtk::{gdk, gio, glib, graphene};
+use image::{gif::GifDecoder, AnimationDecoder};
+use std::env::args;
+use std::io::Cursor;
+
+pub struct Frame {
+    pub width: i32,
+    pub height: i32,
+    pub texture: gdk::Paintable,
+}
+
+impl Frame {
+    pub fn new(f: image::Frame) -> Self {
+        let image = f.into_buffer();
+
+        let samples = image.into_flat_samples();
+        let (stride, width, height) = samples.extents();
+        let width = width as i32;
+        let height = height as i32;
+
+        let bytes = glib::Bytes::from(samples.as_slice());
+
+        let texture =
+            gdk::MemoryTexture::new(width, height, gdk::MemoryFormat::R8g8b8a8, &bytes, stride);
+        Frame {
+            width,
+            height,
+            texture: texture.upcast(),
+        }
+    }
+}
+
+mod imp {
+    use super::*;
+    use glib::subclass;
+    use std::cell::{Cell, RefCell};
+
+    pub struct GifPaintable {
+        pub frames: RefCell<Vec<Frame>>,
+        pub next_frame: RefCell<Option<gdk::Paintable>>,
+        pub width: Cell<i32>,
+        pub height: Cell<i32>,
+        pub current_idx: Cell<usize>,
+    }
+
+    impl ObjectSubclass for GifPaintable {
+        const NAME: &'static str = "GifPaintable";
+        type Type = super::GifPaintable;
+        type ParentType = glib::Object;
+        type Instance = subclass::simple::InstanceStruct<Self>;
+        type Class = subclass::simple::ClassStruct<Self>;
+
+        glib::object_subclass!();
+
+        fn type_init(type_: &mut subclass::InitializingType<Self>) {
+            type_.add_interface::<gdk::Paintable>();
+        }
+
+        fn new() -> Self {
+            Self {
+                next_frame: RefCell::new(None),
+                frames: RefCell::new(vec![]),
+                width: Cell::new(0),
+                height: Cell::new(0),
+                current_idx: Cell::new(0),
+            }
+        }
+    }
+
+    impl ObjectImpl for GifPaintable {
+        fn constructed(&self, obj: &Self::Type) {
+            self.parent_constructed(obj);
+        }
+
+        fn dispose(&self, _obj: &Self::Type) {}
+    }
+
+    impl PaintableImpl for GifPaintable {
+        fn get_intrinsic_height(&self, _paintable: &Self::Type) -> i32 {
+            self.height.get()
+        }
+
+        fn get_intrinsic_width(&self, _paintable: &Self::Type) -> i32 {
+            self.width.get()
+        }
+
+        fn snapshot(
+            &self,
+            _paintable: &Self::Type,
+            snapshot: &gdk::Snapshot,
+            width: f64,
+            height: f64,
+        ) {
+            if let Some(texture) = self.next_frame.borrow_mut().take() {
+                let w = self.width.get() as f64;
+                let h = self.height.get() as f64;
+                texture.snapshot(snapshot, w, h);
+            } else {
+                let snapshot = snapshot.downcast_ref::<gtk::Snapshot>().unwrap();
+                snapshot.append_color(
+                    &gdk::RGBA::black(),
+                    &graphene::Rect::new(0f32, 0f32, width as f32, height as f32),
+                );
+            }
+        }
+    }
+}
+
+glib::wrapper! {
+    pub struct GifPaintable(ObjectSubclass<imp::GifPaintable>) @implements gdk::Paintable;
+}
+
+impl GifPaintable {
+    pub fn new() -> Self {
+        glib::Object::new(&[]).expect("Failed to create object of type GifPaintable")
+    }
+
+    // Loads the bytes of a GIF into the paintable
+    // The loading consists of decoding the gif with a GIFDecoder
+    // Then storing the frames so that the paintable can render them
+    pub fn load_form_bytes(&self, bytes: Vec<u8>) -> Result<(), Box<dyn std::error::Error>> {
+        let self_ = imp::GifPaintable::from_instance(self);
+        self_.current_idx.set(0);
+
+        let read = Cursor::new(bytes);
+        let decoder = GifDecoder::new(read)?;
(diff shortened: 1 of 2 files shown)


examples/src/bin/gif_paintable.rs
```diff
@@ -0,0 +1,305 @@
+// A custom GDK paintable capable of rendering a GIF
+// The paintable makes uses of the awesome image
+// crate to read a gif file and transform it to a Vec<Frame>
```
sdroege, line 3: That sounds like a bad idea, especially if the GIF has loops. You can easily use up all memory like this :)

It would be better to decode on demand.

examples/src/bin/gif_paintable.rs
```diff
@@ -0,0 +1,305 @@
...
+            type_.add_interface::<gdk::Paintable>();
+        }
+
+        fn new() -> Self {
+            Self {
+                next_frame: RefCell::new(None),
+                frames: RefCell::new(vec![]),
+                width: Cell::new(0),
+                height: Cell::new(0),
+                current_idx: Cell::new(0),
```
sdroege, line 74: Maybe use `Option`s here

examples/src/bin/gif_paintable.rs
bilelmoussaoui: > That sounds like a bad idea, especially if the GIF has loops. You can easily use up all memory like this :)
> 
> It would be better to decode on demand.

Ah that sounds better, but for now I have to figure out first why my texture isn't exactly what I'm supposed to get

examples/src/bin/gif_paintable.rs
```diff
@@ -0,0 +1,305 @@
...
+            }
+        }
+    }
+
+    impl ObjectImpl for GifPaintable {
+        fn constructed(&self, obj: &Self::Type) {
+            self.parent_constructed(obj);
+        }
+
+        fn dispose(&self, _obj: &Self::Type) {}
```
sdroege, line 84: These two functions are empty

examples/src/bin/gif_paintable.rs
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+        fn snapshot(
+            &self,
+            _paintable: &Self::Type,
+            snapshot: &gdk::Snapshot,
+            width: f64,
+            height: f64,
+        ) {
+            if let Some(texture) = self.next_frame.borrow_mut().take() {
```
sdroege, line 103: Should it not continue returning the same frame until it's time for the following one?

examples/src/bin/gif_paintable.rs
```diff
@@ -0,0 +1,305 @@
...
+
+impl GifPaintable {
+    pub fn new() -> Self {
+        glib::Object::new(&[]).expect("Failed to create object of type GifPaintable")
+    }
+
+    // Loads the bytes of a GIF into the paintable
+    // The loading consists of decoding the gif with a GIFDecoder
+    // Then storing the frames so that the paintable can render them
+    pub fn load_form_bytes(&self, bytes: Vec<u8>) -> Result<(), Box<dyn std::error::Error>> {
```
sdroege, line 130:
```suggestion
    pub fn load_from_bytes(&self, bytes: Vec<u8>) -> Result<(), Box<dyn std::error::Error>> {
```

examples/src/bin/gif_paintable.rs
bilelmoussaoui: Indeed,

examples/src/bin/gif_paintable.rs
```diff
@@ -0,0 +1,305 @@
...
+            .into_frames()
+            .collect_frames()?
+            .into_iter()
+            .map(Frame::new)
+            .collect::<Vec<Frame>>();
+
+        self_.frames.replace(frames);
+        self.invalidate_contents();
+        glib::timeout_add_local(
+            std::time::Duration::from_millis(60),
```
sdroege, line 146: GIFs usually continue information about the framerate or rather when to schedule frames. Every 60ms is not going to be correct in general.

bilelmoussaoui: Yes, that part is going to be cleaned up a bit, especially that we should stop the timeout at dispose

examples/src/bin/gif_paintable.rs
```diff
@@ -0,0 +1,305 @@
...
+            .transient_for(self)
+            .action(gtk::FileChooserAction::Open)
+            .build();
+
+        file_chooser.add_filter(&gif_filter);
+        file_chooser.connect_response(clone!(@weak self as win => move |dialog, response| {
+            if response == gtk::ResponseType::Accept {
+                let selected_file = dialog.get_file().unwrap();
+                win.open_file(selected_file).unwrap();
+            }
```
sdroege, line 250: Shouldn't this set `self.file_chooser = None`?

examples/src/bin/gif_paintable.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    fn open_file(&self, file: gio::File) -> Result<(), Box<dyn std::error::Error>> {
+        let self_ = win::Window::from_instance(self);
+        let (bytes, _) = file.load_contents(gio::NONE_CANCELLABLE)?;
+        self_.paintable.load_form_bytes(bytes)?;
+        Ok(())
+    }
+
+    fn open_file_dialog(&self) {
+        let self_ = win::Window::from_instance(self);
```
sdroege, line 231: Maybe don't allow creating a new file chooser if one is open already

sdroege (commented):
Otherwise seems like a nice example :)

examples/src/bin/gif_paintable.rs
```diff
@@ -0,0 +1,305 @@
...
+            .accept_label("Open")
+            .cancel_label("Cancel")
+            .modal(true)
+            .title("Open a GIF to display")
+            .transient_for(self)
+            .action(gtk::FileChooserAction::Open)
+            .build();
+
+        file_chooser.add_filter(&gif_filter);
+        file_chooser.connect_response(clone!(@weak self as win => move |dialog, response| {
```
sdroege, line 246: With something like https://github.com/gtk-rs/gtk4-rs/pull/149 you could make this code a bit nicer :)

examples/src/bin/gif_paintable.rs
bilelmoussaoui: The file chooser is modal, so the user can't focus the parent window

sdroege: Worth an assertion here then

sp1ritCS:
Any updates on this one?

I attempted to fix the compilation errors this (not the gui, just the Frame & GifPaintable part) had. I managed to do that, however it not only takes the GifDecoder about thirty seconds to decode a 13.4MB gif, but when I try to add the paintable to an GtkImage it panicks: `assertion failed: self.is::<T>()', /home/admin/.cargo/git/checkouts/gtk-rs-core-7be42ca38bd6361c/41c1e5c/glib/src/object.rs:260:9`

backtrace:
[backtrace.txt.gz](https://github.com/gtk-rs/gtk4-rs/files/6532570/backtrace.txt.gz)

bilelmoussaoui:
> Any updates on this one?
> 
> I attempted to fix the compilation errors this (not the gui, just the Frame & GifPaintable part) had. I managed to do that, however it not only takes the GifDecoder about thirty seconds to decode a 13.4MB gif, but when I try to add the paintable to an GtkImage it panicks: `assertion failed: self.is::<T>()', /home/admin/.cargo/git/checkouts/gtk-rs-core-7be42ca38bd6361c/41c1e5c/glib/src/object.rs:260:9`
> 
> backtrace:
> [backtrace.txt.gz](https://github.com/gtk-rs/gtk4-rs/files/6532570/backtrace.txt.gz)

Do you mind sharing your code somewhere? there's another paintable example you can look at

sp1ritCS:
> Do you mind sharing your code somewhere? there's another paintable example you can look at

sure,  
this is the "fixed" version of the GifPaintable: [paintable.rs.gz](https://github.com/gtk-rs/gtk4-rs/files/6534282/paintable.rs.gz)

and I'm just using it like this:
```rust
let cpaintable = paintable::GifPaintable::new();
let cover = gtk::gio::File::for_path("/home/admin/workspace/Projects/gtk4test/data/dbgpic/floppybuddy.gif");
let (bytes, _) = cover.load_contents(gtk::gio::NONE_CANCELLABLE).expect("Failed reading cover");
cpaintable.load_from_bytes(bytes).unwrap();
---
let lib_item = ImageBuilder::new()
		//.file("/home/admin/workspace/Projects/gtk4test/data/dbgpic/doom.png")
		.paintable(&cpaintable)
		.height_request(192)
		.width_request(128)
		.build();
```

bilelmoussaoui:
the code you shared is super old and won't compile against current gtk4-rs (form the master branch), please look at the other paintable example on how to update it to something working. I don't have the time to look into this one in details right now

sp1ritCS:
> the code you shared is super old and won't compile against current gtk4-rs

well, I tried it fix it up, and it worked with whatever version of gtk4-rs I have installed (fairly recent)

nonetheless I've looked at the custom paintable example and came up with something that doesn't crash. However it's still unable to show any frames (except if I click/select it with the gtk inspector tool on it, then it shows a black frame (i.e texture is empty at this moment)

I've tried to find the bug myself (even going so far to "dump" a frame and [rendering it inside a html canvas](https://github.com/gtk-rs/gtk4-rs/files/6538144/rlog.html.gz)), but I'm unable to find the issue 🤷🏻‍♂️.

See: [paintable.tar.gz](https://github.com/gtk-rs/gtk4-rs/files/6538156/paintable.tar.gz)
```rust
let cpaintable = paintable::CoverPaintable::new();
let cover = std::fs::File::open("/home/admin/workspace/Projects/gtk4test/data/dbgpic/floppybuddy.gif").expect("Failed opening cover");
let lib_item = ImageBuilder::new()
		.paintable(&cpaintable)
		.height_request(192)
		.width_request(128)
		.build();
```

also image-rs seems to be very slow when decoding "larger" images :/, any alternative?

bilelmoussaoui:
Have you tried to use it with a GtkPicture instead of a GtkImage? I think GtkImage is "constant"

sp1ritCS:
> Have you tried to use it with a GtkPicture instead of a GtkImage? I think GtkImage is "constant"

nope, same thing 😐

PR #157, opened 2021-01-10
more manual stuff

bilelmoussaoui:
Adds
- `gdk::ToplevelSize`
- `gsk::ParseLocation`


gdk4/Gir.toml
```diff
@@ -71,6 +71,10 @@ generate = [
     "Gdk.VulkanError",
 ]
 
+ignore = [
+    "Gdk.DrawingContext", # not useful
```
sdroege, line 75: Not? Why not? :)

gdk4/src/toplevel_size.rs
```diff
@@ -0,0 +1,111 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::translate::*;
+
+#[derive(Clone, Copy, Debug)]
+#[repr(C)]
```
sdroege, line 6:
```suggestion
#[repr(transparent)]
```

Probably?

gdk4/src/toplevel_size.rs
```diff
@@ -0,0 +1,111 @@
...
+pub struct ToplevelSize(ffi::GdkToplevelSize);
+
+impl ToplevelSize {
```
sdroege, line 9: Should maybe also get a constructor and accessors for the other fields

gdk4/src/toplevel_size.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+#[doc(hidden)]
+impl FromGlibPtrBorrow<*mut ffi::GdkToplevelSize> for ToplevelSize {
+    unsafe fn from_glib_borrow(ptr: *mut ffi::GdkToplevelSize) -> glib::translate::Borrowed<Self> {
+        glib::translate::Borrowed::new(*(ptr as *mut ToplevelSize))
+    }
+}
+
+impl glib::StaticType for ToplevelSize {
```
sdroege, line 105: No `glib::Value` traits?

gsk4/src/parse_location.rs
```diff
@@ -0,0 +1,38 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::translate::*;
+
+#[derive(Copy, Clone)]
+#[repr(C)]
```
sdroege, line 6:
```suggestion
#[repr(transparent)]
```

gtk4/src/requested_size.rs
```diff
@@ -0,0 +1,28 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::translate::*;
+
+#[derive(Debug, Copy, Clone)]
+#[repr(C)]
```
sdroege, line 6:
```suggestion
#[repr(transparent)]
```

sdroege (changes requested):
Otherwise seems fine

gdk4/Gir.toml
bilelmoussaoui: I couldn't find anything that uses it and it's only defined in sys/lib.rs as 
```
#[repr(C)]
pub struct _GdkDrawingContext(c_void);

pub type GdkDrawingContext = *mut _GdkDrawingContext;
```

sdroege: Indeed seems completely unused. Strange. Please create an upstream issue then :)

bilelmoussaoui: Especially that there's a `GdkDrawContext`

bilelmoussaoui:
This should be ready for a second round of review :)

gdk4/src/toplevel.rs
```diff
@@ -1,18 +1,51 @@
...
 }
 
 impl<O: IsA<Toplevel>> ToplevelExtManual for O {
+    fn connect_compute_size<F: Fn(&Toplevel, &ToplevelSize) + 'static>(
+        &self,
+        f: F,
+    ) -> SignalHandlerId {
+        unsafe extern "C" fn compute_size_trampoline<F: Fn(&Toplevel, &ToplevelSize) + 'static>(
+            this: *mut ffi::GdkToplevel,
+            size: *mut ffi::GdkToplevelSize,
```
bilelmoussaoui, line 30: Not sure how If that's the right to handle size here as it's annotated as `out caller-allocates` does that means the signal should be returning a ToplevelSize ?

sdroege: I don't know what that means in the context of signals.

But is it intended that the signal handler fills in the `size`?

gdk4/src/toplevel_size.rs
sdroege: At least accessors for the fields would be useful then, no?

gdk4/src/toplevel.rs
bilelmoussaoui: Yes, as the C code uses it like this
```c
void
gdk_toplevel_notify_compute_size (GdkToplevel     *toplevel,
                                  GdkToplevelSize *size)
{
  g_signal_emit (toplevel, signals[COMPUTE_SIZE], 0, size);
  gdk_toplevel_size_validate (size);
}
```

sdroege: So you need getters and setters for the values in `ToplevelSize`, and make sure that the signal handlers gets a mutable reference.

The `out caller-allocates` seems strange here but I guess makes sense: the value is allocated by the caller to `g_signal_emit()` and it expects the signal to fill in things.

Can you check if all callers initialize the `ToplevelSize` with something meaningful (like zeroes) or if they just contain random uninitialized memory?

bilelmoussaoui: Can you check if all callers initialize the ToplevelSize with something meaningful (like zeroes) or if they just contain random uninitialized memory?
> it's an initialized ToplevelSize with `gdk_toplevel_size_init(&size, bound_width, bound_height);`

> So you need getters and setters for the values in ToplevelSize, and make sure that the signal handlers gets a mutable reference.
I will have a look at the various gdk backends and see which "getters"/"setters" they use and add them

sdroege: Ok, let me know. I'll do another review after that is solved

gdk4/Gir.toml
bilelmoussaoui: Fixed in https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/3232

gdk4/src/toplevel_size.rs
bilelmoussaoui: I asked about this in iirc

> ebassi: It's an opaque structure
> ebassi: You're supposed to set it up using the given methods
> ebassi: And that's all

bilelmoussaoui:
This should be ready for another round of review :-)

gdk4/src/toplevel_size.rs
```diff
@@ -0,0 +1,144 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::translate::*;
+
+#[derive(Clone, Copy, Debug)]
```
sdroege, line 5: This should probably not be `Copy` and `Clone` if it's supposed to be opaque and just passed around as a pointer

gdk4/src/toplevel_size.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    #[inline]
+    fn to_glib_none_mut(&'a mut self) -> StashMut<'a, *mut ffi::GdkToplevelSize, Self> {
+        let ptr: *mut ToplevelSize = &mut *self;
+        StashMut(ptr as *mut ffi::GdkToplevelSize, self)
+    }
+}
+
+#[doc(hidden)]
+impl FromGlibPtrNone<*const ffi::GdkToplevelSize> for ToplevelSize {
```
sdroege, line 76: Also the `From` traits seem problematic in that regard then. You should not be able to create one

gdk4/src/toplevel.rs
```diff
@@ -1,18 +1,53 @@
...
+    ) -> SignalHandlerId {
+        unsafe extern "C" fn compute_size_trampoline<
+            F: Fn(&Toplevel, &mut ToplevelSize) + 'static,
+        >(
+            this: *mut ffi::GdkToplevel,
+            size: *mut ffi::GdkToplevelSize,
+            f: glib::ffi::gpointer,
+        ) {
+            let f: &F = &*(f as *const F);
+            f(&from_glib_borrow(this), &mut from_glib_none(size))
```
sdroege, line 36: This won't work like this. You're creating a new `ToplevelSize` here and that is then modified by the callback.

```suggestion
            f(&from_glib_borrow(this), &mut *(size as *mut TopLevelSize))
```

This should work and you don't need any of `From` translation traits then

gdk4/src/toplevel_size.rs
```diff
@@ -0,0 +1,144 @@
...
+
+impl glib::StaticType for ToplevelSize {
+    #[doc(alias = "gdk_toplevel_size_get_type")]
+    fn static_type() -> glib::types::Type {
+        skip_assert_initialized!();
+        unsafe { from_glib(ffi::gdk_toplevel_size_get_type()) }
+    }
+}
+
+impl<'a> glib::value::FromValueOptional<'a> for ToplevelSize {
```
sdroege, line 121: Does it make sense to have `StaticType` and all the `GValue` machinery? I don't think so

gdk4/src/toplevel_size.rs
```diff
@@ -0,0 +1,144 @@
...
+    unsafe fn from_value_optional(value: &'a glib::Value) -> Option<Self> {
+        from_glib_full(glib::gobject_ffi::g_value_dup_boxed(value.to_glib_none().0)
```
sdroege, line 123: The boxed GValue functions are wrong for this or not, should be `g_value_XXX_pointer()` instead I think? It's a pointer type and not a boxed type

gsk4/src/auto/render_node.rs
```diff
@@ -17,10 +18,50 @@ glib::wrapper! {
...
 
 impl RenderNode {
-    //#[doc(alias = "gsk_render_node_deserialize")]
-    //pub fn deserialize(bytes: &glib::Bytes, error_func: /*Unimplemented*/FnMut(/*Ignored*/ParseLocation, /*Ignored*/ParseLocation, &glib::Error), user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> Option<RenderNode> {
-    //    unsafe { TODO: call ffi:gsk_render_node_deserialize() }
-    //}
+    #[doc(alias = "gsk_render_node_deserialize")]
+    pub fn deserialize(
+        bytes: &glib::Bytes,
+        error_func: Option<&mut dyn (FnMut(&ParseLocation, &ParseLocation, &glib::Error))>,
```
sdroege, line 24: Should maybe be two functions. `deserialize` and `deserialize_with_error_func`

gtk4/src/requested_size.rs
```diff
@@ -0,0 +1,28 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::translate::*;
+
+#[derive(Debug, Copy, Clone)]
+#[repr(transparent)]
+pub struct RequestedSize(ffi::GtkRequestedSize);
```
sdroege, line 7: As you have to pass in one into `distribute_natural_allocation()`, this should probably get a constructor too and maybe setters? How else would you call that function?

gtk4/src/requested_size.rs
```diff
@@ -0,0 +1,28 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::translate::*;
+
+#[derive(Debug, Copy, Clone)]
+#[repr(transparent)]
```
sdroege, line 6: Doesn't have to be transparent unless you want to cast between the pointers of these two types

gdk4/src/toplevel_size.rs
bilelmoussaoui: I ended up removing all the GValue stuff, I don't think it's needed neither

gtk4/src/requested_size.rs
bilelmoussaoui: I added the setters, no idea how about a constructor for now as the RequestedSize contains a client pointer, and couldn't find what it should hold yet.

bilelmoussaoui: Per mclassen, the pointer is to be used "however you see fits", I ended up passing a null pointer there for now. As alot of the implementations i found in gtk4's tree doesn't make use of it except for CellArea* stuff like https://github.com/GNOME/gtk/blob/4df7cc948c1a4ef9e3e6bde35f56b6f3b4128bd5/gtk/gtkcellareaboxcontext.c#L722 or https://github.com/GNOME/gtk/blob/949f22b8b2778c1f3e83b1e47d904fcad6a5b913/gtk/gtkcellareabox.c#L730-L743

I wonder if there's a way to pass "Any" here? or it doesn't make sense?

sdroege: Who is supposed to provide that pointer and who is supposed to use it, and who is supposed to call `distribute_natural_allocation()` on what kind of widget? :)

It seems like a way to pass arbitrary data around so we'll need to understand how it's supposed to be used or otherwise will run into crashes.

gtk4/src/auto/functions.rs
```diff
@@ -148,10 +149,21 @@ pub fn disable_setlocale() {
...
 
-//#[doc(alias = "gtk_distribute_natural_allocation")]
-//pub fn distribute_natural_allocation(extra_space: i32, n_requested_sizes: u32, sizes: /*Ignored*/&mut RequestedSize) -> i32 {
-//    unsafe { TODO: call ffi:gtk_distribute_natural_allocation() }
-//}
+#[doc(alias = "gtk_distribute_natural_allocation")]
+pub fn distribute_natural_allocation(
+    extra_space: i32,
+    n_requested_sizes: u32,
+    sizes: &mut RequestedSize,
```
sdroege, line 156: This is wrong. It's supposed to be `fn distribute_natural_allocation(extra_space: i32, sizes: &[RequestSize]) -> i32`. And unclear if the sizes should be mutable or not, are they going to be changed in there?

Also the C function is missing the array annotations :)

gsk4/src/auto/render_node.rs
```diff
@@ -17,10 +18,37 @@ glib::wrapper! {
 }
 
 impl RenderNode {
-    //#[doc(alias = "gsk_render_node_deserialize")]
-    //pub fn deserialize(bytes: &glib::Bytes, error_func: /*Unimplemented*/FnMut(/*Ignored*/ParseLocation, /*Ignored*/ParseLocation, &glib::Error), user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> Option<RenderNode> {
-    //    unsafe { TODO: call ffi:gsk_render_node_deserialize() }
-    //}
+    #[doc(alias = "gsk_render_node_deserialize")]
+    pub fn deserialize_with_error_func<P: FnMut(&ParseLocation, &ParseLocation, &glib::Error)>(
```
sdroege, line 22: And another variant without error func, or otherwise the naming would be strange

sdroege: Ah that's in another file

gtk4/src/requested_size.rs
```diff
@@ -0,0 +1,43 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::translate::*;
+
+#[derive(Debug)]
+pub struct RequestedSize(ffi::GtkRequestedSize);
+
+impl RequestedSize {
+    pub fn new() -> Self {
```
sdroege, line 9: You'll get a clippy warning that this should implement `Default`

gtk4/src/auto/functions.rs
bilelmoussaoui: I fixed the annotations upstream and updated the function here. Sorry, I should have been a bit more careful here :)

bilelmoussaoui: > And unclear if the sizes should be mutable or not, are they going to be changed in there?

Yes the minimum size is computed by this function

gtk4/src/requested_size.rs
bilelmoussaoui: I dropped the commit adding RequestedSize out of this PR, as I just noticed all the shaders stuff are not usable yet :-) I will get back to the RequestedSize later in the future by re-implementing a container as an example and figuring out how it should be used and whether it can be safe to bind or not.

Issue #27, opened 2020-10-24
Keycode/keyval in EventControllerKey are plain integers


sdroege:
Should probably be enums. See also https://github.com/gtk-rs/gdk4/issues/20

Similarly for the mouse buttons in the other event controllers.

bilelmoussaoui:
Fixed by #158

PR #158, opened 2021-01-11
gdk: properly export all consts

### LINKED ISSUE #49: Keycode/keyval constants are only in the -sys crate
See `gdk::keys` module for how it could look like.

Also the `gtk4::EventControllerKey` currently provides plain `u32` for keyval/keycode in its events. That should probably be an enum too.

bilelmoussaoui:
Fixes #49 and #27

Not sure about how to handle the keycode part


gdk4/src/key_event.rs
```diff
@@ -0,0 +1,43 @@
...
+        unsafe {
+            let mut keyval = mem::MaybeUninit::uninit();
+            let mut modifiers = mem::MaybeUninit::uninit();
+            let ret = from_glib(ffi::gdk_key_event_get_match(
+                self.to_glib_none().0,
+                keyval.as_mut_ptr(),
+                modifiers.as_mut_ptr(),
+            ));
+            let keyval: Key = keyval.assume_init().into();
+            let modifiers = modifiers.assume_init();
```
sdroege, line 24: It's UB to call `assume_init()` on uninitialized data, which would be the case if `ret == false`. You first need to check `ret`.

sdroege: Same problem in other places FWIW

gdk4/src/lib.rs
```diff
@@ -65,3 +68,39 @@ pub use rectangle::Rectangle;
...
 pub use surface::SurfaceExtManual;
 pub use time_coord::TimeCoord;
+
+// This is the priority that the idle handler processing surface updates is given in the glib::MainLoop.
+#[doc(alias = "GDK_PRIORITY_REDRAW")]
+pub const PRIORITY_REDRAW: u32 = ffi::GDK_PRIORITY_REDRAW as u32;
+
+// A mask covering all entries in ModifierType.
+#[doc(alias = "GDK_MODIFIER_MASK")]
+pub const MODIFIER_MASK: u32 = ffi::GDK_MODIFIER_MASK as u32;
```
sdroege, line 78: Isn't this the same as `gdk::ModifierType::all()`?

gdk4/src/lib.rs
```diff
@@ -65,3 +68,39 @@ pub use rectangle::Rectangle;
...
+#[doc(alias = "GDK_BUTTON_SECONDARY")]
+pub const BUTTON_SECONDARY: u32 = ffi::GDK_BUTTON_SECONDARY as u32;
+
+// Used as the return value for stopping the propagation of an event handler.
+#[doc(alias = "GDK_EVENT_STOP")]
+pub const EVENT_STOP: u32 = ffi::GDK_EVENT_STOP as u32;
+
+// Used as the return value for continuing the propagation of an event handler.
+#[doc(alias = "GDK_EVENT_PROPAGATE")]
+pub const EVENT_PROPAGATE: u32 = ffi::GDK_EVENT_PROPAGATE as u32;
```
sdroege, line 106: This should probably be handled like `glib::Continue`

sdroege (changes requested):
Seems ok otherwise :)

gdk4/src/lib.rs
bilelmoussaoui: Yes, definitely. There's an open issue to properly handle this one.

gdk4/src/key_event.rs
bilelmoussaoui: That was part of the original code in key_even btw

gdk4/src/lib.rs
bilelmoussaoui: Ah Indeed. Does it make sense to re-export then as I think that `gdk::ModifierType::all()` is less discoverablee than `MODIFIER_MASK`

gdk4/src/key_event.rs
sdroege: That's very bad :)

gdk4/src/lib.rs
sdroege: Well, at least export it with the correct type :)

bilelmoussaoui:
This looks fine as is for now, I will go ahead and merge it so that I can progress on #160

PR #159, opened 2021-01-11
Add Dialog::run_async

GuillaumeGomez:
Just like what we talked about in #149, here is the non-future (and non-blocking) version.

cc @sophie-h @sdroege @bilelmoussaoui

### DIFF
--- a/gtk4/src/dialog.rs
+++ b/gtk4/src/dialog.rs
@@ -45,15 +45,19 @@ pub trait DialogExtManual: 'static {
     /// Shows the dialog and returns a `Future` that resolves to the
     /// `ResponseType` on response.
     ///
-    /// ```ignore
-    /// let dialog = gtk::MessageDialogBuilder::new()
-    ///    .buttons(gtk::ButtonsType::OkCancel)
+    /// ```no_run
+    /// use gtk4::prelude::*;
+    ///
+    /// # async fn run() {
+    /// let dialog = gtk4::MessageDialogBuilder::new()
+    ///    .buttons(gtk4::ButtonsType::OkCancel)
     ///    .text("What is your answer?")
     ///    .build();
     ///
     /// let answer = dialog.run_future().await;
     /// dialog.close();
     /// println!("Answer: {:?}", answer);
+    /// # }
     /// ```
     fn run_future<'a>(&'a self) -> Pin<Box<dyn Future<Output = ResponseType> + 'a>>;
 
--- a/gtk4/src/dialog.rs
+++ b/gtk4/src/dialog.rs
@@ -74,12 +74,12 @@ pub trait DialogExtManual: 'static {
     ///    .text("What is your answer?")
     ///    .build();
     ///
-    /// dialog.run(|obj, answer| {
+    /// dialog.run_async(|obj, answer| {
     ///     obj.close();
     ///     println!("Answer: {:?}", answer);
     /// });
     /// ```
-    fn run<F: FnOnce(&Self, ResponseType) + 'static>(&self, f: F);
+    fn run_async<F: FnOnce(&Self, ResponseType) + 'static>(&self, f: F);
 }
 
 impl<O: IsA<Dialog> + IsA<Widget>> DialogExtManual for O {
@@ -122,7 +122,7 @@ impl<O: IsA<Dialog> + IsA<Widget>> DialogExtManual for O {
         })
     }
 
-    fn run<F: FnOnce(&Self, ResponseType) + 'static>(&self, f: F) {
+    fn run_async<F: FnOnce(&Self, ResponseType) + 'static>(&self, f: F) {
         let response_handler = Rc::new(RefCell::new(None));
         let response_handler_clone = response_handler.clone();
         let f = RefCell::new(Some(f));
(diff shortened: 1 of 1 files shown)


gtk4/src/dialog.rs
```diff
@@ -36,9 +36,27 @@ pub trait DialogExtManual: 'static {
...
+    ///    .buttons(gtk::ButtonsType::OkCancel)
+    ///    .text("What is your answer?")
+    ///    .build();
+    ///
+    /// dialog.run(|obj, answer| {
+    ///     obj.close();
+    ///     println!("Answer: {:?}", answer);
+    /// });
+    /// ```
+    fn run<F: Fn(&Self, ResponseType) + 'static>(&self, f: F);
```
sdroege, line 56:
```suggestion
    fn run<F: FnOnce(&Self, ResponseType) + 'static>(&self, f: F);
```

This is useless otherwise

gtk4/src/dialog.rs
```diff
@@ -49,9 +67,16 @@ impl<O: IsA<Dialog>> DialogExtManual for O {
...
-                self.as_ref().to_glib_none().0,
+                AsRef::<Dialog>::as_ref(self).to_glib_none().0,
                 widget.as_ref().to_glib_none().0,
             ))
         }
     }
+
+    fn run<F: Fn(&Self, ResponseType) + 'static>(&self, f: F) {
+        self.connect_response(move |s, response_type| {
+            f(s, response_type);
```
sdroege, line 78: You want to disconnect the signal here and close the dialog

GuillaumeGomez: To do that I have to store it inside a `Arc<RefCell<>>` I guess...

sdroege: Just a `RefCell<Option<_>>` is enough

gtk4/src/dialog.rs
```diff
@@ -49,9 +67,21 @@ impl<O: IsA<Dialog>> DialogExtManual for O {
...
             from_glib(ffi::gtk_dialog_get_response_for_widget(
-                self.as_ref().to_glib_none().0,
+                AsRef::<Dialog>::as_ref(self).to_glib_none().0,
                 widget.as_ref().to_glib_none().0,
             ))
         }
     }
+
+    fn run<F: FnOnce(&Self, ResponseType) + 'static>(&self, f: F) {
+        let response_handler = Arc::new(RefCell::new(None));
```
sdroege, line 77: Just a `RefCell<Option<_>>` is enough

sdroege (changes requested):

gtk4/src/dialog.rs
sdroege: Oh, no needs to be an `Rc` around it too (but not `Arc`).

But you need a `RefCell<Option<_>>` for the closure so you can call it exactly once (it's `FnOnce`!)

sdroege:
> This branch has conflicts that must be resolved

GuillaumeGomez:
CI passed! \o/

gtk4/src/dialog.rs
```diff
@@ -44,17 +45,41 @@ pub trait DialogExtManual: 'static {
...
     /// `ResponseType` on response.
     ///
-    /// ```ignore
-    /// let dialog = gtk::MessageDialogBuilder::new()
-    ///    .buttons(gtk::ButtonsType::OkCancel)
+    /// ```no_run
+    /// use gtk4::prelude::*;
+    ///
+    /// # async fn run() {
+    /// let dialog = gtk4::MessageDialogBuilder::new()
```
A6GibKm, line 52: Is this line properly indented?

sdroege (approved):

gtk4/src/dialog.rs
```diff
@@ -44,17 +45,41 @@ pub trait DialogExtManual: 'static {
...
+    ///    .buttons(gtk4::ButtonsType::OkCancel)
+    ///    .text("What is your answer?")
+    ///    .build();
+    ///
+    /// dialog.run(|obj, answer| {
+    ///     obj.close();
+    ///     println!("Answer: {:?}", answer);
+    /// });
+    /// ```
+    fn run<F: FnOnce(&Self, ResponseType) + 'static>(&self, f: F);
```
sdroege, line 82: Should probably be called `run_async()` because it's not like the original `run()`.

gtk4/src/dialog.rs
GuillaumeGomez: Yes.

gtk4/src/dialog.rs
bilelmoussaoui: ebassi suggested response as a name, which sounds better to me instead of calling it "run"

sophie-h: "response" won't work https://gtk-rs.org/gtk4-rs/gtk4/trait.DialogExt.html#tymethod.response

bilelmoussaoui: Ah :/ I wonder why some signals are named "$signal_name" like the response here while others are named "emit_$signal_name" like the `emit_close`.

sdroege: `emit_XXX()` are the ones we autogenerated if a signal is marked as action signal. `response()` has a emission function in C that is named like that.

sdroege:
@bilelmoussaoui This is ok then?

bilelmoussaoui:
Looks good to me, thanks!

PR #160, opened 2021-01-11
gdk: manually implement Event and it's subclasses

### LINKED ISSUE #50: gdk::Event is not a glib::Object
It's a fundamental type and needs manual bindings for itself and all its subclasses. See `glib::ParamSpec` for another example like that.

CC @bilelmoussaoui

bilelmoussaoui:
because it's a fundemental type
Fixes #50

I'm not sure about a few things, i will mark them on the code


gdk4/src/event.rs
```diff
@@ -0,0 +1,527 @@
...
+}
+
+unsafe impl Send for Event {}
+unsafe impl Sync for Event {}
+
+impl Event {
+    /*
+    pub fn downcast<T: EventTypeTrait>(self) -> Result<T, Event> {
+        unsafe {
+            if self.get_type() == T::static_type() {
```
bilelmoussaoui, line 39: I'm not sure how can I add a get_type method here ^^

gdk4/src/event.rs
```diff
@@ -0,0 +1,527 @@
...
+impl fmt::Display for Event {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("Event")
+    }
+}
+
+pub trait EventTypeTrait: StaticType + FromGlibPtrFull<*mut ffi::GdkEvent> + 'static {}
+
+macro_rules! define_event {
+    ($rust_type:ident, $ffi_type:path, $rust_type_offset:expr) => {
```
bilelmoussaoui, line 150: Does the `$rust_type_offset` makes any sense here? I couldn't understand what it means to do from looking at the ParamSpec example

gdk4/src/event.rs
```diff
@@ -0,0 +1,527 @@
...
+        }
+
+        impl StaticType for $rust_type {
+            fn static_type() -> Type {
+                unsafe { from_glib(*glib::g_param_spec_types.add($rust_type_offset)) }
+            }
+        }
+
+        unsafe impl Send for $rust_type {}
+        unsafe impl Sync for $rust_type {}
```
bilelmoussaoui, line 170: I suppose it's fine to implement those for an event? :) I wonder how can I find if a type can be Sync/Send?

gdk4/src/functions.rs
```diff
@@ -1,12 +1,73 @@
...
+use crate::{ContentDeserializer, ContentSerializer, Event};
 use glib::object::IsA;
 use glib::translate::*;
 use std::future;
+use std::mem;
 use std::pin::Pin;
 use std::ptr;
 
+#[doc(alias = "gdk_events_get_angle")]
+pub fn events_get_angle(event1: &Event, event2: &Event) -> Option<f64> {
```
bilelmoussaoui, line 12: Can the changes to the code be automated somehow? :) or should I just implement whatever allows me to have IsA<T> for my events?

gdk4/src/event.rs
sdroege: That was only needed there because for pspecs the `GType`s have no `get_type()` functions but there is just an array. You don't need that here

gdk4/src/event.rs
sdroege: Probably not, nothing in GTK is thread-safe.

gdk4/src/functions.rs
sdroege: You can probably go via `Deref` here or not? And then implement this function only at the most generic type that has it

gdk4/src/event.rs
sdroege: Do the different event types have different `GType`s?

I think you'll have to go with `gdk_event_get_event_type()` and `GdkEventType` here instead.

gdk4/src/functions.rs
bilelmoussaoui: What I meant here is the change of code from `
pub fn events_get_angle<P: IsA<Event>, Q: IsA<Event>>(event1: &P event2: &Q) -> Option<f64>;` to `pub fn events_get_angle(event1: &Event, event2: &Event) -> Option<f64>;`

There are more occurrences in GTK hence my question to avoid doing more stuff manually if possible

sdroege: Is that valid to call on any event types?

bilelmoussaoui: In some functions like CellArea::event<P: IsA<Event>> yes; but some of them have to be manually done to use the right event type

bilelmoussaoui: There are only 16 of them,  should be pretty doable, but would still love to know if there's a way to make a function accept any event in such case without having to downcast your events manually

sdroege: You'd have to introduce a new trait for that, or work with `AsRef<Event>` or similar

gdk4/src/event.rs
bilelmoussaoui: I ended up adding a new trait that allows me to know which event types are supported by which event :) Seems to build at least

gdk4/src/functions.rs
bilelmoussaoui: I ended up working with AsRef

gdk4/src/event.rs
```diff
@@ -0,0 +1,616 @@
...
+}
+
+impl fmt::Display for Event {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("Event")
+    }
+}
+
+pub trait EventTypeTrait: StaticType + FromGlibPtrFull<*mut ffi::GdkEvent> + 'static {
+    fn event_types() -> Vec<EventType>;
```
sdroege, line 215:
```suggestion
    fn event_types() -> &'static [EventType];
```

Or not? This won't change at runtime :)

gdk4/src/event.rs
```diff
@@ -0,0 +1,616 @@
...
+    }
+}
+
+impl fmt::Display for Event {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("Event")
+    }
+}
+
+pub trait EventTypeTrait: StaticType + FromGlibPtrFull<*mut ffi::GdkEvent> + 'static {
```
sdroege, line 214: I don't like the name. Maybe `EventKind`? Calling a trait `Trait` is strange

sdroege: Or `SpecificEvent`

gdk4/src/event.rs
```diff
@@ -0,0 +1,616 @@
...
+        #[doc(hidden)]
+        impl FromGlibPtrFull<*mut ffi::GdkEvent> for $rust_type {
+            unsafe fn from_glib_full(ptr: *mut ffi::GdkEvent) -> Self {
+                from_glib_full(ptr as *mut $ffi_type)
+            }
+        }
+
+        impl $rust_type {
+            pub fn upcast(self) -> Event {
+                unsafe { from_glib_full(self.to_glib_full() as *mut ffi::GdkEvent) }
```
sdroege, line 267: You could simply transmute here

gdk4/src/event.rs
```diff
@@ -0,0 +1,616 @@
...
+    fn static_type() -> Type {
+        unsafe { from_glib(ffi::gdk_event_get_type()) }
+    }
+}
+
+impl Event {
+    pub fn downcast<T: EventTypeTrait>(self) -> Result<T, Event> {
+        unsafe {
+            if T::event_types().contains(&self.get_event_type()) {
+                Ok(from_glib_full(self.to_glib_full()))
```
sdroege, line 35: You could simply transmute here

gdk4/src/event.rs
```diff
@@ -0,0 +1,616 @@
...
+
+macro_rules! define_event {
+    ($rust_type:ident, $ffi_type:path, $ffi_type_path:path, $event_event_types:expr) => {
+        // Can't use get_type here as this is not a boxed type but another fundamental type
+        glib::wrapper! {
+            #[derive(Debug)]
+            pub struct $rust_type(Shared<$ffi_type>);
+
+            match fn {
+
```
sdroege, line 226: Spurious empty line

gdk4/src/event.rs
```diff
@@ -0,0 +1,616 @@
...
+            }
+
+            pub fn upcast_ref(&self) -> &Event {
+                &*self
+            }
+        }
+
+        impl fmt::Display for $rust_type {
+            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+                f.write_str(stringify!($rust_type))
```
sdroege, line 277: Even nicer if these would print all the data of the event, same for the `Debug` impl

sdroege (commented):
I only looked at the type/trait definition and the casting. That looks good to me apart from the comments :)

The other API I assume you didn't break when copying from autogenerated to here and adjusting it, so I guess it is fine

gdk4/src/event.rs
bilelmoussaoui: I guess this one is not as simple the other one as I keep getting
```
cannot transmute between types of different sizes, or dependently-sized types
source type: `event::Event` (64 bits)
target type: `T` (this type does not have a fixed size)
```
with
```
Ok(mem::transmute::<Event, T>(self))
```

sdroege: I think you'd need to make `T: Sized` for this

sdroege: But should be fine either way, just some additional atomic operations.

bilelmoussaoui: > I think you'd need to make T: Sized for this

I have tried that one already but didn't help :/ 

I guess we can leave this one as is

bilelmoussaoui:
I just noticed I can move the Debug impl for events to the define_event macro. I will take care of that this evening

bilelmoussaoui:
> I only looked at the type/trait definition and the casting. That looks good to me apart from the comments :)

Thanks a lot for the reviews as always :) 

> The other API I assume you didn't break when copying from autogenerated to here and adjusting it, so I guess it is fine

Yes, it's a copy paste of the API with changes to use AsRef instead of IsA, nothing more

PR #161, opened 2021-01-11
examples: Use autobins

sophie-h:
Or is there a reason to use `autobins = false`?

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -3,7 +3,6 @@ name = "gtk4-rs-examples"
 version = "0.1.0"
 authors = ["The Gtk-rs Project Developers"]
 edition = "2018"
-autobins = false
 
 [dependencies]
 chrono = "0.4.19"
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
Because some examples might have specific dependencies such as the #152; or is that can still be made to work with autobins?

sophie-h:
> Because some examples might have specific dependencies

Just keep the explicit definitions for binaries that need additional settings. These things mix just fine.

bilelmoussaoui:
Oh cool, I wasn't aware of that. Do you mind sending a PR to do the same thing for gtk-rs? :)

sophie-h:
Yes, I will also backport run_future in the coming days

bilelmoussaoui:
Thank you!

sdroege:
> Yes, I will also backport run_future in the coming days (Edit: Yes=No, I don't mind, will do laughing)

That would be great, thanks :)

Issue #151, opened 2021-01-07
FileChooserNativeBuilder drops the filters


bilelmoussaoui:
Creating a dialog with a filter like 
```rust
let gif_filter = gtk::FileFilter::new();
gif_filter.add_mime_type("image/gif");
gif_filter.set_name(Some("GIF (image/gif)"));


let file_chooser = gtk::FileChooserNativeBuilder::new()
    .accept_label("Open")
    .cancel_label("Cancel")
    .modal(true)
    .title("Open a GIF to display")
    .transient_for(self)
    .action(gtk::FileChooserAction::Open)
    .filter(&gif_filter)
    .build();
```

wouldn't show up which is normal as the filter isn't a prop as the builder somehow assumes? and should use the method in `FileChooserExt::add_filter` to assign all those filters which would require writing the builder manually I guess

Works fine btw
```rust
let gif_filter = gtk::FileFilter::new();
gif_filter.add_mime_type("image/gif");
gif_filter.set_name(Some("GIF (image/gif)"));


let file_chooser = gtk::FileChooserNativeBuilder::new()
    .accept_label("Open")
    .cancel_label("Cancel")
    .modal(true)
    .title("Open a GIF to display")
    .transient_for(self)
    .action(gtk::FileChooserAction::Open)
    .build();
file_chooser.add_filter(&gif_filter);
```

sophie-h:
Sounds like the documented behavior https://developer.gnome.org/gtk3/stable/GtkFileChooser.html#gtk-file-chooser-set-filter

Maybe it's possible to make the list managed by `add_filter` a property upstream. Then we would get the feature that you want as well.

bilelmoussaoui:
Yeah it's just confusing. I think the property should have been called default-filter or something like that. I'm going to go ahead and close this one as I don't believe it's something actionable from our side. What would be nice to have though is for gir to generate comments for each property setter based on the props documentation if there are any?

piegamesde:
Ran into this again, was really confusing. Is there any reason why we don't panic in builders when setting the same property twice? This would have prompted me to the mistake a lot quicker

Issue #143, opened 2021-01-01
ApplicationBuilder should initialize gtk as well


bilelmoussaoui:
From IRC
> dcomp: Whats the difference between gtk4::Application::new and gtk4::ApplicationBuilder::new().build()? 
I would assume that the Builder just calls methods on a Application instance. But that doesn't seem to be the case
as I get a panic on the activate signal saying that gtk::init hasn't been called

-> bilelmoussaoui :  slomo: it needs it here as well https://github.com/gtk-rs/gtk4-rs/blob/master/gtk4/src/auto/application.rs#L65 i guess

sdroege:
Duplicate of https://github.com/gtk-rs/gtk4-rs/issues/118 or not?

bilelmoussaoui:
Not at all, we just need to manually add that builder and call `rt.init();` while the other issue is related to subclassing `gtk::Application`

PR #162, opened 2021-01-11
file_chooser_native: Fix filters

A6GibKm:
Filters have to be manually added after creating the file chooser gobject as it has no "filter" gproperty. Following the about dialog
builder, the api is changed so that it can take multiple filters.

Fixes: https://github.com/gtk-rs/gtk4-rs/issues/151.


A6GibKm:
The feels when you notice you are in a auto file....

bilelmoussaoui:
The same thing has to be done to both FileChooserDialogBuilder & FileChooserWidgetBuilder. Do you think you can do those as well? :)

sdroege:
Before we implement all builders manually, maybe there's something that can be improved in gir so that only the missing/wrong pieces can be implemented manually?

A6GibKm:
> The same thing has to be done to both FileChooserDialogBuilder & FileChooserWidgetBuilder. Do you think you can do those as well? :)

Yeah of course, but ... 

> Before we implement all builders manually, maybe there's something that can be improved in gir so that only the missing/wrong pieces can be implemented manually?

I don't know how that works to be honest, note that in this case, I had to modify the struct and two methods (build & filter which was renamed to add_filter), maybe the others can still be generated.

sdroege:
You could add a `manual_builder` or something and that would then expect a `struct FooBuilderManual` (that is `pub(crate)` only!). That would then be stored inside the autogenerated builder and provides it with additional fields, would have some kind of "add things to the properties array" function, and you would add another `impl FooBuilder` block in the manual code to add setter functions for the new fields.

bilelmoussaoui:
> You could add a `manual_builder` or something and that would then expect a `struct FooBuilderManual` (that is `pub(crate)` only!). That would then be stored inside the autogenerated builder and provides it with additional fields, would have some kind of "add things to the properties array" function, and you would add another `impl FooBuilder` block in the manual code to add setter functions for the new fields.

That can work, but it would also need a way to omit some properties from being used by the Builder. The issue here is that FileChooser has a filter property along with filters property; it seems the first one is the problematic one as setting it to something just doesn't work. 

I wonder how many other cases where the Builder has to be manually written other than #143.

sophie-h:
I think there is nothing wrong with the current behavior. See https://github.com/gtk-rs/gtk4-rs/issues/151#issuecomment-817380187 for details.

bilelmoussaoui:
Closing, see #151.

PR #163, opened 2021-01-12
Implement Deref for TemplateChild

BrainBlasted:
Allows using template children without .get()

### DIFF
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -980,6 +980,27 @@ where
     }
 }
 
+impl<T> std::ops::Deref for TemplateChild<T>
+where
+    T: ObjectType + FromGlibPtrNone<*mut <T as ObjectType>::GlibType>,
+{
+    type Target = T;
+
+    /// # Safety
+    ///
+    /// Since the template child may not be properly bound,
+    /// this cast is potentially dangerous if, for example,
+    /// the template child isn't bound or is of the wrong type.
+    /// The caller is responsible for ensuring that the template
+    /// child is bound and of the right type.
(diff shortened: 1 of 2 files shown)


gtk4/src/subclass/widget.rs
```diff
@@ -980,6 +980,17 @@ where
...
 }
 
+impl<T> std::ops::Deref for TemplateChild<T>
+where
+    T: ObjectType + FromGlibPtrNone<*mut <T as ObjectType>::GlibType>,
+{
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        unsafe { &*(&self.ptr as *const _ as *const T) }
```
sdroege, line 990: Needs a safety comment

bilelmoussaoui (approved):

gtk4/src/subclass/widget.rs
```diff
@@ -980,6 +980,24 @@ where
...
 
+impl<T> std::ops::Deref for TemplateChild<T>
+where
+    T: ObjectType + FromGlibPtrNone<*mut <T as ObjectType>::GlibType>,
+{
+    type Target = T;
+
+    /// # Safety
+    ///
+    /// Since the template child may not be properly bound,
```
ids1024, line 991: Can it be assumed that it will be impossible to call this from safe code when the template child isn't yet bound? Otherwise this is unsound.

Adding an assertion that it's non-null should be sufficient to avoid that, since it should be initialized to null by `default()`.

sdroege: Yes there should be an assertion here

BrainBlasted: Added.

PR #164, opened 2021-01-13
gtk: subclassing round 8

bilelmoussaoui:
Part of #56 

Adds subclassing support of the following types:

- GtkTreeDragDest
- GtkTreeDragSource
- GtkSelectionModel

Along with the following "empty" interfaces
- GtkAccessible
- GtkAppChooser
- GtkConstraintTarget
- GtkRoot
- GtkNative

### DIFF
--- a/gtk4/src/subclass/tree_drag_source.rs
+++ b/gtk4/src/subclass/tree_drag_source.rs
@@ -1,0 +1,74 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::{TreeDragSource, TreePath};
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::Cast;
+
+pub trait TreeDragSourceImpl: ObjectImpl {
+    fn row_draggable(&self, _tree_drag_source: &Self::Type, _path: &TreePath) -> bool {
+        // Assume the row is draggable by default
+        true
+    }
+    fn drag_data_get(&self, tree_drag_source: &Self::Type, path: &TreePath)
+        -> gdk::ContentProvider;
+    fn drag_data_delete(&self, tree_drag_source: &Self::Type, path: &TreePath) -> bool;
+}
+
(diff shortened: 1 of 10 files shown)


gtk4/src/subclass/tree_drag_source.rs
```diff
@@ -0,0 +1,90 @@
...
+
+pub trait TreeDragSourceImpl: ObjectImpl {
+    fn row_draggable(&self, tree_drag_source: &Self::Type, path: &TreePath) -> bool {
+        unsafe {
+            let type_ = ffi::gtk_tree_drag_source_get_type();
+            let iface = glib::gobject_ffi::g_type_default_interface_ref(type_)
+                as *mut ffi::GtkTreeDragSourceIface;
+            assert!(!iface.is_null());
+
+            let ret = ((*iface).row_draggable.as_ref().unwrap())(
```
sdroege, line 16: There is no default impl for this, is there? So this would instead just segfault, while `gtk_tree_drag_source_row_draggable()` simply returns `TRUE` if it's not implemented.

Please check this also for the other interface.

sdroege (commented):
Otherwise looks good

gtk4/src/subclass/tree_drag_source.rs
bilelmoussaoui: from my understanding is that the default behaviour is the row being draggable, which I think is an okayish default I guess. The other interfaces are mostly based on properties, like the Orientable one :)

sdroege: Yes but there is no default implementation, so this would panic on the `unwrap()`, and the default behaviour of the caller function is to return `true`. So instead you should make this a simple function that returns `true` here.

bilelmoussaoui: Ah indeed. I didn't notice that

Issue #116, opened 2020-12-11
Using custom widgets in ui templates


Malmz:
Trying to use a subclassed widget in another widget's ui template gives a runtime error:

Gtk-CRITICAL **: 18:40:50.481: Error building template class 'ParentCustomWidget' for an instance of type 'ParentCustomWidget': .:0:0 Invalid object type 'ChildCustomWidget'

[This](https://stackoverflow.com/questions/24235937/custom-gtk-widget-with-template-ui) stackoverflow thread seems to have some info.

Malmz:
Thanks! Calling `ChildCustomWidget::static_type()` in `ParentCustomWidget::class_init()` before binding the template works!

bilelmoussaoui:
Nice! Feel free to submit an example with such use case :)

piegamesde:
I think the issue still exists when the child class is not directly referenced as `template_child`. At least I ran into this error again today, and I think this is the cause.

PR #166, opened 2021-01-16
subclass/widget: Ensure TemplateChild types are registered

BrainBlasted:
Ensures types used as TemplateChild<T> are registered.
This gets rid of a step for application developers
using custom types within templates.

Closes https://github.com/gtk-rs/gtk4-rs/issues/116

### DIFF
--- a/examples/src/bin/composite_template.rs
+++ b/examples/src/bin/composite_template.rs
@@ -4,6 +4,7 @@
 
 use glib::subclass::prelude::*;
 use gtk::prelude::*;
+use gtk::subclass::widget::WidgetImplExt;
 use gtk::{gio, glib, CompositeTemplate};
 
 mod imp {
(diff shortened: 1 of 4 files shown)


BrainBlasted:
In testing, this doesn't seem to work unless you actually bind the template children as struct fields :/

sdroege:
Well it would require that the default constructor is actually called, and that the type of the widget is exactly the one declared here. But then it should improve the situation considerably.

Can you provide an example where it does not work?

BrainBlasted:
In one app I'm working on, instead of binding the custom widgets, I have a list of the custom widgets that I iterate over with GtkWidget's API for children/siblings

alatiera:
I usually also only bind only the children that are needed in the code, rather than every single one. Though I am guessing this might only be an issue if you don't bind a widget and it also happens to be your subclass. I don't think its very common to subclass something and not bind it afterwards to use, as opposed to not bindings random boxes and labels.

sdroege:
So all good here? Can we get an example that works with this but wouldn't have worked before? :)

sdroege (approved):
Code makes sense to me

BrainBlasted:
> So all good here? Can we get an example that works with this but wouldn't have worked before? :)

Pushed :)

sdroege (approved):

bilelmoussaoui:
Do you mind rebasing on top of master and update per the subclassing changes that landed in https://github.com/gtk-rs/gtk-rs/pull/270?

BrainBlasted:
Rebased.

PR #169, opened 2021-01-22
Subclassing round 10

bilelmoussaoui:
Adds subclassing support of the following types
- GtkTreeModel
- GtkTreeSortable

Part of #56

### DIFF
--- a/gtk4/src/subclass/tree_model.rs
+++ b/gtk4/src/subclass/tree_model.rs
@@ -1,0 +1,542 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::subclass::prelude::*;
+use crate::{TreeIter, TreeModel, TreeModelFlags, TreePath};
+use glib::translate::*;
+use glib::{Cast, Value};
+
+/// # Safety:
+///
+/// The TreeModel trait is unsafe because it expect you to create
+/// a TreeIter, creating such objects along with the optional
+/// `ref_node` & `unref_node` functions cannot be done with safe Rust
+pub unsafe trait TreeModelImpl: ObjectImpl {
+    fn row_changed(&self, tree_model: &Self::Type, path: &TreePath, iter: &TreeIter) {
+        self.parent_row_changed(tree_model, path, iter)
+    }
+
+    fn row_inserted(&self, tree_model: &Self::Type, path: &TreePath, iter: &TreeIter) {
+        self.parent_row_inserted(tree_model, path, iter)
+    }
+
+    fn row_has_child_toggled(&self, tree_model: &Self::Type, path: &TreePath, iter: &TreeIter) {
+        self.parent_row_has_child_toggled(tree_model, path, iter)
+    }
+
+    fn row_deleted(&self, tree_model: &Self::Type, path: &TreePath) {
+        self.parent_row_deleted(tree_model, path)
+    }
+
+    fn flags(&self, tree_model: &Self::Type) -> TreeModelFlags {
+        self.parent_flags(tree_model)
+    }
+
+    fn n_columns(&self, tree_model: &Self::Type) -> i32;
+    fn column_type(&self, tree_model: &Self::Type, index: i32) -> glib::Type;
+    fn iter(&self, tree_model: &Self::Type, path: &TreePath) -> Option<TreeIter>;
+    fn path(&self, tree_model: &Self::Type, iter: &TreeIter) -> TreePath;
+    fn value(&self, tree_model: &Self::Type, iter: &TreeIter, index: i32) -> Value;
+    fn iter_next(&self, tree_model: &Self::Type) -> Option<TreeIter>;
+
+    fn iter_previous(&self, tree_model: &Self::Type) -> Option<TreeIter> {
+        self.parent_iter_previous(tree_model)
+    }
+
+    fn iter_has_child(&self, tree_model: &Self::Type, iter: &TreeIter) -> bool;
+    fn iter_n_children(&self, tree_model: &Self::Type, iter: Option<&TreeIter>) -> i32;
+    fn iter_nth_child(
+        &self,
+        tree_model: &Self::Type,
+        parent: Option<&TreeIter>,
+        index: i32,
+    ) -> Option<TreeIter>;
+    fn iter_parent(&self, tree_model: &Self::Type, child: &TreeIter) -> Option<TreeIter>;
+
+    fn ref_node(&self, tree_model: &Self::Type, iter: &TreeIter) {
+        self.parent_ref_node(tree_model, iter)
+    }
+
+    fn unref_node(&self, tree_model: &Self::Type, iter: &TreeIter) {
+        self.parent_unref_node(tree_model, iter)
+    }
+}
+
+pub trait TreeModelImplExt: ObjectSubclass {
+    fn parent_row_changed(&self, tree_model: &Self::Type, path: &TreePath, iter: &TreeIter);
+    fn parent_row_inserted(&self, tree_model: &Self::Type, path: &TreePath, iter: &TreeIter);
+    fn parent_row_has_child_toggled(
+        &self,
+        tree_model: &Self::Type,
+        path: &TreePath,
+        iter: &TreeIter,
+    );
+    fn parent_row_deleted(&self, tree_model: &Self::Type, path: &TreePath);
+    fn parent_flags(&self, tree_model: &Self::Type) -> TreeModelFlags;
+    fn parent_iter_previous(&self, tree_model: &Self::Type) -> Option<TreeIter>;
+    fn parent_ref_node(&self, tree_model: &Self::Type, iter: &TreeIter);
+    fn parent_unref_node(&self, tree_model: &Self::Type, iter: &TreeIter);
+}
+
+impl<O: TreeModelImpl> TreeModelImplExt for O {
+    fn parent_row_changed(&self, tree_model: &Self::Type, path: &TreePath, iter: &TreeIter) {
+        unsafe {
+            let type_data = Self::type_data();
+            let parent_iface =
+                type_data.as_ref().parent_interface::<TreeModel>() as *const ffi::GtkTreeModelIface;
+
+            if let Some(f) = (*parent_iface).row_changed.as_ref() {
+                f(
+                    tree_model.unsafe_cast_ref::<TreeModel>().to_glib_none().0,
+                    mut_override(path.to_glib_none().0),
+                    mut_override(iter.to_glib_none().0),
+                );
+            }
+        }
+    }
+
+    fn parent_row_inserted(&self, tree_model: &Self::Type, path: &TreePath, iter: &TreeIter) {
+        unsafe {
+            let type_data = Self::type_data();
+            let parent_iface =
+                type_data.as_ref().parent_interface::<TreeModel>() as *const ffi::GtkTreeModelIface;
+
+            if let Some(f) = (*parent_iface).row_inserted.as_ref() {
+                f(
+                    tree_model.unsafe_cast_ref::<TreeModel>().to_glib_none().0,
+                    mut_override(path.to_glib_none().0),
+                    mut_override(iter.to_glib_none().0),
+                );
+            }
+        }
+    }
+
+    fn parent_row_has_child_toggled(
+        &self,
+        tree_model: &Self::Type,
+        path: &TreePath,
+        iter: &TreeIter,
+    ) {
+        unsafe {
+            let type_data = Self::type_data();
+            let parent_iface =
+                type_data.as_ref().parent_interface::<TreeModel>() as *const ffi::GtkTreeModelIface;
+
+            if let Some(f) = (*parent_iface).row_has_child_toggled.as_ref() {
+                f(
+                    tree_model.unsafe_cast_ref::<TreeModel>().to_glib_none().0,
+                    mut_override(path.to_glib_none().0),
+                    mut_override(iter.to_glib_none().0),
+                );
+            }
+        }
+    }
+
+    fn parent_row_deleted(&self, tree_model: &Self::Type, path: &TreePath) {
+        unsafe {
+            let type_data = Self::type_data();
+            let parent_iface =
+                type_data.as_ref().parent_interface::<TreeModel>() as *const ffi::GtkTreeModelIface;
+
+            if let Some(f) = (*parent_iface).row_deleted.as_ref() {
+                f(
+                    tree_model.unsafe_cast_ref::<TreeModel>().to_glib_none().0,
+                    mut_override(path.to_glib_none().0),
+                );
+            }
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


gtk4/src/subclass/font_chooser.rs
```diff
@@ -0,0 +1,196 @@
...
+use glib::translate::*;
+use glib::{Cast, GString, ObjectExt, Quark};
+use once_cell::sync::Lazy;
+use pango::{FontFace, FontFamily, FontMap};
+
+pub trait FontChooserImpl: ObjectImpl {
+    fn get_font_family(&self, font_chooser: &Self::Type) -> Option<FontFamily>;
+    fn get_font_face(&self, font_chooser: &Self::Type) -> Option<FontFace>;
+    fn get_font_size(&self, font_chooser: &Self::Type) -> i32;
+    // fn set_filter_func(&self, font_chooser: &Self::Type) -> Option<Box<dyn Fn(&FontFamily, &FontFace) -> bool + 'static>>;
```
sdroege, line 14: This is the wrong way around. The caller would pass a filter function in here.

You'd need to define a new type for that which wraps the function pointer, user_data and destroy notify, which can be called from safe Rust, etc.

I think we have such a case elsewhere in GTK already. The foreach function of GtkContainer or so

sdroege: Also this function is not optional. You have to add bindings for it.

gtk4/src/subclass/font_chooser.rs
```diff
@@ -0,0 +1,196 @@
...
+    font_chooser: *mut ffi::GtkFontChooser,
+) -> *mut pango::ffi::PangoFontFamily {
+    let instance = &*(font_chooser as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap = from_glib_borrow::<_, FontChooser>(font_chooser);
+
+    let ret = imp.get_font_family(wrap.unsafe_cast_ref());
+    if let Some(font_family) = ret {
+        let font_family = font_family.to_glib_full();
+        wrap.set_qdata(*FONT_CHOOSER_GET_FONT_FAMILY_QUARK, font_family);
```
sdroege, line 103: You'd leak this pointer here. That's probably not intended

sdroege: Also same below

gtk4/src/subclass/tree_model.rs
```diff
@@ -0,0 +1,503 @@
...
+            from_glib(ret)
+        }
+    }
+
+    fn get_n_columns(&self, tree_model: &Self::Type) -> i32;
+    fn get_column_type(&self, tree_model: &Self::Type, index: i32) -> glib::Type;
+    fn get_iter(&self, tree_model: &Self::Type, path: &TreePath) -> Option<TreeIter>;
+    fn get_path(&self, tree_model: &Self::Type, iter: &TreeIter) -> TreePath;
+    fn get_value(&self, tree_model: &Self::Type, iter: &TreeIter, index: i32) -> Value;
+    fn iter_next(&self, tree_model: &Self::Type) -> Option<TreeIter>;
```
sdroege, line 104: Can we actually implement all these functions? Do we have ways to create a tree iter out of thin air?

gtk4/src/subclass/tree_model.rs
```diff
@@ -0,0 +1,503 @@
...
+    let instance = &*(tree_model as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    let iter: Borrowed<TreeIter> = from_glib_borrow(iterptr);
+
+    imp.get_path(
+        from_glib_borrow::<_, TreeModel>(tree_model).unsafe_cast_ref(),
+        &iter,
+    )
+    .to_glib_full() as *mut _
```
sdroege, line 341: This misses an annotation in the C code btw

gtk4/src/subclass/tree_model.rs
```diff
@@ -0,0 +1,503 @@
...
+    let imp = instance.get_impl();
+
+    let iter: Borrowed<TreeIter> = from_glib_borrow(iterptr);
+
+    let ret = imp.get_value(
+        from_glib_borrow::<_, TreeModel>(tree_model).unsafe_cast_ref(),
+        &iter,
+        index,
+    );
+    *valueptr = *ret.to_glib_none().0;
```
sdroege, line 360: This is wrong. You'd need to `mem::forget()` the value afterwards or use `ManuallyDrop` (better), otherwise the content of the value will be freed twice

Also the C annotations are completely wrong. This should be `out caller-allocates` AFAIU.

gtk4/src/subclass/tree_model.rs
```diff
@@ -0,0 +1,503 @@
...
+    fn iter_n_children(&self, tree_model: &Self::Type, iter: Option<&TreeIter>) -> i32;
+    fn iter_nth_child(
+        &self,
+        tree_model: &Self::Type,
+        parent: Option<&TreeIter>,
+        index: i32,
+    ) -> Option<TreeIter>;
+    fn iter_parent(&self, tree_model: &Self::Type, child: &TreeIter) -> Option<TreeIter>;
+
+    fn ref_node(&self, tree_model: &Self::Type, iter: &TreeIter) {
```
sdroege, line 139: Can we safely (or at all) implement this and the unref function? How is that supposed to work?

gtk4/src/subclass/tree_sortable.rs
```diff
@@ -0,0 +1,176 @@
...
+
+use crate::{SortType, TreeSortable};
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::Cast;
+
+/*
+TODO: figure out
+- set_sort_func
+- sset_default_sort_func
```
sdroege, line 11: They're all mandatory so you'd have to implement them before it can be merged.

gtk4/src/subclass/tree_sortable.rs
```diff
@@ -0,0 +1,176 @@
...
+
+pub trait TreeSortableImpl: ObjectImpl {
+    fn sort_column_changed(&self, tree_sortable: &Self::Type) {
+        unsafe {
+            let type_ = ffi::gtk_print_operation_preview_get_type();
+            let iface = glib::gobject_ffi::g_type_default_interface_ref(type_)
+                as *mut ffi::GtkTreeSortableIface;
+            assert!(!iface.is_null());
+
+            if let Some(f) = (*iface).sort_column_changed.as_ref() {
```
sdroege, line 22: There are no default impls for anything here

gtk4/src/subclass/font_chooser.rs
bilelmoussaoui: > You'd leak this pointer here. That's probably not intended

But the fonction annotates the `FontFamily`/`FontFace` as `return none`, or I misunderstood something? :)

sdroege: You store a raw pointer in the qdata. That pointer is never freed. You would have to store some struct around a raw pointer in the qdata that implements `Drop` in a way that the pointer is freed.

gtk4/src/subclass/tree_model.rs
bilelmoussaoui: alot of thing are not annotated in TreeModel*  :/ I will try to bring that up for gtk4.2 and hopefully fix them. The thing is I'm afraid of that nobody touched for a decade code.

gtk4/src/subclass/tree_model.rs
bilelmoussaoui: The API docs mentions that the functions handles a ref counter of the currently displayed nodes so that they are cached. Here's an example from gtk of how such methods looks like 
```c
static void
gtk_tree_model_ref_count_ref_node (GtkTreeModel *model,
                                   GtkTreeIter  *iter)
{
  NodeInfo *info;
  GtkTreeModelRefCount *ref_model = GTK_TREE_MODEL_REF_COUNT (model);

  info = g_hash_table_lookup (ref_model->priv->node_hash, iter->user_data);
  if (!info)
    {
      info = node_info_new ();

      g_hash_table_insert (ref_model->priv->node_hash, iter->user_data, info);
    }

  info->ref_count++;
}

static void
gtk_tree_model_ref_count_unref_node (GtkTreeModel *model,
                                     GtkTreeIter  *iter)
{
  NodeInfo *info;
  GtkTreeModelRefCount *ref_model = GTK_TREE_MODEL_REF_COUNT (model);

  info = g_hash_table_lookup (ref_model->priv->node_hash, iter->user_data);
  g_assert (info != NULL);
  g_assert (info->ref_count > 0);

  info->ref_count--;
}
```
But I'm not sure where that ref counting is used, as in where the caching happens. I guess we can mark them as unsafe and leave figuring out how this part of the api is supposed to work to someone that actually uses GtkTreeModel

gtk4/src/subclass/tree_model.rs
bilelmoussaoui: We don't actually have a way to create a TreeIter currently no, but it's just a sturct around an int & 3 random user_data pointers; I will have to look at how other bindings handle this (if they do)

gtk4/src/subclass/tree_model.rs
sdroege: That would mean making the trait unsafe, not the function.

gtk4/src/subclass/tree_model.rs
bilelmoussaoui: So none of the bindings provides a way to create a TreeIter but as we made the whole trait unsafe, I guess that's okay as is.

gtk4/src/subclass/font_chooser.rs
bilelmoussaoui: I have did something, but not very sure if what I wrote makes sense or not :)

gtk4/src/subclass/font_chooser.rs
```diff
@@ -0,0 +1,231 @@
...
+
+use crate::FontChooser;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, GString, ObjectExt, Quark};
+use once_cell::sync::Lazy;
+use pango::{FontFace, FontFamily, FontMap};
+
+#[derive(Debug)]
+pub struct FilterCallback {
```
bilelmoussaoui, line 11: I wonder if we should provide some way to convert a rust callback to a FilterCallback or do we expect the user to write unsafe code in order to implement the FontChooser?

gtk4/src/subclass/tree_model.rs
sdroege: I'd rather not have this trait for now then. It clearly shows the it's not usable yet and not enough thought was put into how to make it work in Rust. Having the trait around in an unusable form will just confuse users that try to use it.

Maybe move this into a separate WIP PR?

gtk4/src/subclass/font_chooser.rs
sdroege: That safe function already exists: https://gtk-rs.org/gtk4-rs/gtk4/prelude/trait.FontChooserExt.html#tymethod.set_filter_func

You get this callback passed via C and in the implementation side of the font chooser you only need to pass it around and call it.

sdroege: I.e. you need to also provide a way to safely call this, see https://github.com/gtk-rs/gtk-rs/blob/c58fab4fec1254244f598c4eac7f9310184c5996/gtk/src/subclass/container.rs#L258-L266

And you'll have to understand how this works memory management wise. Does the subclass own it and has to free it? Or the base class? You will probably have to implement `Drop` on this, which then would pass the `user_data` into the `destroy_notify` (if the `destroy_notify` is not `NULL`).

gtk4/src/subclass/font_chooser.rs
sdroege: Looks good, make it `ptr::NonNull` instead of a plain raw pointer and it's perfect :)

gtk4/src/subclass/tree_model_filter.rs
```diff
@@ -0,0 +1,142 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::subclass::object::ObjectImpl;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, Object, Value};
+
+use crate::{TreeIter, TreeModel, TreeModelFilter};
+
+pub trait TreeModelFilterImpl: TreeModelFilterImplExt + ObjectImpl {
```
sdroege, line 10:
```suggestion
pub trait TreeModelFilterImpl: TreeModelFilterImplExt + TreeModelImpl + TreeDragSourceImpl + ObjectImpl {
```

If I read the docs correctly

sdroege: Ah nevermind

gtk4/src/subclass/tree_sortable.rs
```diff
@@ -0,0 +1,224 @@
...
+use glib::Cast;
+
+#[derive(Debug)]
+pub struct TreeIterCompareCallback {
+    compare_func: ffi::GtkTreeIterCompareFunc,
+    user_data: glib::ffi::gpointer,
+    destroy_ptr: glib::ffi::GDestroyNotify,
+}
+
+pub trait TreeSortableImpl: ObjectImpl {
```
sdroege, line 15:
```suggestion
pub trait TreeSortableImpl: TreeModelImpl + ObjectImpl {
```

Though. See the `Prerequisites` in the docs.

gtk4/src/subclass/tree_sortable.rs
```diff
@@ -0,0 +1,224 @@
...
+use crate::{SortType, TreeSortable};
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::Cast;
+
+#[derive(Debug)]
+pub struct TreeIterCompareCallback {
+    compare_func: ffi::GtkTreeIterCompareFunc,
+    user_data: glib::ffi::gpointer,
+    destroy_ptr: glib::ffi::GDestroyNotify,
```
sdroege, line 12: Same question about memory management here

sdroege: And needs to be callable

gtk4/src/subclass/tree_model.rs
bilelmoussaoui: Sure, I can move the FontChooser out of this PR as the rest is dependant on TreeModel. It could definitely use some better approach to make it usable from Rust but requires someone who is actually familiar with the Tree* stuff in GTK :) 
Also I couldn't find that many apps that has custom TreeModels, most of them are written in C like sysprof.

bilelmoussaoui:
I have rebased this one and updated per the latest subclassing changes, just in case someone wants to have a look at it someday :) it should make things simpler to pick up.

bilelmoussaoui:
Closing this one as I won't be finishing it. People are free to take over from that branch or just switch to using Gtk*View widgets instead

PR #171, opened 2021-01-23
gir cleanup

bilelmoussaoui:
- remove unused/removed stuff
- use manual instead of ignore when documented already. The idea is to reduce the usage of ignore & document all the manual stuff

### DIFF
--- a/gtk4/src/auto/actionable.rs
+++ b/gtk4/src/auto/actionable.rs
@@ -35,9 +35,6 @@ pub trait ActionableExt: 'static {
     #[doc(alias = "gtk_actionable_set_action_name")]
     fn set_action_name(&self, action_name: Option<&str>);
 
-    //#[doc(alias = "gtk_actionable_set_action_target")]
-    //fn set_action_target(&self, format_string: &str, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs);
-
     #[doc(alias = "gtk_actionable_set_action_target_value")]
     fn set_action_target_value(&self, target_value: Option<&glib::Variant>);
 
@@ -45,6 +42,11 @@ pub trait ActionableExt: 'static {
     fn set_detailed_action_name(&self, detailed_action_name: &str);
 
     fn connect_property_action_name_notify<F: Fn(&Self) + 'static>(&self, f: F) -> SignalHandlerId;
+
+    fn connect_property_action_target_notify<F: Fn(&Self) + 'static>(
+        &self,
+        f: F,
+    ) -> SignalHandlerId;
 }
 
 impl<O: IsA<Actionable>> ActionableExt for O {
@@ -72,10 +74,6 @@ impl<O: IsA<Actionable>> ActionableExt for O {
             );
         }
     }
-
-    //fn set_action_target(&self, format_string: &str, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs) {
-    //    unsafe { TODO: call ffi:gtk_actionable_set_action_target() }
-    //}
 
     fn set_action_target_value(&self, target_value: Option<&glib::Variant>) {
         unsafe {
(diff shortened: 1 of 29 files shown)


gtk4/src/constraint_guide.rs
```diff
@@ -6,43 +6,70 @@ use glib::IsA;
...
+    fn get_max_size(&self) -> (Option<i32>, Option<i32>) {
         unsafe {
-            ffi::gtk_constraint_guide_get_max_size(
-                self.as_ref().to_glib_none().0,
-                width as *mut _,
-                height as *mut _,
-            );
+            let width = std::ptr::null_mut::<i32>();
+            let height = std::ptr::null_mut::<i32>();
+            ffi::gtk_constraint_guide_get_max_size(self.as_ref().to_glib_none().0, width, height);
```
sdroege, line 23: This is wrong. You must pass a pointer to an integer here.

```rust
let mut width = MaybeUninit::uninit();
let mut height = MaybeUninit::uninit();
ffi::gtk_constraint_guide_get_max_size(self.as_ref().to_glib_none().0, width.as_mut_ptr(), height.as_mut_ptr());
let width = width.assume_init();
let height = height.assume_init();
```

The whole thing also can't return `None`. It will always return some kind of integer for both

gtk4/src/constraint_guide.rs
```diff
@@ -6,43 +6,70 @@ use glib::IsA;
...
-    fn get_min_size(&self, width: i32, height: i32) {
+    fn get_min_size(&self) -> (Option<i32>, Option<i32>) {
         unsafe {
-            ffi::gtk_constraint_guide_get_min_size(
-                self.as_ref().to_glib_none().0,
-                width as *mut _,
-                height as *mut _,
-            );
+            let width = std::ptr::null_mut::<i32>();
+            let height = std::ptr::null_mut::<i32>();
```
sdroege, line 41: As as above

gtk4/src/constraint_guide.rs
```diff
@@ -6,43 +6,70 @@ use glib::IsA;
...
-    fn get_nat_size(&self, width: i32, height: i32) {
+    fn get_nat_size(&self) -> (Option<i32>, Option<i32>) {
         unsafe {
-            ffi::gtk_constraint_guide_get_nat_size(
-                self.as_ref().to_glib_none().0,
-                width as *mut _,
-                height as *mut _,
-            );
+            let width = std::ptr::null_mut::<i32>();
+            let height = std::ptr::null_mut::<i32>();
```
sdroege, line 60: Same as above

gtk4/src/drop_target.rs
```diff
@@ -0,0 +1,38 @@
...
+            F: Fn(&DropTarget, &glib::Value, f64, f64) -> bool + 'static,
+        >(
+            this: *mut ffi::GtkDropTarget,
+            value: *mut glib::gobject_ffi::GValue,
+            x: libc::c_double,
+            y: libc::c_double,
+            f: glib::ffi::gpointer,
+        ) -> glib::ffi::gboolean {
+            let f: &F = &*(f as *const F);
+            f(&from_glib_borrow(this), &from_glib_none(value), x, y).to_glib()
```
sdroege, line 24: You can cast a `*const GValue` or `*mut GValue` directly to a `&glib::Value`. They're equivalent.

`from_glib_none()` here would instead create a copy.

gtk4/src/auto/actionable.rs
```diff
@@ -118,6 +116,33 @@ impl<O: IsA<Actionable>> ActionableExt for O {
...
         }
     }
+
+    fn connect_property_action_target_notify<F: Fn(&Self) + 'static>(
+        &self,
+        f: F,
+    ) -> SignalHandlerId {
+        unsafe extern "C" fn notify_action_target_trampoline<P, F: Fn(&P) + 'static>(
+            this: *mut ffi::GtkActionable,
+            _param_spec: glib::ffi::gpointer,
```
sdroege, line 126: Why ignore the pspec?

gtk4/src/auto/entry_buffer.rs
```diff
@@ -64,16 +64,6 @@ pub trait EntryBufferExt: 'static {
     #[doc(alias = "gtk_entry_buffer_emit_inserted_text")]
     fn emit_inserted_text(&self, position: u32, chars: &str, n_chars: u32);
 
-    fn get_property_length(&self) -> u32;
```
sdroege, line 67: Why do these all disappear?

sdroege (commented):
Looks good to me apart from those comments

gtk4/src/auto/entry_buffer.rs
bilelmoussaoui: Because there are manual version of them, somehow marking them as ignore didn't actually ignore them?

sdroege: That seems like a bug in gir :)

gtk4/src/constraint_guide.rs
bilelmoussaoui: Ah, the annotations mislead me

gtk4/src/auto/actionable.rs
bilelmoussaoui: ah, somehow gir does that all the time for any pspec used in a trampoline

sdroege: That seems like a bug, can you create an issue at least?

gtk4/src/drop_target.rs
sdroege:
```rust
f(&from_glib_borrow(this), &*(value as *const glib::Value), x, y).to_glib()
```

A lot of API using `glib::Value` does such things internally.

gtk4/src/auto/actionable.rs
bilelmoussaoui: Sure, https://github.com/gtk-rs/gir/issues/1042

gtk4/src/drop_target.rs
bilelmoussaoui: Thanks!

bilelmoussaoui:
I opened #173 to fix an issue in EntryBuffer subclassing I noticed from working on this one. I will go ahead and merge this one

PR #174, opened 2021-01-26
Subclassing round 11

bilelmoussaoui:
Part of #56 

- GtkCellArea
- GtkCellAreaContext
- GtkCellEditable
- GtkCellLayout

It still needs some cleanup but should be ready for an initial review :)

### DIFF
--- a/gtk4/src/subclass/cell_area.rs
+++ b/gtk4/src/subclass/cell_area.rs
@@ -1,0 +1,1074 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::subclass::prelude::*;
+use crate::{
+    CellArea, CellAreaContext, CellRenderer, CellRendererState, DirectionType, SizeRequestMode,
+    Snapshot, TreeIter, TreeModel, Widget,
+};
+use glib::translate::*;
+use glib::{Cast, IsA, Object, ParamSpec, Value};
+use std::mem;
+
+#[derive(Debug)]
+pub struct CellCallback {
+    callback: ffi::GtkCellCallback,
+    user_data: glib::ffi::gpointer,
+}
+
+impl CellCallback {
+    pub fn call<R: IsA<CellRenderer>>(&self, cell_renderer: &R) -> bool {
+        unsafe {
+            if let Some(callback) = self.callback {
+                from_glib(callback(
+                    cell_renderer.as_ref().to_glib_none().0,
+                    self.user_data,
+                ))
+            } else {
+                // true to stop iterating over cells
+                true
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct CellCallbackAllocate {
+    callback: ffi::GtkCellAllocCallback,
+    user_data: glib::ffi::gpointer,
+}
+
+impl CellCallbackAllocate {
+    pub fn call<R: IsA<CellRenderer>>(
+        &self,
+        cell_renderer: &R,
+        cell_area: &gdk::Rectangle,
+        cell_background: &gdk::Rectangle,
+    ) -> bool {
+        unsafe {
+            if let Some(callback) = self.callback {
+                from_glib(callback(
+                    cell_renderer.as_ref().to_glib_none().0,
+                    cell_area.to_glib_none().0,
+                    cell_background.to_glib_none().0,
+                    self.user_data,
+                ))
+            } else {
+                // true to stop iterating over cells
+                true
+            }
+        }
+    }
+}
+
+pub trait CellAreaImpl: CellAreaImplExt + ObjectImpl {
+    fn cell_properties() -> &'static [ParamSpec] {
+        &[]
+    }
+
+    fn set_cell_property<R: IsA<CellRenderer>>(
+        &self,
+        _obj: &Self::Type,
+        _renderer: &R,
+        _id: usize,
+        _value: &Value,
+        _pspec: &ParamSpec,
+    ) {
+        unimplemented!()
+    }
+
+    fn cell_property<R: IsA<CellRenderer>>(
+        &self,
+        _obj: &Self::Type,
+        _renderer: &R,
+        _id: usize,
+        _pspec: &ParamSpec,
+    ) -> Value {
+        unimplemented!()
+    }
+
+    fn activate<P: IsA<CellAreaContext>, W: IsA<Widget>>(
+        &self,
+        cell_area: &Self::Type,
+        context: &P,
+        widget: &W,
+        area: &gdk::Rectangle,
+        flags: CellRendererState,
+        edit_only: bool,
+    ) -> bool {
+        self.parent_activate(cell_area, context, widget, area, flags, edit_only)
+    }
+
+    fn add<R: IsA<CellRenderer>>(&self, cell_area: &Self::Type, renderer: &R) {
+        self.parent_add(cell_area, renderer)
+    }
+
+    fn apply_attributes<M: IsA<TreeModel>>(
+        &self,
+        cell_area: &Self::Type,
+        tree_model: &M,
+        iter: &TreeIter,
+        is_expander: bool,
+        is_expanded: bool,
+    ) {
+        self.parent_apply_attributes(cell_area, tree_model, iter, is_expander, is_expanded)
+    }
+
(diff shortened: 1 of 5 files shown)


gtk4/src/subclass/cell_layout.rs
```diff
@@ -0,0 +1,368 @@
...
+use crate::{CellArea, CellLayout, CellRenderer};
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::Cast;
+
+#[derive(Debug)]
+pub struct CellLayoutDataCallback {
+    callback: ffi::GtkCellLayoutDataFunc,
+    user_data: glib::ffi::gpointer,
+    destroy_notify: glib::ffi::GDestroyNotify,
```
sdroege, line 12: Same as elsewhere: memory management and making it callable

gtk4/src/subclass/cell_layout.rs
```diff
@@ -0,0 +1,368 @@
...
+}
+
+unsafe extern "C" fn cell_layout_get_cells<T: CellLayoutImpl>(
+    cell_layout: *mut ffi::GtkCellLayout,
+) -> *mut glib::ffi::GList {
+    let instance = &*(cell_layout as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    let cells = imp.get_cells(from_glib_borrow::<_, CellLayout>(cell_layout).unsafe_cast_ref());
+    ToGlibContainerFromSlice::to_glib_container_from_slice(&cells).0
```
sdroege, line 367: You need to keep the cells themselves alive, they're `transfer none` and only the container is owned by the caller

gtk4/src/subclass/cell_layout.rs
```diff
@@ -0,0 +1,368 @@
...
+    fn get_cells(&self, cell_layout: &Self::Type) -> Vec<CellRenderer> {
+        unsafe {
+            let type_ = ffi::gtk_cell_layout_get_type();
+            let iface = glib::gobject_ffi::g_type_default_interface_ref(type_)
+                as *mut ffi::GtkCellLayoutIface;
+            assert!(!iface.is_null());
+
+            let f = (*iface).get_cells.as_ref().unwrap();
+
+            let cells = f(cell_layout.unsafe_cast_ref::<CellLayout>().to_glib_none().0);
```
sdroege, line 69: This is `transfer container`, not `transfer none`

gtk4/src/subclass/cell_area_context.rs
```diff
@@ -0,0 +1,208 @@
...
+    natural_width: *mut libc::c_int,
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<CellAreaContext> = from_glib_borrow(ptr);
+
+    let ret = imp.get_preferred_width_for_height(wrap.unsafe_cast_ref(), height);
+    if let Some((min_width, nat_width)) = ret {
+        *minimum_width = min_width;
+        *natural_width = nat_width;
```
sdroege, line 194: You should probably always set these to something instead of leaving them uninitialized

gtk4/src/subclass/cell_area_context.rs
```diff
@@ -0,0 +1,208 @@
...
+    natural_height: *mut libc::c_int,
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<CellAreaContext> = from_glib_borrow(ptr);
+
+    let ret = imp.get_preferred_height_for_width(wrap.unsafe_cast_ref(), width);
+    if let Some((min_height, nat_height)) = ret {
+        *minimum_height = min_height;
+        *natural_height = nat_height;
```
sdroege, line 177: Here too

gtk4/src/subclass/cell_area.rs
```diff
@@ -0,0 +1,1017 @@
...
+use glib::subclass::object::ObjectImpl;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::{Cast, IsA, Object};
+use std::mem;
+
+#[derive(Debug)]
+pub struct CellCallback {
+    callback: ffi::GtkCellCallback,
+    user_data: glib::ffi::gpointer,
```
sdroege, line 16: Needs to be callable (memory management is clear here at least)

gtk4/src/subclass/cell_area.rs
```diff
@@ -0,0 +1,1017 @@
...
+}
+
+#[derive(Debug)]
+pub struct CellCallbackAllocate {
+    callback: ffi::GtkCellAllocCallback,
+    user_data: glib::ffi::gpointer,
```
sdroege, line 22: Same

sdroege (commented):
Otherwise seems ok but I didn't review in detail

gtk4/src/subclass/cell_layout.rs
bilelmoussaoui: Not sure how I'm supposed to figure out the memory management here? I suppose the difference between this callback & the others on this merge request is the destroy_notify. Does the when to call the `destroy_notify` is the figuring memory management part?

sdroege: You need to call it as part of the `Drop` impl on this type

bilelmoussaoui: Ah, I see. Thanks!

sdroege: And for calling you'd add some kind of `rn call(&self, all, the arguments) -> return_type` that internally calls `callback` with `user_data` and the other arguments.

sdroege: Yeah looks good how you implemented that now. You'd then pass a `CellLayoutDataCallback` through all the Rust code so that implementors can call it whenever they want and also store it away somewhere for calling it later.

gtk4/src/subclass/cell_layout.rs
bilelmoussaoui: I suppose storing the cells on a qdata is enough to keep them alive? :)

bilelmoussaoui:
This should be good enough for another review

gtk4/src/subclass/cell_layout.rs
sdroege: I guess so

gtk4/src/subclass/cell_layout.rs
```diff
@@ -0,0 +1,489 @@
...
+
+    fn cells(&self, cell_layout: &Self::Type) -> Vec<CellRenderer> {
+        self.parent_cells(cell_layout)
+    }
+
+    fn set_cell_data_func<R: IsA<CellRenderer>>(
+        &self,
+        cell_layout: &Self::Type,
+        cell: &R,
+        callback: Option<&CellLayoutDataCallback>,
```
sdroege, line 71:
```suggestion
        callback: Option<&CellLayoutDataCallback>,
```

No reference here and in the `parent` function. Somehow the callback needs to be stored by the impl.

gtk4/src/subclass/cell_layout.rs
```diff
@@ -0,0 +1,489 @@
...
+            let cells = f(cell_layout.unsafe_cast_ref::<CellLayout>().to_glib_none().0);
+            FromGlibPtrArrayContainerAsVec::from_glib_container_as_vec(cells)
+        }
+    }
+
+    fn parent_set_cell_data_func<R: IsA<CellRenderer>>(
+        &self,
+        cell_layout: &Self::Type,
+        cell: &R,
+        callback: Option<&CellLayoutDataCallback>,
```
sdroege, line 194:
```suggestion
        callback: Option<CellLayoutDataCallback>,
```

And here also to prevent double frees.

sdroege (approved):
Looks good otherwise (and the other callbacks must be references!)

bilelmoussaoui:
> Looks good otherwise (and the other callbacks must be references!)

Thanks for the reviews as always :) so if I understood correctly, only the ones that don't provide a `GDestroyNotify` have to be passed by reference?

sdroege:
> so if I understood correctly, only the ones that don't provide a `GDestroyNotify` have to be passed by reference?

`(scope notified)` vs. `(scope call)`, yes.

sdroege (approved):

PR #180, opened 2021-01-28
widget: Implement Debug for TickCallbackId

A6GibKm:
So that any custom widget which has to store this id can be derived.

### DIFF
--- a/gtk4/src/widget.rs
+++ b/gtk4/src/widget.rs
@@ -79,6 +79,12 @@ pub struct TickCallbackId {
     widget: WeakRef<Widget>,
 }
 
+impl PartialEq for TickCallbackId {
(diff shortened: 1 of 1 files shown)


sdroege:
Or maybe implement https://github.com/gtk-rs/gtk-rs/issues/45 and then this and various other traits could simply be derived here :)

bilelmoussaoui:
Thanks! Do you think you want to tackle the issue mentioned by @sdroege? otherwise we can merge this as is :)

A6GibKm:
@bilelmoussaoui  Yes it is something I want to look into, but considering that it is in another repo and might take me a non-trivial amount of effort I suggest this can be merged as is, and should be updated at a later time.

PR #183, opened 2021-02-10
gdk: bind gdk_gl_texture_new

elmarco:
Fixes: https://github.com/gtk-rs/gtk-rs/issues/295

### DIFF
--- a/gdk4/Gir.toml
+++ b/gdk4/Gir.toml
@@ -33,7 +33,6 @@ generate = [
     "Gdk.FullscreenMode",
     "Gdk.GLContext",
     "Gdk.GLError",
-    "Gdk.GLTexture",
     "Gdk.Gravity",
     "Gdk.InputSource",
     "Gdk.KeyMatch",
@@ -315,6 +314,13 @@ status = "generate"
     [[object.function]]
     name = "read_async"
     manual = true # const ptr
+
+[[object]]
+name = "Gdk.GLTexture"
+status = "generate"
+    [[object.function]]
+    name = "new"
+    manual = true
 
 [[object]]
 name = "Gdk.Monitor"
--- a/gdk4/src/gl_texture.rs
+++ b/gdk4/src/gl_texture.rs
@@ -1,0 +1,44 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::{GLContext, GLTexture, Texture};
+use glib::translate::*;
+use glib::Cast;
+
+impl GLTexture {
+    #[doc(alias = "gdk_gl_texture_new")]
+    pub unsafe fn new(context: &GLContext, id: u32, width: i32, height: i32) -> Self {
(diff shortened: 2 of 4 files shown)


gdk4/Gir.toml
```diff
@@ -355,3 +354,10 @@ manual_traits = ["ToplevelExtManual"]
     [[object.function]]
     name = "show_window_menu"
     manual = true # event is a fundemental type
+
+[[object]]
```
bilelmoussaoui, line 358: there's a test to ensure the objects are ordered alphabetically on the gir file

elmarco: ok, thanks

gdk4/src/gl_texture.rs
```diff
@@ -0,0 +1,44 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::{GLContext, GLTexture, Texture};
+use glib::translate::*;
+
+impl GLTexture {
+    #[allow(clippy::new_ret_no_self)]
+    #[doc(alias = "gdk_gl_texture_new")]
+    pub unsafe fn new(context: &GLContext, id: u32, width: i32, height: i32) -> Texture {
```
sdroege, line 9: Shouldn't we return a `GLTexture` here by casting? Just like `gtk_button_new()` returns a `GtkWidget*` but `gtk::Button::new()` returns a `gtk::Button`.

sdroege (commented):
Looks good to me otherwise

gdk4/src/gl_texture.rs
elmarco: I don't mind, I was respecting the C API. But if in Rust we return the actual type, it's best to be consistent

elmarco: done

sdroege: Because you can directly use the API of all parent classes, that's what we do. In C the only reason why it is like that is because it's boring to write `GTK_WIDGET(button)` everywhere.

gdk4/src/gl_texture.rs
```diff
@@ -0,0 +1,44 @@
...
+            width,
+            height,
+            None,
+            std::ptr::null_mut(),
+        );
+        Texture::from_glib_full(ptr).unsafe_cast()
+    }
+
+    #[doc(alias = "gdk_gl_texture_new")]
+    pub unsafe fn new_with_released<F: FnOnce() + 'static>(
```
sdroege, line 22:
```suggestion
    pub unsafe fn with_release_func<F: FnOnce() + 'static>(
```

Or so?

elmarco: done

sdroege (approved):
@bilelmoussaoui ?

bilelmoussaoui (approved):

bilelmoussaoui:
Thanks!

PR #186, opened 2021-02-14
correctly make RenderNode not a glib::Object

bvinc:
I took a stab at fixing #48.

I made a macro to help the boiler plate of defining a new render node.  But I wanted each render node in its own file, so I ended up exporting the macro, which is unfortunate.  Do you have any other suggestions?  I'm open to getting rid of the macro.

Also, there must be something that I don't understand, because when I run generator.py I get dozens of unrelated files changed.  I hand picked the files that I changed in this PR.


gsk4/src/render_node.rs
```diff
@@ -0,0 +1,171 @@
...
+        #[doc(hidden)]
+        impl FromGlibPtrFull<*mut ffi::GskRenderNode> for $rust_type {
+            unsafe fn from_glib_full(ptr: *mut ffi::GskRenderNode) -> Self {
+                from_glib_full(ptr as *mut $ffi_type)
+            }
+        }
+
+        impl ::std::fmt::Display for $rust_type {
+            fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+                f.write_str($display_name)
```
bilelmoussaoui, line 167: why not just stringify the rust_type?

gsk4/Gir.toml
```diff
@@ -72,86 +70,27 @@ manual = [
...
-[[object]]
-name = "Gsk.ShadowNode"
+name = "Gsk.Renderer"
 status = "generate"
+manual_traits = ["RendererManualExt"]
     [[object.function]]
-    name = "new"
-    # uses &[&Shadow] instead of a &[Shadow]
+    name = "render"
+    # gir
```
bilelmoussaoui, line 89: please make this more descriptive of the reason why it's manual, same for `render_texture`

gsk4/src/render_node.rs
```diff
@@ -0,0 +1,171 @@
...
+
+impl RenderNode {
+    //#[doc(alias = "gsk_render_node_deserialize")]
+    //pub fn deserialize(bytes: &glib::Bytes, error_func: /*Unimplemented*/FnMut(/*Ignored*/ParseLocation, /*Ignored*/ParseLocation, &glib::Error), user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> Option<RenderNode> {
+    //    unsafe { TODO: call ffi:gsk_render_node_deserialize() }
+    //}
+
+    pub fn downcast<T: IsARenderNode>(self) -> Result<T, RenderNode> {
+        unsafe {
+            if self.get_type() == T::static_type() {
```
bilelmoussaoui, line 33: You should probably use get_node_type here, see gdk::Event for an example

gsk4/src/render_node.rs
```diff
@@ -0,0 +1,171 @@
...
+                Some(&*(self as *const RenderNode as *const T))
+            } else {
+                None
+            }
+        }
+    }
+
+    pub fn get_type(&self) -> glib::Type {
+        Self::static_type()
+    }
```
bilelmoussaoui, line 53: This doesn't make much sense, as you should be using something else for downcast/downcast_ref

gsk4/src/render_node.rs
```diff
@@ -0,0 +1,171 @@
...
+
+            match fn {
+                ref => |ptr| ffi::gsk_render_node_ref(ptr as *mut ffi::GskRenderNode) as *mut $ffi_type,
+                unref => |ptr| ffi::gsk_render_node_unref(ptr as *mut ffi::GskRenderNode),
+                get_type => || $get_type(),
+            }
+        }
+
+        unsafe impl Send for $rust_type {}
+        unsafe impl Sync for $rust_type {}
```
bilelmoussaoui, line 128: Are the render nodes actually Send?

bilelmoussaoui:
Thanks for looking at this one!

gsk4/src/render_node.rs
```diff
@@ -0,0 +1,171 @@
...
+            #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+            pub struct $rust_type(Shared<$ffi_type>);
+
+            match fn {
+                ref => |ptr| ffi::gsk_render_node_ref(ptr as *mut ffi::GskRenderNode) as *mut $ffi_type,
+                unref => |ptr| ffi::gsk_render_node_unref(ptr as *mut ffi::GskRenderNode),
+                get_type => || $get_type(),
+            }
+        }
+
```
bilelmoussaoui, line 126: impl StaticType is missing here

gsk4/src/render_node.rs
bvinc: good idea!

gsk4/Gir.toml
bvinc: done

gsk4/src/render_node.rs
bvinc: Yup, I got rid of that method.

gsk4/src/render_node.rs
bvinc: I guess I'm not sure.  I removed it so it should be the same as it was before my change.

gsk4/src/render_node.rs
bvinc: This is implemented for me by specifying the get_type function.

bvinc:
I used an associated constant for each RenderNode type, that's not going to cause a problem like with a minimum rust version, is it?

bilelmoussaoui:
Looks good to me, what do you think @sdroege ?

bilelmoussaoui:
> I used an associated constant for each RenderNode type, that's not going to cause a problem like with a minimum rust version, is it?

That should be fine as we require 1.48 anyway

sdroege:
Also we use associated constants everywhere else already

gsk4/src/render_node.rs
```diff
@@ -0,0 +1,161 @@
...
+        ref => |ptr| ffi::gsk_render_node_ref(ptr),
+        unref => |ptr| ffi::gsk_render_node_unref(ptr),
+        get_type => || ffi::gsk_render_node_type_get_type(),
+    }
+}
+
+impl RenderNode {
+    //#[doc(alias = "gsk_render_node_deserialize")]
+    //pub fn deserialize(bytes: &glib::Bytes, error_func: /*Unimplemented*/FnMut(/*Ignored*/ParseLocation, /*Ignored*/ParseLocation, &glib::Error), user_data: /*Unimplemented*/Option<Fundamental: Pointer>) -> Option<RenderNode> {
+    //    unsafe { TODO: call ffi:gsk_render_node_deserialize() }
```
sdroege, line 23: This looks relatively easy to add, maybe create an issue if you don't want to do it as part of this PR.

gsk4/src/render_node.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    pub fn get_node_type(&self) -> RenderNodeType {
+        unsafe { from_glib(ffi::gsk_render_node_get_node_type(self.to_glib_none().0)) }
+    }
+
+    pub fn serialize(&self) -> Option<glib::Bytes> {
+        unsafe { from_glib_full(ffi::gsk_render_node_serialize(self.to_glib_none().0)) }
+    }
+
+    pub fn write_to_file(&self, filename: &str) -> Result<(), glib::Error> {
```
sdroege, line 68: Should the filename be UTF-8 or should it maybe be a `std::path::Path`? (Also missing annotation in the C code if that's the case)

gsk4/src/render_node.rs
```diff
@@ -0,0 +1,161 @@
...
+            ffi::gsk_render_node_get_bounds(self.to_glib_none().0, bounds.to_glib_none_mut().0);
+            bounds
+        }
+    }
+
+    pub fn get_node_type(&self) -> RenderNodeType {
+        unsafe { from_glib(ffi::gsk_render_node_get_node_type(self.to_glib_none().0)) }
+    }
+
+    pub fn serialize(&self) -> Option<glib::Bytes> {
```
sdroege, line 64: I don't think this can return `None`

gsk4/src/render_node.rs
```diff
@@ -0,0 +1,161 @@
...
+            if error.is_null() {
+                Ok(())
+            } else {
+                Err(from_glib_full(error))
+            }
+        }
+    }
+}
+
+pub trait IsARenderNode:
```
sdroege, line 85: Not a fan of the name. I would've called it `RenderNode` but that exists already. `RenderNodeType`? Not sure :)

gsk4/src/render_node.rs
```diff
@@ -0,0 +1,161 @@
...
+                }
+            }
+        }
+
+        impl crate::render_node::IsARenderNode for $rust_type {
+            const NODE_TYPE: RenderNodeType = $node_type;
+
+            fn upcast(self) -> crate::RenderNode {
+                unsafe {
+                    from_glib_full(self.to_glib_full() as *mut ffi::GskRenderNode)
```
sdroege, line 139: You can also just transmute here

gsk4/src/shadow_node.rs
```diff
@@ -23,4 +29,14 @@ impl ShadowNode {
         assert!(i < self.get_n_shadows());
         unsafe { from_glib_none(ffi::gsk_shadow_node_get_shadow(self.to_glib_none().0, i)) }
     }
+
+    #[doc(alias = "gsk_shadow_node_get_child")]
+    pub fn get_child(&self) -> Option<RenderNode> {
```
sdroege, line 34: Can this actually return `None`?

gsk4/src/radial_gradient_node.rs
```diff
@@ -29,4 +37,53 @@ impl RadialGradientNode {
             ))
         }
     }
+
+    #[doc(alias = "gsk_radial_gradient_node_get_center")]
+    pub fn get_center(&self) -> Option<graphene::Point> {
```
sdroege, line 42: I don't think this can return `None`

gsk4/src/linear_gradient_node.rs
```diff
@@ -24,4 +32,38 @@ impl LinearGradientNode {
...
+        unsafe { from_glib_none(ffi::gsk_linear_gradient_node_get_end(self.to_glib_none().0)) }
+    }
+
+    #[doc(alias = "gsk_linear_gradient_node_get_n_color_stops")]
+    pub fn get_n_color_stops(&self) -> usize {
+        unsafe { ffi::gsk_linear_gradient_node_get_n_color_stops(self.to_glib_none().0) }
+    }
+
+    #[doc(alias = "gsk_linear_gradient_node_get_start")]
+    pub fn get_start(&self) -> Option<graphene::Point> {
```
sdroege, line 62: I don't think this can return `None`

gsk4/src/linear_gradient_node.rs
```diff
@@ -24,4 +32,38 @@ impl LinearGradientNode {
...
+                    n_stops.as_mut_ptr(),
+                ),
+                n_stops.assume_init() as usize,
+            );
+            ret
+        }
+    }
+
+    #[doc(alias = "gsk_linear_gradient_node_get_end")]
+    pub fn get_end(&self) -> Option<graphene::Point> {
```
sdroege, line 52: I don't think this can return `None`

gsk4/src/conic_gradient_node.rs
```diff
@@ -23,4 +31,38 @@ impl ConicGradientNode {
             ))
         }
     }
+
+    #[doc(alias = "gsk_conic_gradient_node_get_center")]
+    pub fn get_center(&self) -> Option<graphene::Point> {
```
sdroege, line 36: I don't think this can return `None`

gsk4/src/border_node.rs
```diff
@@ -1,19 +1,35 @@
...
+    //    unsafe { TODO: call ffi:gsk_border_node_new() }
+    //}
+
+    #[doc(alias = "gsk_border_node_get_colors")]
+    pub fn get_colors(&self) -> Option<gdk::RGBA> {
+        unsafe { from_glib_none(ffi::gsk_border_node_get_colors(self.to_glib_none().0)) }
+    }
+
+    #[doc(alias = "gsk_border_node_get_outline")]
+    pub fn get_outline(&self) -> Option<RoundedRect> {
```
sdroege, line 25: I don't think this can return `None`

gsk4/src/border_node.rs
```diff
@@ -1,19 +1,35 @@
...
+);
+
+impl BorderNode {
+    //#[doc(alias = "gsk_border_node_new")]
+    //pub fn new(outline: &RoundedRect, border_width: /*Unimplemented*/FixedArray TypeId { ns_id: 0, id: 20 }; 4, border_color: /*Unimplemented*/FixedArray TypeId { ns_id: 11, id: 79 }; 4) -> BorderNode {
+    //    unsafe { TODO: call ffi:gsk_border_node_new() }
+    //}
+
+    #[doc(alias = "gsk_border_node_get_colors")]
+    pub fn get_colors(&self) -> Option<gdk::RGBA> {
```
sdroege, line 20: I don't think this can return `None`

gsk4/src/auto/renderer.rs
```diff
@@ -95,30 +84,6 @@ impl<O: IsA<Renderer>> RendererExt for O {
         }
     }
 
-    fn render<P: IsA<RenderNode>>(&self, root: &P, region: Option<&cairo::Region>) {
```
sdroege, line 98: These should probably not all disappear, otherwise the whole of GSK becomes useless.

Can you add a manual implementation for them into `src/renderer.rs` into a `RendererExtManual` trait?

sdroege (commented):
Looks mostly ok to me, thanks!

gsk4/src/render_node.rs
bilelmoussaoui: It's handled by #157 already

gsk4/src/auto/renderer.rs
bilelmoussaoui: Ah good catch. I assumed it was already the case from the `Gir.toml` as it mentions `RendererManualExt` (which is wrong btw)

gsk4/src/render_node.rs
bvinc: This was copied from the generated code.  I'm not sure of the requirements of `gsk_render_node_write_to_file` but std::path::Path is probably preferred.  I'll change it.

gsk4/src/render_node.rs
bvinc: Also copied from generated code.  I'll change it.

gsk4/src/render_node.rs
bvinc: To be honest, I'm not a fan of the name either.  Not only does `RenderNode` exist, but also `RenderNodeType` also exists.  At first, I was using the name `RenderNodeSubType`.  Do you have any other suggestions?

bvinc:
I addressed all of the feedback, except using transmute on the `upcast` method.  I'm not confident enough about the memory layout and reference counting of these objects.  You'll need to show me how the transmute should look.

I added the RendererExtManual trait and included it in the prelude.

I also quit using the `get_type` section of the wrapper macro, since it fulfills some other traits that I don't want.  This is just like GDK Event.

gsk4/src/shadow_node.rs
bvinc: fixed

gsk4/src/linear_gradient_node.rs
bvinc: fixed

bvinc:
I addressed all of the feedback.

I added the RendererExtManual trait and included it in the prelude.

I also quit using the get_type section of the wrapper macro, since it fulfills some other traits that I don't want. This is just like GDK Event.

I think the only issue left to nail down is maybe renaming `IsARenderNode`?

gsk4/src/render_node.rs
bilelmoussaoui: Maybe just `IsRenderNode` ?

gsk4/src/render_node.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    pub fn get_node_type(&self) -> RenderNodeType {
+        unsafe { from_glib(ffi::gsk_render_node_get_node_type(self.to_glib_none().0)) }
+    }
+
+    pub fn serialize(&self) -> glib::Bytes {
+        unsafe { from_glib_full(ffi::gsk_render_node_serialize(self.to_glib_none().0)) }
+    }
+
+    pub fn write_to_file(&self, filename: &Path) -> Result<(), glib::Error> {
```
bilelmoussaoui, line 76:
```suggestion
    pub fn write_to_file<P: AsRef<std::path::Path>(&self, filename: &P) -> Result<(), glib::Error> {
```

gsk4/src/render_node.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub fn serialize(&self) -> glib::Bytes {
+        unsafe { from_glib_full(ffi::gsk_render_node_serialize(self.to_glib_none().0)) }
+    }
+
+    pub fn write_to_file<P: AsRef<Path>>(&self, filename: &P) -> Result<(), glib::Error> {
+        unsafe {
+            let mut error = ptr::null_mut();
+            let _ = ffi::gsk_render_node_write_to_file(
+                self.to_glib_none().0,
+                filename.to_glib_none().0,
```
bilelmoussaoui, line 81:
```suggestion
                filename.as_ref().to_glib_none().0,
```

bvinc: Thanks.  I got it working.

gsk4/src/render_node.rs
sdroege: I mean, `IsRenderNode` sounds better than `IsARenderNode` to me :)

gsk4/src/render_node.rs
```diff
@@ -0,0 +1,173 @@
...
+
+pub const NONE_RENDER_NODE: Option<&RenderNode> = None;
+
+impl fmt::Display for RenderNode {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("RenderNode")
+    }
+}
+
+#[macro_export]
```
sdroege, line 112: We probably don't want to export this outside the crate

gsk4/src/render_node.rs
```diff
@@ -0,0 +1,173 @@
...
+                    &*(self as *const $rust_type as *const crate::RenderNode)
+                }
+            }
+        }
+
+        impl crate::render_node::IsARenderNode for $rust_type {
+            const NODE_TYPE: RenderNodeType = $node_type;
+
+            fn upcast(self) -> crate::RenderNode {
+                unsafe { ::std::mem::transmute(self) }
```
sdroege, line 152: If you prefer you can also transmute in `downcast`, or do the `from_glib_full(to_glib_full)` also here. Let's just keep it consistent in both places.

gsk4/src/border_node.rs
```diff
@@ -1,19 +1,35 @@
...
+define_render_node!(
+    BorderNode,
+    ffi::GskBorderNode,
+    ffi::gsk_border_node_get_type,
+    RenderNodeType::BorderNode
+);
+
+impl BorderNode {
+    //#[doc(alias = "gsk_border_node_new")]
+    //pub fn new(outline: &RoundedRect, border_width: /*Unimplemented*/FixedArray TypeId { ns_id: 0, id: 20 }; 4, border_color: /*Unimplemented*/FixedArray TypeId { ns_id: 11, id: 79 }; 4) -> BorderNode {
```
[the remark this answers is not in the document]
bvinc: Fixed.

gsk4/src/render_node.rs
bvinc: I don't know how to use transmute on downcast, since it returns a T.  I just changed it back to `from_glib_full(to_glib_full)` so it can be just like `ParamSpec`.

gsk4/src/render_node.rs
bvinc: There might be something that I don't understand here, but I couldn't figure out how to make a crate-scoped macro.  It seems like I can either define a macro that's valid for the rest of the file, or I can export it.  Since I really wanted to keep these render nodes in their own separate files, I was forced to export it.

gsk4/src/render_node.rs
sdroege: Ok, all the same really as this is not a performance-critical path :)

gsk4/src/border_node.rs
```diff
@@ -1,19 +1,46 @@
...
+    pub fn new(
+        outline: &RoundedRect,
+        border_width: &[f32; 4],
+        border_color: &[gdk::RGBA; 4],
+    ) -> BorderNode {
+        unsafe {
+            from_glib_full(ffi::gsk_border_node_new(
+                outline.to_glib_none().0,
+                border_width,
+                mem::transmute(border_color),
```
sdroege, line 25:
```suggestion
                border_color.as_ptr() as *const gdk::ffi::GdkRGBA,
```

gsk4/src/border_node.rs
```diff
@@ -1,19 +1,46 @@
...
+    #[doc(alias = "gsk_border_node_new")]
+    pub fn new(
+        outline: &RoundedRect,
+        border_width: &[f32; 4],
+        border_color: &[gdk::RGBA; 4],
+    ) -> BorderNode {
+        unsafe {
+            from_glib_full(ffi::gsk_border_node_new(
+                outline.to_glib_none().0,
+                border_width,
```
sdroege, line 24:
```suggestion
                border_width.as_ptr(),
```

bvinc:
I renamed `IsARenderNode` to `IsRenderNode` and also managed to remove the macro export.

gsk4/src/render_node.rs
bvinc: https://danielkeep.github.io/tlborm/book/mbe-min-scoping.html  Ah, yes, I needed to do that, and also use `#[macro_use]` to allow the macro to escape the module.

bilelmoussaoui (approved):

PR #191, opened 2021-02-19
CI: don't use --all-features for docs

bilelmoussaoui:
because it cancels out the embed-lgpl-docs with purge-lgpl-docs


bilelmoussaoui:
This should half fix the docs issue :)

PR #196, opened 2021-02-27
gtk: make sure gtk::disable_setlocale has to be called before gtk::init()

### LINKED ISSUE #195: gtk::disable_setlocale() shouldn't panic if called before gtk::init()
When calling gtk::disable_setlocale() after gtk::init(), GTK prints the following warning:

```
(dev.Cogitri.Health:140494): Gtk-WARNING **: 19:35:21.489: gtk_disable_setlocale() must be called before gtk_init()
```

However, when calling it before gtk::init() gtk-rs panics with:

```
thread 'main' panicked at 'GTK has not been initialized. Call `gtk::init` first.', /var/home/rasmus/Projects/Jinsei/build/target/cargo-home/git/checkouts/gtk4-rs-e74ad56283dfeb5e/6301405/gtk4/src/auto/functions.rs:129:5
```

Cogitri:
fixes #195

### DIFF
--- a/gtk4/Gir.toml
+++ b/gtk4/Gir.toml
@@ -335,6 +335,9 @@ trust_return_value_nullability = false
     [[object.function]]
     name = "accelerator_valid"
     manual = true # to make use of gdk::keys::Key
+    [[object.function]]
+    name = "disable_setlocale"
+    assertion = "not-initialized"
 
 [[object]]
 name = "Gtk.AboutDialog"
--- a/gtk4/src/auto/functions.rs
+++ b/gtk4/src/auto/functions.rs
@@ -126,7 +126,7 @@ pub fn css_parser_warning_quark() -> glib::Quark {
 
 #[doc(alias = "gtk_disable_setlocale")]
 pub fn disable_setlocale() {
-    assert_initialized_main_thread!();
+    assert_not_initialized!();
     unsafe {
         ffi::gtk_disable_setlocale();
     }


gtk4/Gir.toml
```diff
@@ -335,6 +335,9 @@ trust_return_value_nullability = false
     [[object.function]]
     name = "accelerator_valid"
     manual = true # to make use of gdk::keys::Key
+    [[object.function]]
+    name = "disable_setlocale"
+    manual = true # to use assert_not_initialized
```
bilelmoussaoui, line 340:
```suggestion
    assertions = "skip"
```
should be enough, no need to make it a `manual` function

Cogitri: That'd skip the entire assert though - the manual function adds an assert that it's called before gtk::init

bilelmoussaoui: see https://github.com/gtk-rs/gir/pull/1064

Cogitri: Thanks, pushed a new update

Cogitri:
More cleanup in #198

bilelmoussaoui:
This will currently fail until #197 is merged, see the 2nd commit which contains a fix of the examples.

gtk4/src/auto/functions.rs
```diff
@@ -126,7 +126,7 @@ pub fn css_parser_warning_quark() -> glib::Quark {
 
 #[doc(alias = "gtk_disable_setlocale")]
 pub fn disable_setlocale() {
-    assert_initialized_main_thread!();
+    assert_not_initialized!();
```
sdroege, line 129: Don't you have to add this macro now? I don't think we had that before or did we?

Otherwise looks good to me

sdroege (approved):

gtk4/src/auto/functions.rs
Cogitri:
```
$ rg 'assert_not_initialized'
gtk4/src/rt.rs
33:macro_rules! assert_not_initialized {
```

Seems to live in rt.rs already (and builds fine for me locally), or did you mean something else?

sdroege: No, all good then. Thanks!

bilelmoussaoui:
Do you mind rebasing this one? :-)

Cogitri:
Rebased

PR #201, opened 2021-03-04
gtk: add Expression bindings

### LINKED ISSUE #21: Handle Gtk.*Expression
Needs something similar to 

quoting slomo
>Actually nevermind, it shouldn't have a new() function but basically work like https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/master/gstreamer/src/param_spec.rs#L13 . That way it will work well with the subclass property API.

bilelmoussaoui:
Initial expressions support

Should fix #21, at least the very basic bindings of it :-) 

It's still a WIP so far.


gtk4/src/expression.rs
```diff
@@ -0,0 +1,185 @@
...
+                Some(value)
+            } else {
+                None
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct ExpressionWatch(ffi::GtkExpressionWatch);
```
bilelmoussaoui, line 63: This one is supposed to be an opaque struct per the docs: "An opaque structure representing a watched GtkExpression."

I have no idea how to handle the exposed ref/unref methods in such case. Any suggestions?

gtk4/Gir.toml
```diff
@@ -757,6 +762,11 @@ status = "generate"
     name = "set_sort_func"
     manual = true
 
+[[object]]
+name = "Gtk.Expression"
+status = "manual"
+final_type = true
```
bilelmoussaoui, line 768: This is "provisory". I wonder if we should have a configuration in gir side, where you can define the trait to use instead of `glib::IsA<T>`. So that I can use here `IsExpression` . I think it would be nice & reduce a bunch of manual code. Although, no idea how feasible all of this :-) Any ideas?

gtk4/src/expression.rs
sdroege: In what sense? You can use `glib_wrapper!` here as usual, there is just no `get_type()` function. Can't this even be autogenerated?

gtk4/Gir.toml
sdroege: That would require the information in the .gir that there's a type hierarchy. Does that exist?

gtk4/src/bool_filter.rs
```diff
@@ -0,0 +1,26 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::{BoolFilter, Expression};
+use glib::translate::*;
+
+impl BoolFilter {
+    #[doc(alias = "gtk_bool_filter_new")]
+    pub fn new<E: AsRef<Expression>>(expression: Option<&E>) -> BoolFilter {
```
sdroege, line 8: It's going to be a bit annoying to pass in `None` here, just like with `IsA`. Maybe you want to add a `NONE_EXPRESSION` constant in `expression.rs`

gtk4/src/expression.rs
```diff
@@ -0,0 +1,362 @@
...
+    pub fn bind<T: IsA<Object>>(
+        &self,
+        target: &T,
+        property_name: &str,
+        this: Option<&T>,
+    ) -> ExpressionWatch {
+        assert_initialized_main_thread!();
+        unsafe {
+            let ret = ffi::gtk_expression_bind(
+                self.to_glib_none().0,
```
sdroege, line 47:
```suggestion
                self.to_glib_full(),
```

sdroege: Please also check if there are other `self` parameters that are `transfer full` in this API

gtk4/src/expression.rs
```diff
@@ -0,0 +1,362 @@
...
+        let callback_data: Box_<F> = Box_::new(notify);
+        unsafe {
+            let ret = ffi::gtk_expression_watch(
+                self.to_glib_none().0,
+                this.map(|t| t.as_ref()).to_glib_none().0,
+                Some(notify_trampoline::<F> as _),
+                Box_::into_raw(callback_data) as *mut _,
+                Some(destroy_func::<F> as _),
+            );
+            ExpressionWatch(*ret)
```
sdroege, line 97: This is wrong

> Note that the only reference held to the watch will be released when the watch is unwatched which can happen automatically, and not just via gtk_expression_watch_unwatch(). You should call gtk_expression_watch_ref() if you want to keep the watch around.

You need to use `from_glib_full()` here or otherwise increase the reference count, or otherwise we'll easily have a dangling pointer.

gtk4/src/expression.rs
```diff
@@ -0,0 +1,362 @@
...
+}
+
+impl glib::value::SetValueOptional for Expression {
+    unsafe fn set_value_optional(value: &mut glib::Value, this: Option<&Self>) {
+        ffi::gtk_value_set_expression(value.to_glib_none_mut().0, this.to_glib_none().0)
+    }
+}
+
+/// Register Expression's ParamSpec support
+trait GtkParamSpecExt {
```
sdroege, line 121: There are other pspecs in gtk or not? Maybe move it already to a separate file for that reason :)

gtk4/src/expression.rs
```diff
@@ -0,0 +1,362 @@
...
+    fn expression(name: &str, nick: &str, blurb: &str, flags: glib::ParamFlags) -> Self;
+}
+
+impl GtkParamSpecExt for glib::ParamSpec {
+    #[doc(alias = "gtk_param_spec_expression")]
+    fn expression(name: &str, nick: &str, blurb: &str, flags: glib::ParamFlags) -> Self {
+        skip_assert_initialized!();
+        unsafe {
+            from_glib_full(ffi::gtk_param_spec_expression(
```
sdroege, line 130: Is that really `transfer full`? pspecs have floating references, so maybe this should be actually `transfer none` aka `transfer floating` like for all the other pspecs.

gtk4/src/expression.rs
```diff
@@ -0,0 +1,362 @@
...
+        }
+    }
+
+    #[doc(alias = "gtk_expression_watch_unwatch")]
+    pub fn unwatch(&self) {
+        unsafe { ffi::gtk_expression_watch_unwatch(self.0 as *mut _) }
+    }
+}
+
+#[macro_export]
```
sdroege, line 165: Wouldn't this make the macro visible outside the crate too? We probably don't want that or are there expressions defined outside GTK (or is that possible)?

gtk4/src/expression.rs
```diff
@@ -0,0 +1,362 @@
...
+}
+
+impl glib::StaticType for Expression {
+    #[doc(alias = "gtk_expression_get_type")]
+    fn static_type() -> Type {
+        unsafe { from_glib(ffi::gtk_expression_get_type()) }
+    }
+}
+
+impl Expression {
```
sdroege, line 24: You probably want to add `downcast()` and `downcast_ref()` here too

sdroege: And a `get_type()`

sdroege: And `is()` too maybe

gtk4/src/expression.rs
```diff
@@ -0,0 +1,362 @@
...
+            from_glib_full(ffi::gtk_property_expression_new(
+                this_type.to_glib(),
+                expression.map(|e| e.as_ref()).to_glib_none().0,
+                property_name.to_glib_none().0,
+            ))
+        }
+    }
+
+    #[doc(alias = "gtk_property_expression_new_for_pspec")]
+    pub fn new_for_pspec<E: AsRef<Expression>>(
```
sdroege, line 256:
```suggestion
    pub fn for_pspec<E: AsRef<Expression>>(
```

gtk4/src/expression.rs
```diff
@@ -0,0 +1,362 @@
...
+
+define_expression!(
+    ClosureExpression,
+    ffi::GtkClosureExpression,
+    ffi::gtk_closure_expression_get_type
+);
+
+impl ClosureExpression {
+    #[doc(alias = "gtk_closure_expression_new")]
+    pub fn new(value_type: Type, closure: glib::Closure, params: &[Expression]) -> Self {
```
sdroege, line 322: If you feel like it you can also add a variant here that takes a Rust closure. That would also allow you to get rid of the `Send + Sync` requirement of `glib::Closure` as this is GTK and everything is on the main thread anyway.

gtk4/src/expression.rs
```diff
@@ -0,0 +1,362 @@
...
+);
+
+impl ConstantExpression {
+    #[doc(alias = "gtk_constant_expression_new")]
+    pub fn new<V: ToValue>(value: &V) -> Self {
+        skip_assert_initialized!();
+        unsafe {
+            from_glib_full(ffi::gtk_constant_expression_new(
+                value.to_value_type().to_glib(),
+                [value.to_value().to_glib_none().0],
```
sdroege, line 348: This is wrong. It's a varargs function that collects a `GValue` from the following arguments. You'd call it e.g. like `gtk_constant_expression_new(G_TYPE_INT, 123, NULL)` and that would internally collect the `123` into a `GValue`.

Simply call `gtk_constant_expression_new_for_value()` here too

sdroege:
Otherwise looks good to me. I still don't know how this API is actually used in practice, but at least the bindings should be OK like this and in the spirit of the C API :)

gtk4/src/expression.rs
bilelmoussaoui: > You probably want to add `downcast()` and `downcast_ref()` here too

The thing is I couldn't find a way to get the type or any information that would allow me to check whether the T::static_type() = self's type. That's the reason why I didn't add any of the those functions yet

gtk4/src/expression.rs
bilelmoussaoui: It's the only pspec gtk has, i guess it's fine to keep it here

gtk4/src/expression.rs
bilelmoussaoui: I don't think we want that macro to be visible, no. Removed it

gtk4/src/expression.rs
bilelmoussaoui: >  Can't this even be autogenerated?

aha, Indeed. Although, it miss-generates the evaluate function due to missing annotations. I have kept it manual for now :) thanks!

gtk4/src/expression.rs
bilelmoussaoui: Aha, right. Makes sense actually as the former's a very C api.

gtk4/src/expression.rs
bilelmoussaoui: It was the only one, thanks!

gtk4/src/expression.rs
sdroege: It's a `GTypeInstance` or not?

```cpp
struct _GTypeClass
{
  GType g_type;
};

struct _GTypeInstance
{
  GTypeClass *g_class;
};
```

Cast the pointer and get the type out of it :)

bilelmoussaoui: Nice, thanks!

gtk4/src/expression.rs
bilelmoussaoui: It's not, not sure why I wrote that. Fixed

gtk4/Gir.toml
bilelmoussaoui: Indeed. I will keep the `final_type = true` here for now to not re-write the *Builders manually as otherwise they will use `IsA<Expression>`, using a plain `&Expression` should be fine as the "sub-classes" implement Deref.

gtk4/src/expression.rs
```diff
@@ -0,0 +1,451 @@
...
+            "Some Expression",
+            glib::ParamFlags::CONSTRUCT_ONLY | glib::ParamFlags::READABLE,
+        );
+    }
+
+    #[test]
+    fn test_property_expression() {
+        crate::init().unwrap();
+        let expr =
+            PropertyExpression::new(crate::FontChooser::static_type(), NONE_EXPRESSION, "font");
```
bilelmoussaoui, line 433: It doesn't like the `NONE_EXPRESSION` here at all because the former expect `AsRef<Expression>` and the later `Option<&Expression>` which doesn't work because `Expression` doesn't implement `AsRef<Expression>`.

gtk4/src/expression.rs
```diff
@@ -0,0 +1,451 @@
...
+    #[doc(alias = "gtk_expression_bind")]
+    pub fn bind<T: IsA<Object>>(
+        &self,
+        target: &T,
+        property_name: &str,
+        this: Option<&T>,
+    ) -> ExpressionWatch {
+        assert_initialized_main_thread!();
+        unsafe {
+            from_glib_full(ffi::gtk_expression_bind(
```
sdroege, line 80:
```suggestion
            from_glib_none(ffi::gtk_expression_bind(
```

as commented before

gtk4/src/expression.rs
```diff
@@ -0,0 +1,451 @@
...
+        unsafe extern "C" fn notify_trampoline<F: Fn() + 'static>(user_data: glib::ffi::gpointer) {
+            let f: &F = &*(user_data as *const F);
+            f()
+        }
+        unsafe extern "C" fn destroy_func<F: Fn() + 'static>(user_data: glib::ffi::gpointer) {
+            let _callback: Box_<Option<Box_<F>>> = Box_::from_raw(user_data as *mut _);
+        }
+        let callback_data: Box_<F> = Box_::new(notify);
+        unsafe {
+            from_glib_full(ffi::gtk_expression_watch(
```
sdroege, line 123:
```suggestion
            from_glib_none(ffi::gtk_expression_watch(
```

gtk4/src/expression.rs
sdroege: Well, you don't implement `AsRef<Expression>` for any of the types :) Why not? You only implement `Deref`, so maybe just add `AsRef` in addition

gtk4/src/expression.rs
bilelmoussaoui: glib::Closure has a local variant, I used it here so that we have a simpler way to use closure expression.

gtk4/src/expression.rs
bilelmoussaoui: My bad, I somehow thought I already did.

sdroege (approved):

PR #210, opened 2021-03-09
more manual gtk stuff

bilelmoussaoui:


gtk4/src/auto/widget.rs
```diff
@@ -75,11 +75,8 @@ pub trait WidgetExt: 'static {
     #[doc(alias = "gtk_widget_activate")]
     fn activate(&self) -> bool;
 
-    //#[doc(alias = "gtk_widget_activate_action")]
-    //fn activate_action(&self, name: &str, format_string: &str, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs) -> bool;
-
     #[doc(alias = "gtk_widget_activate_action_variant")]
```
bilelmoussaoui, line 78: hmm, not great; it would have be nice to be able to set another doc_alias from gir I guess.

bilelmoussaoui: https://github.com/gtk-rs/gir/issues/1067

gtk4/src/drop_target.rs
```diff
@@ -2,11 +2,35 @@
...
 use std::mem::transmute;
 
 impl DropTarget {
+    #[doc(alias = "gtk_drop_target_set_gtypes")]
+    pub fn set_gtypes(&self, types: &[Type]) {
+        let types: Vec<glib::ffi::GType> = types.iter().map(|t| t.to_glib()).collect();
+        unsafe {
+            ffi::gtk_drop_target_set_gtypes(
+                self.to_glib_none().0,
+                mut_override(types.as_ptr()),
```
sdroege, line 17: Can't do `to_glib_none()` here?

gtk4/src/gesture_stylus.rs
```diff
@@ -0,0 +1,25 @@
...
+    #[doc(alias = "gtk_gesture_stylus_get_axes")]
+    pub fn get_axes(&self, axes: Vec<AxisUse>) -> Option<Vec<f64>> {
+        let mut values: Vec<f64> = Vec::new();
+        unsafe {
+            let mut axes1: Vec<gdk::ffi::GdkAxisUse> = axes.iter().map(|a| a.to_glib()).collect();
+            axes1.push(gdk::ffi::GDK_AXIS_IGNORE);
+            if from_glib(ffi::gtk_gesture_stylus_get_axes(
+                self.to_glib_none().0,
+                axes1.as_mut_ptr(),
+                values.as_mut_ptr() as *mut *mut f64,
```
sdroege, line 17: This seems wrong. AFAIU this function returns an allocated array of doubles here. So this would be something like

```rust
let mut values = ptr::null_mut();
..._get_axes(..., &mut values);
```

And then you have `axes.len()` values in there.

gtk4/src/drop_target.rs
bilelmoussaoui: Sadly not, we don't have a default impl  for `ToGlibContainerFromSlice` for usize.

gtk4/src/gesture_stylus.rs
```diff
@@ -0,0 +1,25 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::GestureStylus;
+use gdk::AxisUse;
+use glib::translate::*;
+
+impl GestureStylus {
+    #[doc(alias = "gtk_gesture_stylus_get_axes")]
+    pub fn get_axes(&self, axes: Vec<AxisUse>) -> Option<Vec<f64>> {
```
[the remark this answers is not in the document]
bilelmoussaoui: Otherwise it would return `False` and we end up with `None` here from what I understood

gtk4/src/drop_target.rs
sdroege: Oh well!

gtk4/src/gesture_stylus.rs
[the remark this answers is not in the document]
sdroege: Ok, that doesn't seem like a problem then

sdroege (approved):

PR #211, opened 2021-03-10
more manual stuff 2

bilelmoussaoui:


gdk4/src/auto/clipboard.rs
```diff
@@ -249,13 +244,8 @@ impl Clipboard {
...
     }
 
-    //#[doc(alias = "gdk_clipboard_set_valist")]
-    //pub fn set_valist(&self, type_: glib::types::Type, args: /*Unknown conversion*//*Unimplemented*/Unsupported) {
-    //    unsafe { TODO: call ffi:gdk_clipboard_set_valist() }
-    //}
-
     #[doc(alias = "gdk_clipboard_set_value")]
-    pub fn set_value(&self, value: &glib::Value) {
+    pub fn set(&self, value: &glib::Value) {
```
sdroege, line 248: Why not `fn set(&self, value: &dyn ToValue)`?

gdk4/src/display.rs
```diff
@@ -39,6 +39,30 @@ impl Display {
         }
     }
 
+    #[doc(alias = "gdk_display_map_keycode")]
+    pub fn map_keycode(&self, keycode: u32) -> Option<(Vec<KeymapKey>, Vec<Key>)> {
```
sdroege, line 43:
```suggestion
    pub fn map_keycode(&self, keycode: u32) -> Option<Vec<(KeymapKey, Key)>> {
```

Maybe?

gdk4/src/auto/clipboard.rs
sdroege: Maybe the same elsewhere

bilelmoussaoui: Yes, that's much nicer

gdk4/src/display.rs
bilelmoussaoui: Thanks!

gdk4/src/texture.rs
```diff
@@ -0,0 +1,20 @@
...
+use glib::translate::*;
+use glib::IsA;
+
+pub trait TextureExtManual: 'static {
+    #[doc(alias = "gdk_texture_download")]
+    fn download(&self, stride: usize) -> Vec<u8>;
+}
+
+impl<O: IsA<Texture>> TextureExtManual for O {
+    fn download(&self, stride: usize) -> Vec<u8> {
```
bilelmoussaoui, line 13: Not sure if this is nicer than having to pass to a `mut [u8]`?

sdroege: Yes, `&mut [u8]` here and `&[u8]` for upload

PR #212, opened 2021-03-10
WIP: internal gir files

bilelmoussaoui:
depends on https://github.com/gtk-rs/gir/pull/1041


gdk4-wayland/Gir.toml
```diff
@@ -1,5 +1,5 @@
 [options]
-girs_dir = "gir-files"
+girs_directories = ["gir-files", "gtk4-gir-files"]
```
MarijnS95, line 2: [Same as on `gstreamer-rs`] If you're including the directory with `-d` on the cmdline (so that there's no issues with relative paths, these should be `../gir-files` etc _anyway_), these are better removed.

generator.py
```diff
@@ -79,7 +80,7 @@ def regen_crates(path, conf):
     elif path.match("Gir*.toml"):
         print('==> Regenerating "{}"...'.format(path))
 
-        args = [conf.gir_path, "-c", path, "-o", path.parent, "-d", conf.gir_files_path]
+        args = [conf.gir_path, "-c", path, "-o", path.parent, "-d", conf.gir_files_path, "-d", GTK4_GIR_FILES_DIRECTORY]
```
MarijnS95, line 83: That was my initial thought as well, but you could perhaps pick a more general solution from https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/719/diffs?commit_id=d6f94214959ea9e5fd2e631d7ab6069961b97e9c.

Unfortunately the name of this secondary directory is now the first difference between gtk/gtk4/gstreamer `generator.py`, perhaps we should find a cleaner solution (and move this file to a more sensible, _shared_ place like `gir-files` itself). CC @sdroege @GuillaumeGomez.

gdk4-wayland/Gir.toml
bilelmoussaoui: Actually, after giving it a second thought, I think I will keep gtk4's GIR files in gir-files. Because there will be a ton of new libraries (sourceview, libahandy/libadwaita, webkit2gtk4....) that will need the gtk4 gir files somehow. Having to fetch the whole gtk4-rs project for that isn't great :/

MarijnS95: @bilelmoussaoui You can still have a separate repo purely for gir files? I'm asking @sdroege whether we can have a separate `gst-gir-files` repo purely for such files as well.

Either way that changes nothing to this parameter being unncessary, in favour of passing `-d` to `gir` from `generator.py`.

bilelmoussaoui: They are necessary as sometimes I call gir manually with `not_bound` mode to check if there are anything missing. Or gir doesn't read them at all?

MarijnS95: Gir reads them, but `-d` overrides it. Perhaps we can pass `not_bound` mode directly from the generator so that you don't need to call it directly?

In hindsight `generator.py` is getting quite a few responsibilities: updating repos, finding gir files and all that jazz, making it perhaps too complicated while at the same time being circumvented due to lacking functionality like this...

bilelmoussaoui: indeed, especially that we have the same script in 3 repositories at least that we have to keep in sync somehow.

MarijnS95: Correct... Except that now with these changes we might add `-d` for the special gir directories, that would make them different again.

At least we're using `gst-files` from the repo now, that's where I'd move `generator.py`. Then we'd get another `Generator.toml` or sth to configure these paths :see_no_evil:

GuillaumeGomez: Or maybe we shouldn't set default directories in `generator.py` and rely by default on the ones in the `Gir.toml` files. :)

MarijnS95: @GuillaumeGomez That's horrible for things like `gstreamer-gl/sys/egl/` where we need `../../../gst-gir-files` and `../../../gir-files`, which show up like that in every generated file because relative canonicalization isn't implemented in `gir`.

GuillaumeGomez: Each repository can handle it however they want. I think it'd be more appropriate here to remove the usage of `-d` here.

generator.py
sdroege: I would remove the default for `-d` in `generator.py` and rely on the `Gir.toml` files by default. If `-d` is provided it would be for overriding the `Gir.toml`.

bilelmoussaoui:
I'm going to go ahead and close this one as I believe this won't be a good thing for other libraries that depends on gtk4.

PR #213, opened 2021-03-10
more manual stuff

bilelmoussaoui:


gtk4/Gir.toml
```diff
@@ -503,6 +501,9 @@ manual_traits = ["BuilderExtManual"]
     name = "get_object"
     manual = true # add dynamic casting
     doc_trait_name = "BuilderExtManual"
+    [[object.function]]
+    name = "value_from_string"
+    manual = true # ParamSpec is a fundemental type
```
sdroege, line 506: Why does that matter? You can treat it just like a boxed type and that should be autogeneratable or not?

gtk4/src/auto/bool_filter.rs
```diff
@@ -25,7 +25,7 @@ glib::wrapper! {
 
 impl BoolFilter {
     #[doc(alias = "gtk_bool_filter_get_expression")]
-    pub fn get_expression(&self) -> Expression {
+    pub fn get_expression(&self) -> Option<Expression> {
```
sdroege, line 28: This is not nullable

sdroege: but maybe it should be marked as such?

gtk4/src/auto/editable.rs
```diff
@@ -25,13 +25,6 @@ glib::wrapper! {
     }
 }
 
-impl Editable {
-    //#[doc(alias = "gtk_editable_install_properties")]
-    //pub fn install_properties(object_class: /*Ignored*/&mut glib::ObjectClass, first_prop: u32) -> u32 {
```
sdroege, line 30: That's an interesting one that should probably get some special bindings together with `Editable` subclassing support

gtk4/src/auto/media_stream.rs
```diff
@@ -26,9 +26,6 @@ pub trait MediaStreamExt: 'static {
     #[doc(alias = "gtk_media_stream_ended")]
     fn ended(&self);
 
-    //#[doc(alias = "gtk_media_stream_error")]
-    //fn error(&self, domain: glib::Quark, code: i32, format: &str, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs);
```
sdroege, line 30: You probably want to implement the `gtk_media_stream_gerror` variant though. But who is even supposed to call this function?

gtk4/src/auto/tree_model.rs
```diff
@@ -56,7 +53,7 @@ pub trait TreeModelExt: 'static {
     fn get_string_from_iter(&self, iter: &TreeIter) -> Option<glib::GString>;
 
     #[doc(alias = "gtk_tree_model_get_value")]
-    fn get_value(&self, iter: &TreeIter, column: i32) -> glib::Value;
+    fn get(&self, iter: &TreeIter, column: i32) -> glib::Value;
```
sdroege, line 56: Why is this not nullable? If you pass in a wrong column you surely get back `NULL` or something? :)

gtk4/src/cell_area.rs
```diff
@@ -48,6 +61,35 @@ impl<O: IsA<CellArea>> CellAreaExtManual for O {
...
+            );
+            value
+        }
+    }
+
+    fn cell_set<P: IsA<CellRenderer>>(
+        &self,
+        renderer: &P,
+        property_name: &str,
+        value: &dyn ToValue,
```
sdroege, line 81: Do we need to implement property type checking here?

gtk4/src/editable.rs
```diff
@@ -15,33 +15,38 @@ impl Editable {
     #[doc(alias = "gtk_editable_delegate_get_property")]
     pub fn delegate_get_property<P: IsA<Editable> + IsA<glib::Object>>(
```
sdroege, line 16: This is for subclassing only and probably needs some special integration

gtk4/src/media_stream.rs
```diff
@@ -0,0 +1,35 @@
...
+            ffi::gtk_media_stream_error(
+                self.as_ref().to_glib_none().0,
+                domain.to_glib(),
+                code,
+                message.to_glib_none().0,
+            )
+        }
+    }
+
+    fn gerror(&self, error: &glib::Error) {
```
sdroege, line 27:
```suggestion
    fn gerror(&self, error: glib::Error) {
```

gtk4/src/media_stream.rs
```diff
@@ -0,0 +1,35 @@
...
+        unsafe {
+            ffi::gtk_media_stream_gerror(
+                self.as_ref().to_glib_none().0,
+                mut_override(error.to_glib_full()),
```
sdroege, line 31: And do the usual workaround to steal the `GError` and pass it to this function :)

gtk4/src/subclass/editable.rs
```diff
@@ -156,3 +156,15 @@ unsafe extern "C" fn editable_set_selection_bounds<T: EditableImpl>(
         end_position,
     )
 }
+
+pub unsafe trait EditableClassSubclassExt: ClassStruct {
```
sdroege, line 160: This probably needs special subclassing integration

gtk4/src/tree_model_filter.rs
```diff
@@ -18,3 +19,57 @@ impl TreeModelFilter {
...
+            >(
+                model: *mut ffi::GtkTreeModel,
+                iter: *mut ffi::GtkTreeIter,
+                value: *mut glib::gobject_ffi::GValue,
+                column: i32,
+                user_data: glib::ffi::gpointer,
+            ) {
+                let f: &F = &*(user_data as *const F);
+                let ret = f(&from_glib_borrow(model), &from_glib_borrow(iter), column);
+                *value = *ret.to_glib_none().0;
```
sdroege, line 53: This looks wrong memory management-wise. Shouldn't the return value be `transfer full` and owned by the caller? If so, you want `*value = ret.into_raw();`

I guess we also want an `into_raw()` for `GError`...

gtk4/src/tree_row_reference.rs
```diff
@@ -0,0 +1,25 @@
...
+        iter: &TreeIter,
+        new_order: &[i32],
+    ) {
+        assert_initialized_main_thread!();
+        unsafe {
+            ffi::gtk_tree_row_reference_reordered(
+                proxy.as_ref().to_glib_none().0,
+                mut_override(path.to_glib_none().0),
+                mut_override(iter.to_glib_none().0),
+                mut_override(new_order.as_ptr()),
```
sdroege, line 21: Needs some code somewhere to enforce that `new_order.len()` is correct. I don't know what the function is expecting, but it must expect something

gtk4/Gir.toml
bilelmoussaoui: How can we mark ParamSpec as Boxed from gir? i don't think it's possible for now

gtk4/src/auto/bool_filter.rs
bilelmoussaoui: It's not nullable yes, I just fixed the gir warning here and want to tackle the rest of #68 or at least parts of it soon :-) I would leave this for a separate PR

gtk4/src/subclass/editable.rs
bilelmoussaoui: Isn't that enough for install_properties? we probably need another trait that has the delegate_[get|set]_property methods that's auto-implemented for EditableImpl though.

gtk4/Gir.toml
sdroege: That's probably true :)

gtk4/src/auto/bool_filter.rs
sdroege: OTOH you can create a bool filter with a `NULL` expression so... what would this return then?

gtk4/src/subclass/editable.rs
sdroege: Well, try it and let me know :) I'm a bit worried about users having to pass in the property id offset correctly there, and also it has to be called *after* the normal properties are installed.

bilelmoussaoui: That's definitely not great indeed. I will create a custom editable and play around with the subclassing code to figure out something nice to use if possible :-)

sdroege: Maybe simply by checking `<Self as ObjectImpl>::PROPERTIES.len()` but OTOH that's not very composeable then. It would only work if there is one such function like in `GEditable` for adding more properties.

Or you could check with `g_object_class_list_properties()` how many are already installed at this point.

In any case, I think this function should be automatically called from `<Editable as IsImplementable>::interface_init()`

gtk4/Gir.toml
bilelmoussaoui: It's, what would be nice is to make Gir understand the difference between an object and a fundemental and just not use `IsA` for fundemental types.

sdroege: Yes, please create an issue for that

bilelmoussaoui: https://github.com/gtk-rs/gir/issues/1069

gtk4/src/subclass/editable.rs
```diff
@@ -282,6 +362,24 @@ unsafe extern "C" fn editable_get_text<T: EditableImpl>(
...
+    let instance = &*(editable as *mut T::Instance);
+    let imp = instance.get_impl();
+
+    let wrap = from_glib_borrow::<_, Editable>(editable);
+
+    let delegate = imp.get_delegate(wrap.unsafe_cast_ref());
+
+    let delegateptr = delegate.to_glib_full();
+    wrap.set_qdata(*EDITABLE_GET_DELEGATE_QUARK, delegateptr);
+    delegateptr
```
bilelmoussaoui, line 380: This is a transfer none, i believe this the right way to handle it? :)

gtk4/src/subclass/editable.rs
```diff
@@ -354,3 +452,21 @@ unsafe extern "C" fn editable_set_selection_bounds<T: EditableImpl>(
...
+pub unsafe trait EditableClassSubclassExt: ClassStruct {
+    #[doc(alias = "gtk_editable_install_properties")]
+    fn install_properties(&mut self) -> u32 {
+        unsafe {
+            let object_class = self as *mut _ as *mut glib::gobject_ffi::GObjectClass;
+
+            let mut first_prop = std::mem::MaybeUninit::uninit();
+            glib::gobject_ffi::g_object_class_list_properties(object_class, first_prop.as_mut_ptr());
+
+            ffi::gtk_editable_install_properties(object_class, 1)
```
bilelmoussaoui, line 465: This is just temporarily, I tried using the reported list length from `list_properties` but i switched to 1 for testing purposes as it's doesn't work as it should for now

gtk4/src/subclass/editable.rs
```diff
@@ -4,10 +4,68 @@ use super::widget::WidgetImpl;
 use crate::Editable;
 use glib::subclass::prelude::*;
 use glib::translate::*;
-use glib::{Cast, GString};
+use glib::{Cast, GString, ObjectExt, Quark};
 use libc::{c_char, c_int};
+use once_cell::sync::Lazy;
 
 pub trait EditableImpl: WidgetImpl {
+    fn properties(&self) -> &'static [String] {
```
bilelmoussaoui, line 12: This is definitely not very nice, but see the example on how I handled it :) it's probably the only way to make it work without too much involvement from the user.

bilelmoussaoui:
@sdroege I need a bit of help here with the Editable subclassing support. I managed to make it somehow work except the `delegate_[get|set]_property` ones. I'm getting a weird 

```
total props: 8
Setter: enable-undo - ID: 7

(custom_editable:1378583): GLib-GObject-CRITICAL **: 19:15:24.062: g_value_get_float: assertion 'G_VALUE_HOLDS_FLOAT (value)' failed

Getter: enable-undo - ID: 7

(custom_editable:1378583): GLib-GObject-CRITICAL **: 19:15:24.062: g_value_set_float: assertion 'G_VALUE_HOLDS_FLOAT (value)' failed

(custom_editable:1378583): GLib-GObject-WARNING **: 19:15:24.062: ../gobject/gtype.c:4322: type id '0' is invalid

(custom_editable:1378583): GLib-GObject-WARNING **: 19:15:24.062: can't peek value table for type '<invalid>' which is not currently referenced
```

Trying to get one of the properties might cause a crash (which is the case in the example I pushed with `editable.get_enable_undo()`. It needs to be commented for the example to work btw)

One thing I'm wondering about, is the id = the order in which we insert a property? if that's the case, why we would get ID = 7 for `enable-undo` if it's the 8th property that gets added https://gitlab.gnome.org/GNOME/gtk/-/blob/master/gtk/gtkeditable.c#L970-977 ?

What I was thinking of is whether there would be an offset between the ID we received at `ObjectImpl::[get|set]_property` and the  id we should pass to `delegate_[get|set]_property` as the properties do exists, we just fail to get/set the right one per the `g_value_set_float` that fails due to having a not float value. 

Note the code I wrote is heavily based on the the tagged entry demo here https://github.com/GNOME/gtk/blob/master/demos/gtk-demo/demotaggedentry.c

Any ideas?

sdroege:
> @sdroege I need a bit of help here with the Editable subclassing support.

I'll take a look later, I think I know what the problem is. Can you put an example into this PR?

bilelmoussaoui:
> > @sdroege I need a bit of help here with the Editable subclassing support.
> 
> I'll take a look later, I think I know what the problem is. Can you put an example into this PR?

Thanks! Yes, I pushed an example in the latest commit

examples/src/bin/custom_editable.rs
```diff
@@ -0,0 +1,140 @@
...
+            let total_props = klass.install_properties();
+            println!("total props: {}", total_props);
+            klass.set_layout_manager_type::<gtk::BoxLayout>();
+            klass.set_css_name("entry");
+            klass.set_accessible_role(gtk::AccessibleRole::TextBox);
+        }
+    }
+
+    impl ObjectImpl for CustomEditable {
+        fn set_property(
```
sdroege, line 43: Can you also add a property to `CustomEditable` so we know that it also works in combination?

examples/src/bin/custom_editable.rs
```diff
@@ -0,0 +1,140 @@
...
+            println!("Getter: {} - ID: {:#?}",pspec.get_name(), id);
+            if EditableImpl::properties(self).contains(&pspec.get_name().to_string()) {
+                EditableImpl::get_property(self, editable, id, pspec).unwrap()
+            } else {
+                unimplemented!();
+            }
+        }
+
+        fn constructed(&self, editable: &Self::Type) {
+            editable.init_delegate();
```
sdroege, line 73: Why is this and `finish_delegate()` needed? Can't that be automatically done somehow as part of implementing the interface?

gtk4/src/subclass/editable.rs
sdroege: Well, you could also make it a `&'static [&'static str]` as they're all string literals, or go one step further and make it a `&[glib::ParamSpec]` :)

sdroege: You could also add a function that checks if the pspec from set/get_property() is one that is handled by this interface. That might be even easier for the user

sdroege: Also all these 3 things should be in an `EditableImplExt` trait. Nobody should ever override that

gtk4/src/subclass/editable.rs
sdroege: You probably want to ensure that it's always the same pointer too, and you want to make sure the pointer is unreffed when the object is destroyed. So why not store `delegate.clone()`?

gtk4/src/subclass/editable.rs
sdroege: Also you should dynamically get these properties instead of hardcoding, or it will fall apart if a new property is added at some point

sdroege:
See https://github.com/gtk-rs/gtk-rs/pull/368 for the fix.

gtk4/src/subclass/editable.rs
```diff
@@ -4,10 +4,68 @@ use super::widget::WidgetImpl;
...
+
+    #[doc(alias = "gtk_editable_delegate_get_property")]
+    fn get_property(
+        &self,
+        editable: &Self::Type,
+        prop_id: usize,
+        pspec: &glib::ParamSpec,
+    ) -> Option<glib::Value> {
+        unsafe {
+            let mut value = glib::Value::uninitialized();
```
sdroege, line 36: `value` must be initialized to the target property type or it will cause critical warnings.

sdroege:
```suggestion
            let mut value = glib::Value::from_type(pspec.get_value_type());
```

examples/src/bin/custom_editable.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    #[glib::object_subclass]
+    impl ObjectSubclass for CustomEditable {
+        const NAME: &'static str = "CustomEditable";
+        type Type = super::CustomEditable;
+        type ParentType = gtk::Widget;
+        type Interfaces = (gtk::Editable,);
+
+        fn class_init(klass: &mut Self::Class) {
+            let total_props = klass.install_properties();
```
sdroege, line 34: This could be automatically called as part of `<Editable as IsImplementable>::class_init()`

gtk4/src/subclass/editable.rs
bilelmoussaoui: Actually, I did something much nicer :-) I stored the range of the props ids that was added at `install_properties` stage and provided a function to check whether the id is in the range of the added props or not.

examples/src/bin/custom_editable.rs
bilelmoussaoui: Not really, often, you don't want to override all the text handling functions, in such case, you would return a None for `get_delegate` and override all the other functions. But that's not what often people are after, and in such  case you have to call init_delegate/finish_delegate. Which basically just wires up/down the signal machinery from the delegate (your other Editable implementation  like gtk::Text) to your new custom Editable. So we can't really call those by default

examples/src/bin/custom_editable.rs
bilelmoussaoui: Can it actually be done at that stage? we only have `interface_init` and I don't have any idea how to cast `iface` to a `ClassStruct` or just a `GObjectClass`

examples/src/bin/custom_editable.rs
sdroege: We could also add a bool associated constant to the `Impl` struct to automate this. Still easier than having to call a function :) Your call

examples/src/bin/custom_editable.rs
sdroege: Get the class struct for the type implementing the interface: https://gtk-rs.org/gtk-rs/glib/object/struct.Interface.html#method.get_instance_type and https://gtk-rs.org/gtk-rs/glib/object/struct.Class.html#method.from_type

examples/src/bin/custom_editable.rs
```diff
@@ -0,0 +1,376 @@
...
+            nat: &mut i32,
+            min_base: &mut i32,
+            nat_base: &mut i32,
+        ) {
+            let (c_min, c_nat, c_min_base, c_nat_base) =
+                self.container.measure(orientation, for_size);
+            *min = c_min;
+            *nat = c_nat;
+            *min_base = c_min_base;
+            *nat_base = c_nat_base;
```
bilelmoussaoui, line 249: This is definitely not very nice I think; we should probably just return `(i32, i32, i32, i32)` with the subclassing code.

bilelmoussaoui:
I have updated the editable example to something more useful. It's a copy of one gtk's editable demos
image

examples/src/bin/custom_editable.rs
sdroege: I guess so :) But a) do those 4 pointers contain values we might have to read, b) do all 4 always have to be returned?

sdroege:
That looks nice :) Compared to the C version, is it easier/smaller/better or the opposite?

bilelmoussaoui:
> That looks nice :) Compared to the C version, is it easier/smaller/better or the opposite?

370 lines compared to 500 header included, not bad. The largest boilerplate is get_property/set_property/properties.

examples/src/bin/custom_editable.rs
bilelmoussaoui: Ah, that can work, can we override the constructed/dispose functions to call them automatically somehow? if that's the case, we can remove them from the generated bindings and only call them automatically if the new associated bool is set to true.

sdroege: Hmm no, only `instance_init` currently. Not sure how the others could be covered.

sdroege: Let's worry about that in a separate PR I guess

gtk4/src/subclass/editable.rs
```diff
@@ -354,3 +447,31 @@ unsafe extern "C" fn editable_set_selection_bounds<T: EditableImpl>(
...
 }
+
+pub unsafe trait EditableClassSubclassExt: ClassStruct {
+    #[doc(alias = "gtk_editable_install_properties")]
+    fn install_properties(&mut self) {
+        unsafe {
+            let object_class = self as *mut _ as *mut glib::gobject_ffi::GObjectClass;
+
+            let mut first_prop = std::mem::MaybeUninit::uninit();
+            glib::gobject_ffi::g_object_class_list_properties(
```
sdroege, line 458: You need to `g_free()` the return value. Here you leak now instead :)

gtk4/src/subclass/editable.rs
```diff
@@ -40,6 +47,58 @@ pub trait EditableImpl: WidgetImpl {
 }
 
 pub trait EditableImplExt: ObjectSubclass {
+    fn is_delegate_property(&self, id: usize) -> bool {
```
sdroege, line 50: Instead of such things, you can also always just all `set_property()` / `get_property()`. If they return `true` then it was one of the editable properties, otherwise it's a property to be handled still.

The implementations of the functions already do exactly the same check you're doing here.

examples/src/bin/custom_editable.rs
bilelmoussaoui: I will leave this for another time and opened to not forget about it https://github.com/gtk-rs/gtk4-rs/issues/222

examples/src/bin/custom_editable.rs
bilelmoussaoui: I have added a property to show/hide a spinner, like in the gtk demo

gtk4/src/auto/tree_model.rs
bilelmoussaoui: Seems like the C function doesn't return a `gboolean` to tell whether the value was actually set to something or not. It probably needs at least a check whether `column < get_n_column()`

gtk4/src/auto/media_stream.rs
bilelmoussaoui: This is called by custom MediaStream/MediaFile implementations to set the media in error state. This can be retrieved by widgets making use of it to display the error. I suppose, we should only go with the `gerror` variant? Not sure if the `gtk_media_stream_error` variant is any useful.

bilelmoussaoui: I ended up removing the `error` variant & renaming the `gerror` to `error`

gtk4/src/auto/tree_model.rs
sdroege: Does it also need the `GValue` to be initialized with the correct type before passing a mutable reference to the C function?

gtk4/src/auto/media_stream.rs
sdroege: That makes most sense, yes

gtk4/src/tree_row_reference.rs
bilelmoussaoui: It expects the `new_order.len()` to be equal to what `gtk_tree_model_iter_n_children` returns for the passed iter. Added an assertion for that.

gtk4/src/tree_model_filter.rs
bilelmoussaoui: > This looks wrong memory management-wise. Shouldn't the return value be transfer full and owned by the caller? If so, you want *value = ret.into_raw();

The value is annotated as out caller allocates, I have changed it to use `into_raw`.

gtk4/src/auto/tree_model.rs
bilelmoussaoui: From the use cases I found in the code, it only gets passed a `G_VALUE_INIT`

bilelmoussaoui: Added an assertion for `column < get_n_column()`

gtk4/src/auto/bool_filter.rs
bilelmoussaoui: Hmm, indeed. The set_expression would fail if you pass it a `NULL` for the expression, it shouldn't work when you initialize it neither. Opened an upstream issue for now https://gitlab.gnome.org/GNOME/gtk/-/issues/3779

bilelmoussaoui:
Ok, i'm happy with the status of this merge request

gtk4/src/auto/bool_filter.rs
sdroege: I guess the answer is in the issue now. Can you update the PR accordingly?

sdroege: I.e. the expression must be either `NULL` or its value type must be `bool`.

examples/src/bin/custom_editable.rs
```diff
@@ -0,0 +1,434 @@
...
+    use super::*;
+    use glib::subclass::Signal;
+    use glib::ParamSpec;
+    use once_cell::sync::Lazy;
+    use std::cell::{Cell, RefCell};
+
+    pub struct CustomEditable {
+        pub text: gtk::Text,
+        pub spinner: RefCell<Option<gtk::Spinner>>,
+        pub show_spinner: Cell<bool>,
```
sdroege, line 49: Do these fields have to be completely `pub`?

examples/src/bin/custom_editable.rs
```diff
@@ -0,0 +1,434 @@
...
+            } else {
+                match pspec.get_name() {
+                    "show-spinner" => self.show_spinner.get().to_value(),
+                    _ => unimplemented!(),
+                }
+            }
+        }
+
+        fn constructed(&self, editable: &Self::Type) {
+            editable.init_delegate();
```
sdroege, line 124: Some comments about these things would be useful

examples/src/bin/custom_editable.rs
```diff
@@ -0,0 +1,434 @@
...
+    }
+    impl EditableImpl for CustomEditable {
+        fn get_delegate(&self, _editable: &Self::Type) -> Option<gtk::Editable> {
+            Some(self.text.clone().upcast())
+        }
+    }
+
+    #[derive(Debug)]
+    pub struct CustomTag {
+        pub container: gtk::Box,
```
sdroege, line 154: Same question about these

examples/src/bin/custom_editable.rs
```diff
@@ -0,0 +1,434 @@
...
+            SIGNALS.as_ref()
+        }
+
+        fn constructed(&self, tag: &Self::Type) {
+            self.container.set_parent(tag);
+            self.container.append(&self.label);
+
+            let gesture = gtk::GestureClick::new();
+            gesture.connect_released(clone!(@weak tag => move |_gesture, _n_press, _x, _y| {
+                tag.emit(CustomTag::signals()[1].signal_id(), &[]).unwrap();
```
sdroege, line 245: Without a named `const` or something, the `1` here is less understandable than doing `emit_by_name("clicked")`

gtk4/src/cell_area.rs
```diff
@@ -48,6 +61,61 @@ impl<O: IsA<CellArea>> CellAreaExtManual for O {
...
+            let cell_class = glib::Class::<CellArea>::from_type(O::static_type()).unwrap();
+            let pspec: Option<glib::ParamSpec> =
+                from_glib_none(ffi::gtk_cell_area_class_find_cell_property(
+                    cell_class.as_ref() as *const _ as *mut ffi::GtkCellAreaClass,
+                    property_name.to_glib_none().0,
+                ));
+            let pspec = pspec.unwrap_or_else(|| {
+                panic!("The CellArea property {} doesn't exists", property_name)
+            });
+            if pspec.get_value_type() != value.to_value_type() {
```
sdroege, line 102: Not `is_a()`?

gtk4/src/media_stream.rs
```diff
@@ -0,0 +1,22 @@
...
+    #[doc(alias = "gtk_media_stream_error")]
+    fn error(&self, error: glib::Error);
+}
+
+impl<O: IsA<MediaStream>> MediaStreamExtManual for O {
+    fn error(&self, error: glib::Error) {
+        unsafe {
+            ffi::gtk_media_stream_gerror(
+                self.as_ref().to_glib_none().0,
+                mut_override(error.to_glib_none().0),
```
sdroege, line 18: This is `transfer full`

sdroege (approved):
Looks good to me otherwise

examples/src/bin/custom_editable.rs
bilelmoussaoui: Added more details & a bit of explanations

gtk4/src/auto/bool_filter.rs
bilelmoussaoui: I added checks to both BoolFilter & StringFilter :-)

bilelmoussaoui:
Thanks for the review @sdroege. Going to go ahead and merge this for now.

PR #215, opened 2021-03-14
interfaces chain up

bilelmoussaoui:
Details at https://github.com/gtk-rs/gtk-rs/pull/353


bilelmoussaoui:
The build is still failing with 
```

error[E0277]: the trait bound `c_void: std::marker::Copy` is not satisfied
   --> gtk4/src/subclass/app_chooser.rs:9:32
    |
9   | unsafe impl<T: AppChooserImpl> IsImplementable<T> for AppChooser {
    |                                ^^^^^^^^^^^^^^^^^^ the trait `std::marker::Copy` is not implemented for `c_void`
    | 
   ::: /home/bilelmoussaoui/.cargo/git/checkouts/gtk-rs-48ef14c1f17c79fb/a2568f0/glib/src/subclass/types.rs:134:42
    |
134 |     <Self as ObjectType>::GlibClassType: Copy,
    |                                          ---- required by this bound in `glib::subclass::types::IsImplementable`

error[E0277]: the trait bound `c_void: std::marker::Copy` is not satisfied
   --> gtk4/src/subclass/file_chooser.rs:8:33
    |
8   | unsafe impl<T: FileChooserImpl> IsImplementable<T> for FileChooser {
    |                                 ^^^^^^^^^^^^^^^^^^ the trait `std::marker::Copy` is not implemented for `c_void`
    | 
```
@sdroege any ideas? :)

sdroege:
Yeah, both are final types and you can't actually subclass them. The interface struct is not in the headers.

Well, not really final types. Only implementable from inside gtk :) Like `pub(crate)` for the implementation, `pub` for the usage.

bilelmoussaoui:
> Yeah, both are final types and you can't actually subclass them. The interface struct is not in the headers.

I just noticed that after looking inside the Gir file. I will go ahead and remove them. Thanks!

gdk4/src/subclass/paintable.rs
```diff
@@ -7,88 +7,105 @@ use glib::Cast;
 
 pub trait PaintableImpl: ObjectImpl {
     fn get_current_image(&self, paintable: &Self::Type) -> Paintable {
-        unsafe {
-            let type_ = ffi::gdk_paintable_get_type();
-            let iface = glib::gobject_ffi::g_type_default_interface_ref(type_)
-                as *mut ffi::GdkPaintableInterface;
-            assert!(!iface.is_null());
+        self.parent_get_current_image(paintable)
```
sdroege, line 10: Are there default impls provided for all these in C? Only provide default impls in the trait for the functions where that is the case :) Generally for interface `Impl` traits

gtk4/src/subclass/color_chooser.rs
```diff
@@ -13,12 +13,80 @@ pub trait ColorChooserImpl: ObjectImpl {
...
+    ) {
+        self.parent_add_palette(color_chooser, orientation, colors_per_line, colors);
+    }
+
+    fn color_activated(&self, color_chooser: &Self::Type, rgba: RGBA) {
+        self.parent_color_activated(color_chooser, rgba);
+    }
+
     fn get_rgba(&self, color_chooser: &Self::Type) -> RGBA;
     fn set_rgba(&self, color_chooser: &Self::Type, rgba: RGBA);
```
sdroege, line 25: And functions that don't have default impls still should get a function to chain up on the `Ext` trait. The parent class might implement the interface already and the subclass might want to call that.

gdk4/src/subclass/paintable.rs
bilelmoussaoui: Yes, there are defaults for all those functions

sdroege (commented):
Seems fine otherwise apart from those generic comments that probably apply to everything

gtk4/src/subclass/color_chooser.rs
bilelmoussaoui: Ah, right. I somehow missed that :-)

PR #217, opened 2021-03-18
update to 4.1.2

bilelmoussaoui:
Depends on https://github.com/gtk-rs/gir-files/pull/76


gdk4/src/auto/popup_layout.rs
```diff
@@ -6,6 +6,9 @@ use crate::AnchorHints;
 use crate::Gravity;
 use crate::Rectangle;
 use glib::translate::*;
+#[cfg(any(feature = "v4_2", feature = "dox"))]
```
sdroege, line 9: You'll want to set an override in `Cargo.toml` so that `v4_2` does not check for >=4.2 via pkg-config but only for `>=4.1.2`.

gdk4/sys/Cargo.toml
```diff
@@ -48,3 +49,6 @@ features = ["dox"]
 [package.metadata.system-deps.gtk4]
 name = "gtk4"
 version = "4.0.0"
+
+[package.metadata.system-deps.gtk4.v4_2]
+version = "4.2"
```
sdroege, line 54:
```suggestion
version = "4.1.2"
```

gsk4/src/auto/mod.rs
```diff
@@ -19,6 +19,13 @@ mod gl_shader;
 pub use self::gl_shader::GLShader;
 pub use self::gl_shader::GLShaderBuilder;
 
+#[cfg(any(feature = "v4_2", feature = "dox"))]
+#[cfg_attr(feature = "dox", doc(cfg(feature = "v4_2")))]
+mod ngl_renderer;
```
sdroege, line 24: Why is this public API at all?

gsk4/sys/Cargo.toml
```diff
@@ -48,3 +49,6 @@ features = ["dox"]
 [package.metadata.system-deps.gtk4]
 name = "gtk4"
 version = "4.0.0"
+
+[package.metadata.system-deps.gtk4.v4_2]
+version = "4.2"
```
sdroege, line 54:
```suggestion
version = "4.1.2"
```

gsk4/src/auto/mod.rs
bilelmoussaoui: so that you can force a specific renderer I guess?

gtk4/src/auto/file_chooser.rs
```diff
@@ -86,7 +86,7 @@ pub trait FileChooserExt: 'static {
     fn set_current_folder<P: IsA<gio::File>>(&self, file: &P) -> Result<(), glib::Error>;
 
     #[doc(alias = "gtk_file_chooser_set_current_name")]
-    fn set_current_name<P: AsRef<std::path::Path>>(&self, name: P);
+    fn set_current_name(&self, name: &str);
```
sdroege, line 89: Is this change correct?

gtk4/src/auto/flags.rs
```diff
@@ -647,6 +647,68 @@ impl SetValue for PopoverMenuFlags {
     }
 }
 
+bitflags! {
+    pub struct PrintCapabilities: u32 {
```
sdroege, line 651: This probably wants some `Since` markers?

gtk4/src/auto/functions.rs
```diff
@@ -149,10 +150,41 @@ pub fn disable_setlocale() {
 }
 
 //#[doc(alias = "gtk_distribute_natural_allocation")]
-//pub fn distribute_natural_allocation(extra_space: i32, n_requested_sizes: u32, sizes: /*Ignored*/&mut RequestedSize) -> i32 {
+//pub fn distribute_natural_allocation(extra_space: i32, sizes: /*Ignored*/&[&RequestedSize]) -> i32 {
 //    unsafe { TODO: call ffi:gtk_distribute_natural_allocation() }
 //}
 
+#[doc(alias = "gtk_enumerate_printers")]
+pub fn enumerate_printers<P: Fn(&Printer) -> bool + Send + Sync + 'static>(func: P, wait: bool) {
```
sdroege, line 158: Missing `Since`

gtk4/src/auto/im_context.rs
```diff
@@ -183,6 +191,27 @@ impl<O: IsA<IMContext>> IMContextExt for O {
         }
     }
 
+    fn get_surrounding_with_selection(&self) -> Option<(glib::GString, i32, i32)> {
```
sdroege, line 194: Missing `Since`

gtk4/src/auto/im_context.rs
```diff
@@ -219,6 +248,19 @@ impl<O: IsA<IMContext>> IMContextExt for O {
         }
     }
 
+    fn set_surrounding_with_selection(&self, text: &str, cursor_index: i32, anchor_index: i32) {
```
sdroege, line 251: Missing `Since`

gtk4/src/auto/mod.rs
```diff
@@ -701,6 +711,16 @@ pub use self::print_operation_preview::{PrintOperationPreview, NONE_PRINT_OPERAT
 mod print_settings;
 pub use self::print_settings::PrintSettings;
 
+#[cfg(any(target_os = "linux", feature = "dox"))]
+#[cfg_attr(feature = "dox", doc(cfg(target_os = "linux")))]
+mod print_unix_dialog;
+#[cfg(any(target_os = "linux", feature = "dox"))]
+#[cfg_attr(feature = "dox", doc(cfg(target_os = "linux")))]
+pub use self::print_unix_dialog::PrintUnixDialog;
```
sdroege, line 719: Not `v4_2`?

gtk4/sys/src/lib.rs
```diff
@@ -8055,6 +8112,7 @@ extern "C" {
     // GtkOrdering
     //=========================================================================
     pub fn gtk_ordering_get_type() -> GType;
+    pub fn gtk_ordering_from_cmpfunc(cmpfunc_result: c_int) -> GtkOrdering;
```
sdroege, line 8115: Missing `Since`

gtk4/sys/src/lib.rs
```diff
@@ -8348,6 +8406,11 @@ extern "C" {
     //=========================================================================
     pub fn gtk_popover_menu_flags_get_type() -> GType;
 
+    //=========================================================================
+    // GtkPrintCapabilities
+    //=========================================================================
+    pub fn gtk_print_capabilities_get_type() -> GType;
```
sdroege, line 8412: Missing `Since`

gtk4/sys/src/lib.rs
```diff
@@ -8522,6 +8585,7 @@ extern "C" {
     //=========================================================================
     // GtkExpressionWatch
     //=========================================================================
+    pub fn gtk_expression_watch_get_type() -> GType;
```
sdroege, line 8588: Missing `Since`

gtk4/sys/src/lib.rs
```diff
@@ -11895,6 +11959,12 @@ extern "C" {
         text: *mut *mut c_char,
         cursor_index: *mut c_int,
     ) -> gboolean;
+    pub fn gtk_im_context_get_surrounding_with_selection(
```
sdroege, line 11962: Missing `Since`

gtk4/sys/src/lib.rs
```diff
@@ -11907,6 +11977,13 @@ extern "C" {
         len: c_int,
         cursor_index: c_int,
     );
+    pub fn gtk_im_context_set_surrounding_with_selection(
```
sdroege, line 11980: Missing `Since`

gtk4/sys/src/lib.rs
```diff
@@ -13165,6 +13242,29 @@ extern "C" {
         group_name: *const c_char,
     );
 
+    //=========================================================================
+    // GtkPageSetupUnixDialog
+    //=========================================================================
+    pub fn gtk_page_setup_unix_dialog_get_type() -> GType;
```
sdroege, line 13248: Missing `Since`

gtk4/sys/src/lib.rs
```diff
@@ -13348,6 +13448,70 @@ extern "C" {
         dpi_y: c_double,
     );
 
+    //=========================================================================
+    // GtkPrintJob
+    //=========================================================================
+    pub fn gtk_print_job_get_type() -> GType;
```
sdroege, line 13454: Missing `Since`

gtk4/sys/src/lib.rs
```diff
@@ -13629,6 +13793,111 @@ extern "C" {
     );
     pub fn gtk_print_settings_unset(settings: *mut GtkPrintSettings, key: *const c_char);
 
+    //=========================================================================
+    // GtkPrintUnixDialog
+    //=========================================================================
+    pub fn gtk_print_unix_dialog_get_type() -> GType;
```
sdroege, line 13799: Missing `Since`

gtk4/sys/src/lib.rs
```diff
@@ -13629,6 +13793,111 @@ extern "C" {
...
+    );
+    pub fn gtk_print_unix_dialog_set_support_selection(
+        dialog: *mut GtkPrintUnixDialog,
+        support_selection: gboolean,
+    );
+
+    //=========================================================================
+    // GtkPrinter
+    //=========================================================================
+    pub fn gtk_printer_get_type() -> GType;
```
sdroege, line 13859: Missing `Since`

gtk4/sys/src/lib.rs
```diff
@@ -17233,6 +17513,12 @@ extern "C" {
         n_requested_sizes: c_uint,
         sizes: *mut GtkRequestedSize,
     ) -> c_int;
+    pub fn gtk_enumerate_printers(
```
sdroege, line 17516: Missing `Since`

sdroege (commented):
Otherwise seems fine. Add those `Since` markers to the `Gir.toml` and also to the C code :)

gtk4/src/auto/flags.rs
bilelmoussaoui: They were always there, just not exposed to the Gir file, do they really need one in such case?

gsk4/src/auto/mod.rs
sdroege: Ok :)

gtk4/src/auto/flags.rs
sdroege: Were they exported by the library in 4.0? Then this is OK

bilelmoussaoui: Yes, they were exported at least in the C api since v3 at least. It just never made it way to the gir files, the bindings neither.

sdroege: Ah nice, then everything is less bad than it appeared first :)

gtk4/sys/src/lib.rs
bilelmoussaoui: This one was always there, just wasn't inside `#ifdef __GI_SCANNER__`. I think we don't need it here, we already have From/Into GtkOrdering to Rust's ordering enum. Maybe could be used inside a From<i32> implementation.

gtk4/src/auto/im_context.rs
bilelmoussaoui: Fixed upstream https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/3315, let's wait till the next patch release if there would be any

gtk4/src/auto/im_context.rs
bilelmoussaoui: Same

gtk4/src/auto/file_chooser.rs
bilelmoussaoui: Yes, it takes a file name, not a file path.

gtk4/sys/src/lib.rs
bilelmoussaoui: I don't think there's a way to mark the `get_type` function with a Since annotation, the type existed before. It's already guarded by the `GDK_AVAILABLE_IN_4_2` macro, but it doesn't seem to help, see https://gitlab.gnome.org/GNOME/gtk/-/blob/master/gtk/gtkexpression.h#L74

gtk4/src/auto/im_context.rs
bilelmoussaoui: Actually, as it's small enough of a change, I patched it here as well :)

gtk4/sys/src/lib.rs
sdroege: Yes, that can only be marked as such in the `Gir.toml`. There's no way to annotate it accordingly otherwise.

This is only needed if the `get_type()` function is newer than the type itself, like here.

gtk4/src/expression.rs
```diff
@@ -210,6 +210,13 @@ impl ExpressionWatch {
     }
 }
 
+impl glib::StaticType for ExpressionWatch {
```
sdroege, line 213: This needs a `v4_2` here too, just like in the sys crate

gtk4/Gir.toml
```diff
@@ -1513,6 +1512,13 @@ generate_builder = true
     name = "set_expression"
     manual = true # use AsRef<Expression>
 
+[[object]]
+name = "Gtk.Ordering"
+status = "generate"
+    [[object.function]]
+    name = "from_cmpfunc"
+    ignore = true # not useful for Rust, conversion From/Into cmp::Ordering is possible
```
sdroege, line 1520: Also ignore it in the `sys` crate.

The problem is that otherwise things won't like with MSVC for people using GTK 4.0, even if the functions in question are not actually used!

gtk4/src/subclass/check_button.rs
```diff
@@ -11,10 +11,15 @@ pub trait CheckButtonImpl: CheckButtonImplExt + WidgetImpl {
     fn toggled(&self, check_button: &Self::Type) {
         self.parent_toggled(check_button)
     }
+
+    fn activate(&self, check_button: &Self::Type) {
```
bilelmoussaoui, line 15: This one is problematic probably no? Not sure if we can do much about it.

sdroege: Problematic in what sense?

bilelmoussaoui: it's a new virtual method that got added in v4_2. It's already marked with `Since` but that doesn't show up in `GtkCheckButtonClass`

sdroege: That's ok, you'd put the `#[cfg]` yourself into the `subclass/check_button.rs`. It doesn't matter for -sys here.

PR #218, opened 2021-03-18
Update gir (Gir.toml format update for gir directories)

GuillaumeGomez:
From https://github.com/gtk-rs/gir/pull/1041.

I still wonder if we should keep the default value for `--gir-files-directory` in the `generator.py`...


MarijnS95:
I guess this supersedes #212 in some way? That PR also separates out the files (which should be done separately for review anyway) and has review suggesting to use `-d` instead of `girs_directories` for less clutter.

bilelmoussaoui:
Oups, i completely missed this one. Do you mind rebasing ?

GuillaumeGomez:
I'll fully re-make it. It'll be simpler. ;)

GuillaumeGomez:
Updated.

bilelmoussaoui:
Thanks!

PR #221, opened 2021-03-19
dockerfile: bump to gtk 4.1.2

bilelmoussaoui:
Needed for #217


bilelmoussaoui:
Oh this is bad. The image gets updated on pull request event as well and not on push only.

bilelmoussaoui:
As #217 is ready, I will go ahead and merge this

PR #225, opened 2021-03-22
gdk: unmark *_async callbacks as Send

### LINKED ISSUE #202: Non-future, async gdk4::Clipboard functions require Send-type callbacks
The following functions demand a callback function which includes `Send` in its type parameters:

- `gdk4::Clipboard::read_async`
- `gdk4::Clipboard::read_texture_async`
- `gdk4::Clipboard::read_text_async`
- `gdk4::Clipboard::read_value_async`
- `gdk4::Clipboard::store_async`

This seems to be redundant, and can prevent their usage with thread-unsafe variables.

bilelmoussaoui:
Fixes #202


gdk4/Gir.toml
```diff
@@ -181,9 +181,8 @@ generate_builder = true
     name = "set"
     manual = true # use ToValue
     [[object.function]]
-    name = "read_async"
-    manual = true # array
-
+    pattern = "(read|read_texture|read_text|read_value|store)_async"
+    manual = true # don't enforce Send
```
sdroege, line 185: `concurrency = "none"` or not?

bilelmoussaoui: Can only be set at object configuration level. also I believe the Send requirement comes from the _async_future variant somehow.

sdroege: Oh yes, can only be set global, per-object or for signals. Not for functions. That seems like something worth changing.

> also I believe the Send requirement comes from the _async_future variant somehow.

In what sense?

bilelmoussaoui: > also I believe the Send requirement comes from the _async_future variant somehow.
Because the `_async` variant callback is called in the `async_future` one and there's no way to enforce `Send` there no?

bilelmoussaoui: > Oh yes, can only be set global, per-object or for signals. Not for functions. That seems like something worth changing.

Opened an issue to handle that on gir side https://github.com/gtk-rs/gir/issues/1078

sdroege: > Because the _async variant callback is called in the async_future one and there's no way to enforce Send there no?

Why would the async_future variant need `Send` though?

bilelmoussaoui: > Why would the async_future variant need Send though?

Because the future might be spawned on a different thread? I don't know. If it's not needed, gir should probably not automatically add the `Send` condition here then.

sdroege: I think that would panic already anyway? Can you make a small example that breaks?

bilelmoussaoui:
I have rebased this one, i will add an example that uses the clipabord first so we have a way to test that it works

bilelmoussaoui:
It seems everything works as expected actually

sdroege (approved):

PR #229, opened 2021-03-24
gtk: more manual stuff 4

bilelmoussaoui:
Pretty small one

- FileChooser::add_choice
- PrintSettings::set_page_ranges
- EntryBuffer::inserted-text
- EntryBuffer::deleted-text

### DIFF
--- a/examples/custom_entry_buffer/entry_buffer/imp.rs
+++ b/examples/custom_entry_buffer/entry_buffer/imp.rs
@@ -1,0 +1,128 @@
+use glib::{ParamFlags, ParamSpec, Value};
+use gtk::glib;
+use gtk::prelude::*;
+use gtk::subclass::prelude::*;
+use once_cell::sync::Lazy;
+use std::cell::{Cell, RefCell};
+use std::rc::Rc;
+
+#[derive(Debug)]
+pub struct CustomEntryBuffer {
+    text: Rc<RefCell<String>>,
+    length: Cell<u16>,
+    max_length: Cell<u16>,
+}
+
+impl Default for CustomEntryBuffer {
+    fn default() -> Self {
+        Self {
+            text: Rc::new(RefCell::new(String::new())),
+            length: Cell::new(0),
+            max_length: Cell::new(0),
+        }
+    }
+}
+
+#[glib::object_subclass]
+impl ObjectSubclass for CustomEntryBuffer {
+    const NAME: &'static str = "CustomEntryBuffer";
+    type Type = super::CustomEntryBuffer;
+    type ParentType = gtk::EntryBuffer;
+}
+
+impl ObjectImpl for CustomEntryBuffer {
+    fn properties() -> &'static [ParamSpec] {
+        static PROPERTIES: Lazy<Vec<ParamSpec>> = Lazy::new(|| {
+            vec![
+                ParamSpec::new_uint(
+                    "length",
+                    "length",
+                    "Length",
+                    0,
+                    u16::MAX as u32,
+                    0,
+                    ParamFlags::READABLE,
+                ),
+                ParamSpec::new_uint(
+                    "max-length",
+                    "Maximum length",
+                    "Maximum number of characters for this entry",
+                    0,
+                    u16::MAX as u32,
+                    0,
+                    ParamFlags::READWRITE | ParamFlags::EXPLICIT_NOTIFY,
+                ),
+                ParamSpec::new_string(
+                    "text",
+                    "Text",
+                    "The contents of the buffer",
+                    Some(""),
+                    ParamFlags::READWRITE,
+                ),
+            ]
+        });
+        PROPERTIES.as_ref()
+    }
+
+    fn property(&self, _obj: &Self::Type, _id: usize, pspec: &ParamSpec) -> Value {
+        match pspec.name() {
+            "length" => (self.length.get() as u32).to_value(),
+            "max-length" => (self.max_length.get() as u32).to_value(),
+            "text" => self.text.borrow().to_value(),
+            _ => unreachable!(),
+        }
+    }
+
+    fn set_property(&self, _obj: &Self::Type, _id: usize, value: &Value, pspec: &ParamSpec) {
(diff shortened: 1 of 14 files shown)


gtk4/src/file_chooser.rs
```diff
@@ -0,0 +1,28 @@
...
+use glib::translate::*;
+use glib::IsA;
+
+pub trait FileChooserExtManual: 'static {
+    #[doc(alias = "gtk_file_chooser_add_choice")]
+    fn add_choice(&self, id: &str, label: &str, options: &[(Option<&str>, Option<&str>)]);
+}
+
+impl<O: IsA<FileChooser>> FileChooserExtManual for O {
+    fn add_choice(&self, id: &str, label: &str, options: &[(Option<&str>, Option<&str>)]) {
```
sdroege, line 13: I don't think these can be `None`? But you can pass `NULL` to both pointers.

gtk4/src/entry_buffer.rs
```diff
@@ -125,4 +135,67 @@ impl<O: IsA<EntryBuffer>> EntryBufferExtManual for O {
...
+            f: glib::ffi::gpointer,
+        ) where
+            P: IsA<EntryBuffer>,
+        {
+            let f: &F = &*(f as *const F);
+            f(
+                &EntryBuffer::from_glib_borrow(this).unsafe_cast_ref(),
+                to_u16!(position),
+                &glib::GString::from_glib_borrow(chars),
+                to_u16!(n_chars),
```
sdroege, line 186: Should `chars` be limited to `n_chars`?

gtk4/src/file_chooser.rs
bilelmoussaoui: They can actually both be None. At least per the docs. The first item of the tuple is the option identifier, if it's set to None, the option is set as a boolean (checkbox) instead of a menu. 

I will give it a try in an example and see if everything works as expected :)

sdroege: That's not how I understand it. The whole thing can be `NULL`, or you have two string arrays of the same length. The string arrays are `NULL` terminated, so you can't have `NULL`s in the middle of them.

gtk4/src/entry_buffer.rs
bilelmoussaoui: As this needs a bit of testing, I added an example and i will try to use some of those APIs and see how that goes

examples/custom_entry_buffer/entry_buffer/imp.rs
```diff
@@ -0,0 +1,128 @@
...
+                self.text.replace(value.get().unwrap());
+            }
+            _ => unreachable!(),
+        }
+    }
+}
+
+impl EntryBufferImpl for CustomEntryBuffer {
+    fn text(&self, _entry_buffer: &Self::Type) -> glib::GString {
+        self.text.borrow().clone().into()
```
bilelmoussaoui, line 94: This is all not really great to use, is there a better way to handle this than having to clone the string? or is it fine to do so as the string can't be that huge due to the EntryBuffer limitation of being at max of a `u16::MAX` length?

examples/custom_entry_buffer/entry_buffer/imp.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    fn length(&self, _entry_buffer: &Self::Type) -> u16 {
+        self.text.borrow().chars().count() as u16
+    }
+
+    fn insert_text(&self, entry_buffer: &Self::Type, _position: u16, chars: &str) -> u16 {
```
bilelmoussaoui, line 101: Both the current insert_text & delete_text methods are wrong due to their complexity on figuring out the bytes position back from the chars position where to insert/delete the string. I was wondering if there might some kind of API/external crate that helps for such situations?

gtk4/src/auto/entry_buffer.rs
```diff
@@ -75,27 +69,6 @@ pub trait EntryBufferExt: 'static {
 }
 
 impl<O: IsA<EntryBuffer>> EntryBufferExt for O {
-    fn emit_deleted_text(&self, position: u32, n_chars: u32) {
```
bilelmoussaoui, line 78: I moved those to manual as well to make use of u16 so that the api is consistent

gtk4/src/subclass/entry_buffer.rs
```diff
@@ -4,31 +4,23 @@ use crate::subclass::prelude::*;
...
 use once_cell::sync::Lazy;
 
 pub trait EntryBufferImpl: EntryBufferImplExt + ObjectImpl {
-    fn delete_text(&self, entry_buffer: &Self::Type, position: u32, n_chars: Option<u32>) -> u32 {
-        self.parent_delete_text(entry_buffer, position, n_chars)
-    }
+    fn delete_text(&self, entry_buffer: &Self::Type, position: u16, n_chars: Option<u16>) -> u16;
 
-    fn deleted_text(&self, entry_buffer: &Self::Type, position: u32, n_chars: Option<u32>) {
+    fn deleted_text(&self, entry_buffer: &Self::Type, position: u16, n_chars: Option<u16>) {
```
bilelmoussaoui, line 13: Those methods are required and don't have default implementations

examples/custom_entry_buffer/entry_buffer/imp.rs
sdroege: I don't know. You could in theory just return a reference here if it isn't required to be NUL-terminated.

examples/custom_entry_buffer/entry_buffer/imp.rs
sdroege: The `std` API is not enough? How do actual implementors of this in C handle it?

bilelmoussaoui:
As much as i would like to get this one in before the release. I will definitely not have the time to finish it. So worth moving it to the next milestone

sdroege:
You could also get it into 0.1.1 or so :) It's new API, not something requiring API changes.

Issue #232, opened 2021-03-25
Compile application with gtk4-rs on MSYS2 Windows


mmMike:
Hi, 

I'm trying to compile the minimal example (basic.rs) on Windows. 
I have a clean MSYS2 setup and Rust also via MSYS2. The packages for Freetype, Gtk3 and Gtk4 are installed.

The Gtk3 version of basic.rs works as expected. 

The Gtk4 version gives the following problems, all but one releated to printing:

```
...
   Compiling gtk4 v0.1.0 (C:\Projekte\gtk4)
error: linking with `x86_64-w64-mingw32-gcc` failed: exit code: 1
  |
[...]

PR #233, opened 2021-03-25
gtk: disable more printer stuff on non-unix platforms

bilelmoussaoui:
Part of #232


gsk4/src/text_node.rs
```diff
@@ -64,6 +64,8 @@ impl TextNode {
     }
 
     #[doc(alias = "gsk_text_node_has_color_glyphs")]
+    #[cfg(any(feature = "v4_2", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v4_2")))]
```
sdroege, line 68: This should probably get an explanatory comment

sdroege (approved):

PR #237, opened 2021-03-26
generator: Iterate gir directories passed on the command line

MarijnS95:
Picked from gtk-rs, as requested in https://github.com/gtk-rs/gtk-rs/pull/385#issuecomment-808199013.

---

This change cleans up handling of the default gir directory, and makes sure the `-d` flag is prepended to every path. Before the command would contain a single `-d` flag followed by a `list` of paths, which cannot be passed to an executable verbatim:

    The following error occurred: expected str, bytes or os.PathLike object, not list
    Do you want to continue? [y/N] ^CTraceback (most recent call last):
      File "./generator.py", line 189, in <module>

Fixes: c62171ab5 ("Don't set a default value for the -d option")

### DIFF
--- a/generator.py
+++ b/generator.py
@@ -79,9 +79,9 @@ def regen_crates(path, conf):
     elif path.match("Gir*.toml"):
         print('==> Regenerating "{}"...'.format(path))
 
-        args = [conf.gir_path, "-c", path, "-o", path.parent]
-        if len(conf.gir_files_paths) > 0:
-            args += ["-d", conf.gir_files_paths]
+        args = [conf.gir_path, "-c", path, "-o", path.parent] + [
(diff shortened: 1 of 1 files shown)


generator.py
```diff
@@ -135,8 +135,8 @@ def parse_args():
         help="Paths in which to look for Gir.toml files",
     )
     parser.add_argument(
-        "--gir-files-directory",
```
MarijnS95, line 138: @GuillaumeGomez Apparently this was not changed as part of the gtk4-rs commit, only gtk-rs. Fwiw I use `git format-patch -1` and `git am` to move patches between repos :)

GuillaumeGomez (approved):

PR #245, opened 2021-03-27
gtk: add features to re-export gdk-x11 & gdk-wayland

### LINKED ISSUE #243: Expose `gdk-wayland` & `gdk-x11` with a feature flag
It would be nice to be able to include both gdkx11 & gdkwayland directly into the re-exported crates from gtk if those features are enabled. 

Of course, the features shouldn't be set by default. 

It's mostly to avoid having to 
```toml
gdk4wayland = {package = "gdk4-wayland", git = "https://github.com/gtk-rs/gtk4-rs", optional = true}
gdk4x11 = {package = "gdk4-x11", git = "https://github.com/gtk-rs/gtk4-rs", optional = true}
gtk4 = {git = "https://github.com/gtk-rs/gtk4-rs", optional = true}
```

and instead it should probably be
```toml
gtk4 = {git = "https://github.com/gtk-rs/gtk4-rs", optional = true, features = ["gdkx11", "gdkwayland"]}
```

bilelmoussaoui:
fixes #243


sdroege:
Seems OK to me except for the concern I wrote in the issue

PR #247, opened 2021-03-28
docs: make use of the dox feature

bilelmoussaoui:
otherwise, all the docs are displayed without a note on which feature has to be enabled
for an item.

### DIFF
--- a/gtk4/Cargo.toml
+++ b/gtk4/Cargo.toml
@@ -22,7 +22,7 @@ name = "gtk4"
 [features]
 purge-lgpl-docs = ["gtk-rs-lgpl-docs", "gdk/purge-lgpl-docs", "gsk/purge-lgpl-docs"]
 embed-lgpl-docs = ["gtk-rs-lgpl-docs", "gdk/embed-lgpl-docs", "gsk/embed-lgpl-docs"]
-dox = ["gdk/dox", "gsk/dox", "ffi/dox"]
+dox = ["gdk/dox", "gsk/dox", "ffi/dox", "x11", "wayland"]
 x11 = ["gdk-x11"]
 wayland = ["gdk-wayland"]
 v4_2 = ["ffi/v4_2", "gdk/v4_2", "gsk/v4_2"]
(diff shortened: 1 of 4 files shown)


gtk4/Cargo.toml
```diff
@@ -22,7 +22,7 @@ name = "gtk4"
 [features]
 purge-lgpl-docs = ["gtk-rs-lgpl-docs", "gdk/purge-lgpl-docs", "gsk/purge-lgpl-docs"]
 embed-lgpl-docs = ["gtk-rs-lgpl-docs", "gdk/embed-lgpl-docs", "gsk/embed-lgpl-docs"]
-dox = ["gdk/dox", "gsk/dox", "ffi/dox"]
+dox = ["gdk/dox", "gsk/dox", "ffi/dox", "x11", "wayland"]
```
bilelmoussaoui, line 25: I'm not sure whether this is right to do. But then we don't have any other way to do it, if we want to show the possible feature there's.

sdroege: Will this still build on a system without x11/Wayland then? Probably not?

GuillaumeGomez: Normally we don't link when the `dox` feature is enabled, so it should be fine. Is the CI checking if `cargo doc` works with the `dox` feature? If so we should be fine.

bilelmoussaoui: > Will this still build on a system without x11/Wayland then? Probably not?

If you're building docs locally? I can't try on those platforms, I suppose it should. Maybe we can add another cfg to only enable them under linux as those apis are not useful for macOS/Windows.

> Is the CI checking if cargo doc works with the dox feature? If so we should be fine.

Yes, the CI does build the docs, it just only publish them if the job was running on the master branch.

bilelmoussaoui: I have added an extra cfg to ensure it's only set for linux. Although, it seems rustdoc doesn't display the configured feature attribute for the re-exports. 

Anyway, this seems to be working fine locally

GuillaumeGomez:
It's because you need to add `doc(cfg())`. :)

PR #250, opened 2021-03-30
CI: properly link to external docs

### LINKED ISSUE #249: Resolve links to other gtk-rs crates in docs
This is a known problem, but not documented yet as far as I can see.
If we for example look at the implemented traits of Button, the link to `glib::object::ObjectExt` is broken.
https://gtk-rs.org/gtk4-rs/gtk4/struct.Button.html#implements

bilelmoussaoui:
Half fix #249


bilelmoussaoui:
The other missing half is the `Implements` doc section generated by gir? which contains a broken link to glib::Object::ObjectExt.

PR #251, opened 2021-04-02
regen with intra-doc links 1/3

bilelmoussaoui:


gdk4-wayland/src/auto/versions.txt
```diff
@@ -1,2 +1,2 @@
-Generated by gir (https://github.com/gtk-rs/gir @ 56b95ac)
-from gir-files (https://github.com/gtk-rs/gir-files @ 6088bb6)
+Generated by gir (https://github.com/gtk-rs/gir @ 4e4ba32)
+from gir-files (@ 6088bb6)
```
sdroege, line 2: Seems like this doesn't work as it should @MarijnS95 ?

sdroege:
Apart from that doesn't actually change anything with regards to docs :)

gdk4-wayland/src/auto/versions.txt
MarijnS95: Strange indeed, it works fine for GStreamer... Let me check `.gitmodules` here...

MarijnS95:
@sdroege Indeed, this doesn't change anything for the docs themselves. I have a change in progress to teach `generator.py` to invoke `gir` for that (for GStreamer-rs), but gtk(4)-rs seem to do it in the CI. Let's move to IRC and discuss the best way forward :)

gdk4-wayland/src/auto/versions.txt
MarijnS95: Should be fixed with https://github.com/gtk-rs/gir/pull/1090.

bilelmoussaoui: Seems to be working now! Thanks @MarijnS95 :D

MarijnS95: Yup, looks like we didn't strip `.git` though. As with all things I keep repeating this one sentence: "GStreamer-rs doesn't have it, didn't notice" :grimacing: 

(Though over at `GStreamer-rs`  we _do_ have `.git` for the fd.o-hosted `gst-gir-files` repo because GitLab prints annoying redirects)

Anyway, this regen doesn't make any difference because `docs/` aren't checked in; maybe you want to wait with the comment-regeneration noise until something interesting is merged to `gir`?

PR #253, opened 2021-04-04
CI: add a book build-deploy job

bilelmoussaoui:


.github/workflows/book.yml
```diff
@@ -0,0 +1,65 @@
...
+          override: true
+
+      - run: cargo build
+        working-directory: book/listings
+
+  build-deploy:
+    runs-on: ubuntu-20.04
+    name: build
+    container:
+      image: ghcr.io/gtk-rs/gtk4-rs/gtk4:latest
```
Hofer-Julian, line 32: Why do I need gtk4 for deploying?

.github/workflows/book.yml
```diff
@@ -0,0 +1,65 @@
...
+          destination_dir: book/git
+
+  codespell:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: codespell-project/actions-codespell@master
+        with:
+          check_filenames: true
+          path: book
```
Hofer-Julian, line 65: I guess it wouldn't hurt to have codespell for the whole project?

bilelmoussaoui: Most of the code is auto-generated and will need a bunch of fixes which are unrelated to the book, maybe in the future

.github/workflows/book.yml
bilelmoussaoui: Good point, removed

bilelmoussaoui:
Let's give this a try and see if it works as expected o/

PR #265, opened 2021-04-04
Finish up prerequisites

### LINKED ISSUE #257: Settings: Link to a meson example while mentioning it
See https://gitlab.gnome.org/Hofer-Julian/gtk-rs-book/-/issues/16

Hofer-Julian:
Fixes #257


.github/workflows/book.yml
```diff
@@ -38,6 +38,8 @@ jobs:
 
       - name: Setup mdBook
         uses: peaceiris/actions-mdbook@v1
+        with:
+          mdbook-version: "latest"
```
bilelmoussaoui, line 42: latest is the default no?

Hofer-Julian: Mmmh, but I used the latest on Gitlab as well...

Then that might not be it.

book/src/prerequisites.md
```diff
@@ -1,21 +1,73 @@
...
 
-# Buildsystem
+Now you can run your application by executing:
+```bash
+cargo run
+```
+
+## Cargo + Meson
+
+Cargo is *nearly* enough, but it is not well suited for handling resources such as icons or UI files.
```
bilelmoussaoui, line 34: I would add handling translation through gettext as well

.github/workflows/book.yml
Hofer-Julian: I also tried it locally, and there this problem does not exist.

Maybe install mdbook with cargo for now?

Hofer-Julian: Or like the book does
https://github.com/rust-lang/book/blob/master/.github/workflows/main.yml

book/src/prerequisites.md
```diff
@@ -1,21 +1,73 @@
...
+You only have to assure that Flatpak is installed on your system, check this [website](https://flatpak.org/setup/) to see if any steps are necessary on your distribution.
+Also download the [gtk-rust-template](https://gitlab.gnome.org/bilelmoussaoui/gtk-rust-template) and initialize it.
+
+Then either install:
+
+- [GNOME Builder](https://flathub.org/apps/details/org.gnome.Builder) or
+- [VSCodium](https://flathub.org/apps/details/com.vscodium.codium) together with the [rust-analyzer](https://open-vsx.org/extension/matklad/rust-analyzer) and [flatpak](https://open-vsx.org/extension/bilelmoussaoui/flatpak-vscode) extensions.
+
+That is it.
+Both IDEs will install all necessary dependencies for you, with GNOME Builder you only have to press the run button for that.
```
bilelmoussaoui, line 71: That's not actually true, only Builder is capable of doing so. Unless you meant the application dependencies

book/src/prerequisites.md
Hofer-Julian: Kind of included in "GTK functionality".

I can replace "GTK functionality" with actual examples, but I wonder if that might be confusing for total GTK beginners

book/src/prerequisites.md
Hofer-Julian: I thought runtime is installed by the extension or is this only planned?

bilelmoussaoui: It was never the case? planned, yes, no idea when i will get the time to look at it. Also, currently, only the GNOME 40 runtime contains gtk4, so you can probably just be explicit about how to install both the rust sdk & gnome one. and we can remove that once both IDEs can handle installing the extensions just fine.

book/src/prerequisites.md
bilelmoussaoui: I don't see any mention of "GTK functionality"? I wouldn't go too specific in details but still give few reasons of what Meson would simplify for the developer.

Hofer-Julian: Sure, which ones would you deem the most important?

.github/workflows/book.yml
bilelmoussaoui: This can be dropped per the matrix conversation :-)

book/src/prerequisites.md
bilelmoussaoui: It simplifies a bunch of things: 
- Translations: Simplify a bunch of stuff through the gettext module like translating appdata/desktop files or installing translations
- Building/installing GResources (UI files and all)
- Installing files such as icons

book/src/prerequisites.md
```diff
@@ -1,21 +1,73 @@
-# Installation
+# Prerequisites
+
+There are a few recommended ways to set up your working station in order to develop `gtk-rs` applications.
```
sdroege, line 3:
```suggestion
There are a few recommended ways to set up your workstation in order to develop `gtk-rs` applications.
```

Or "development environment" maybe?

book/src/prerequisites.md
```diff
@@ -1,21 +1,73 @@
-# Installation
+# Prerequisites
```
sdroege, line 1: Should this include a paragraph about getting GTK4 and installing the Rust toolchain (e.g. via rustup)?

sdroege: I mean, separate from the `cargo` / `cargo + meson` sections

book/src/prerequisites.md
sdroege: Also settings, the schemas

book/src/prerequisites.md
```diff
@@ -1,21 +1,73 @@
...
+Now you can run your application by executing:
+```bash
+cargo run
+```
+
+## Cargo + Meson
+
+Cargo is *nearly* enough, but it is not well suited for handling resources such as icons or UI files.
+That is why we recommend to use [Meson](https://mesonbuild.com/) on top of it.
+It is cross-platform, its syntax is very readable and it offers many convenience functionalities for GTK.
```
sdroege, line 36:
```suggestion
It is cross-platform, its syntax is very readable and it offers many convenience features for GTK.
```

book/src/prerequisites.md
Hofer-Julian: I thought about that. 
However, they are only two sentences and I am afraid that the overview gets worse if I put this separate.

Hofer-Julian:
Should be fine now

book/src/prerequisites.md
```diff
@@ -1,21 +1,76 @@
...
+```
+
+## Cargo + Meson
+
+Cargo is *nearly* enough, but it is not well suited for handling resources such as icons or UI files.
+That is why we recommend to use [Meson](https://mesonbuild.com/) on top of it.
+It is cross-platform and its syntax is very readable.
+Meson takes care of
+- translations
+- building and installing GResources
```
bilelmoussaoui, line 39: Could use a link to what a GResource is https://developer.gnome.org/gio/stable/GResource.html

book/src/prerequisites.md
```diff
@@ -1,21 +1,76 @@
...
+- installing auxiliary files such as icons and schemas
```
bilelmoussaoui, line 40: Same for the schemas here https://developer.gnome.org/GSettings/ I guess

book/src/prerequisites.md
Hofer-Julian: I think, I'll link to the corresponding chapter in the book instead.
Good point though :)

PR #266, opened 2021-04-05
CI: simplify by using a configuration matrix

bilelmoussaoui:


sdroege:
Does this also apply to the gtk-rs repo?

bilelmoussaoui:
It's what gtk-rs uses actually, just copied it back here

PR #268, opened 2021-04-05
more widget class methods

bilelmoussaoui:
Depends on #251 being merged first

As those widget_class methods are mostly wrappers around gtk::Shortcut and the C ones not being bindings friendly, I decided to rewrote them instead. 

Part of #111

### DIFF
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -983,6 +983,67 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
         }
     }
 
+    #[doc(alias = "gtk_widget_class_add_binding")]
+    fn add_binding<
+        F: Fn(&<<Self as ClassStruct>::Type as ObjectSubclass>::Type, Option<&Variant>) -> bool
+            + 'static,
+    >(
+        &mut self,
+        keyval: gdk::keys::Key,
+        mods: gdk::ModifierType,
+        callback: F,
+        arguments: Option<&glib::Variant>,
+    ) {
+        let shortcut = crate::Shortcut::new(
+            Some(&crate::KeyvalTrigger::new(keyval, mods)),
+            Some(&crate::CallbackAction::new(move |widget, args| -> bool {
+                unsafe { callback(widget.unsafe_cast_ref(), args) }
+            })),
+        );
+        shortcut.set_arguments(arguments);
+        self.add_shortcut(&shortcut);
+    }
+
+    #[doc(alias = "gtk_widget_class_add_binding_action")]
+    fn add_binding_action(
+        &mut self,
+        keyval: gdk::keys::Key,
+        mods: gdk::ModifierType,
+        action_name: &str,
+        arguments: Option<&glib::Variant>,
+    ) {
+        let shortcut = crate::Shortcut::new(
+            Some(&crate::KeyvalTrigger::new(keyval, mods)),
+            Some(&crate::NamedAction::new(action_name)),
+        );
+        shortcut.set_arguments(arguments);
+        self.add_shortcut(&shortcut);
+    }
+
+    #[doc(alias = "gtk_widget_class_add_binding_signal")]
+    fn add_binding_signal(
+        &mut self,
+        keyval: gdk::keys::Key,
+        mods: gdk::ModifierType,
+        signal_name: &str,
+        arguments: Option<&glib::Variant>,
+    ) {
+        let type_ = <Self::Type as ObjectSubclassType>::type_();
+        if glib::subclass::SignalId::lookup(signal_name, type_).is_none() {
+            panic!(
+                "Signal '{}' doesn't exists for type '{}'",
+                signal_name, type_
+            );
+        }
+
+        let shortcut = crate::Shortcut::new(
+            Some(&crate::KeyvalTrigger::new(keyval, mods)),
+            Some(&crate::SignalAction::new(signal_name)),
+        );
+        shortcut.set_arguments(arguments);
+        self.add_shortcut(&shortcut);
+    }
+
     #[doc(alias = "gtk_widget_class_add_shortcut")]
     fn add_shortcut(&mut self, shortcut: &Shortcut) {
         unsafe {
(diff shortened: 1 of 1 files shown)


gtk4/Gir.toml
```diff
@@ -561,6 +560,15 @@ name = "Gtk.BuilderListItemFactory"
 status = "generate"
 trust_return_value_nullability = false
 
+[[object]]
+name = "Gtk.CallbackAction"
+status = "generate"
+    [[object.function]]
+    name = "new"
+    [[object.function.parameter]]
```
sdroege, line 568: Indentation please :)

gtk4/Gir.toml
```diff
@@ -561,6 +560,15 @@ name = "Gtk.BuilderListItemFactory"
...
+    name = "callback"
+    nullable = false
```
sdroege, line 570: Why? It's not marked as nullable here

sdroege: I.e. did someone break annotations after 4.0.3 or why is this currently necessary here?

gtk4/src/subclass/widget.rs
```diff
@@ -983,6 +983,75 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+        self.add_shortcut(&shortcut);
+    }
+
+    #[doc(alias = "gtk_widget_class_add_binding_signal")]
+    fn add_binding_signal(
+        &mut self,
+        keyval: gdk::keys::Key,
+        mods: gdk::ModifierType,
+        signal_name: &str,
+        arguments: Option<&glib::Variant>,
```
sdroege, line 1029: Shouldn't this get some checks to see whether the signal exists and has the correct parameter type(s)?

gtk4/src/subclass/widget.rs
```diff
@@ -983,6 +983,75 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+        self.add_shortcut(&shortcut);
+    }
+
+    #[doc(alias = "gtk_widget_class_add_binding_signal")]
+    fn add_binding_signal_id(
+        &mut self,
+        keyval: gdk::keys::Key,
+        mods: gdk::ModifierType,
+        signal: glib::subclass::SignalId,
+        arguments: Option<&glib::Variant>,
```
sdroege, line 1045: Shouldn't this get some checks to see whether the signal exists and has the correct parameter type(s)?

gtk4/src/subclass/widget.rs
```diff
@@ -983,6 +983,75 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
+        self.add_shortcut(&shortcut);
+    }
+
+    #[doc(alias = "gtk_widget_class_add_binding_action")]
+    fn add_binding_action(
+        &mut self,
+        keyval: gdk::keys::Key,
+        mods: gdk::ModifierType,
+        action_name: &str,
+        arguments: Option<&glib::Variant>,
```
sdroege, line 1013: Shouldn't this get some checks to see whether the action exists and has the correct parameter type(s)?

gtk4/Gir.toml
bilelmoussaoui: Otherwise you have to pass in `Option<Box<F>>`, I don't see why you would create a `CallbackAction` that does nothing?

sdroege: Sure but in my .gir the callback is not marked as `nullable` so why is it necessary to override this here?

bilelmoussaoui: The CallbackAction takes a `GtkShortcutFunc` defined as such `
pub type GtkShortcutFunc = 
    Option<unsafe extern "C" fn(*mut GtkWidget, *mut glib::GVariant, gpointer) -> gboolean>;`, like all other typed callbacks. I suppose maybe is this reason why its end up with a nullable parameter?

sdroege: Seems like a gir bug then :)

gtk4/src/subclass/widget.rs
bilelmoussaoui: Ideally yes, but upstream doesn't neither and just has `/* XXX: validate variant format for signal */` same for the actions. I will see how doable that's

sdroege: The problem I see here is that this can easily cause memory corruption if it's not checked on the C side either.

bilelmoussaoui: What's a bit difficult here is how to compare the types that you can get from a `SignalQuery::param_types` (returns a `Vec<SignalType>`) and `Variant::type_` (returns a `VariantTy`). Any ideas how to do that?

sdroege: How is the C code converting between them?

gtk4/src/subclass/widget.rs
bilelmoussaoui: After asking in #gtk channel, we can't really check whether the action exists as it can be registered later on. It can even be defined in a different action group that's added to this widget. Which makes the validation much more complex than what I imagined. I added a TODO comment about that for now

gtk4/Gir.toml
bilelmoussaoui: Opened an issue at https://github.com/gtk-rs/gir/issues/1104

bilelmoussaoui:
After digging a bit further the ShortcutAction code, it seems that gtk does actually validate the parameter types of both signals (https://github.com/GNOME/gtk/blob/master/gtk/gtkshortcutaction.c#L767-L846) & actions (https://github.com/GNOME/gtk/blob/master/gtk/gtkshortcutaction.c#L1040-L1075) and does show a warning when the params are of an incorrect type. So it shouldn't actually cause an issue per see. 

It also uses some internal gtk functions during the validation, duplicating them here seems rather of a complex task and I don't really see the benefits of doing so right now. As at the end we will just have a warning/panic during the shortcut registration phase instead of the invocation phase. 

I would say, we can land this as is for now. As they are very handy helpers :)

sdroege:
> I would say, we can land this as is for now. As they are very handy helpers :)

Ok then :)

gtk4/Gir.toml
fengalin: IIUC, the callback parameter is marked `nullable = "1"` in `Gtk-4.0.gir`. See line 10 305.

Issue #270, opened 2021-04-06
Change URL of docs and book


Hofer-Julian:
The urls are like this:
https://gtk-rs.org/gtk4-rs/book/git/index.html
https://gtk-rs.org/gtk4-rs/docs/git/gtk4/

Relative paths would work if they would be like this:
https://gtk-rs.org/gtk4-rs/git/book/index.html
https://gtk-rs.org/gtk4-rs/git/docs/gtk4/

The same should be done for future stable book and docs.

PR #271, opened 2021-04-07
Accomodate Gio changes

Hofer-Julian:
- Fixes examples
- Fixes books
- Let the book tests also run when the book was not changed (don't think that fixes the problem we discussed though @bilelmoussaoui)


.github/workflows/book.yml
```diff
@@ -5,8 +5,6 @@ on:
   push:
     branches:
       - "master"
-    paths:
```
sdroege, line 8: The changes to the book directory and this here should probably be in a separate commit from the "Fix examples" commit?

Hofer-Julian: Done

bilelmoussaoui: This should fix the issue yes, although, I wonder if there's a way to specify a whole directory with the files in it to trigger a build.

Hofer-Julian: In that case, that's not what we want though.
Changes outside the book can break the book examples as well.

bilelmoussaoui: They won't pass the CI though as the job is always run for every PR

Hofer-Julian: Ah, I see

bilelmoussaoui:
I don't believe the CI changes were acked by me. Please don't merge directly next time.

Hofer-Julian:
Sorry @bilelmoussaoui, I thought this PR already had one approval, but it indeed did not.

sdroege:
Anything needs fixes here still?

bilelmoussaoui:
Not really, it's just the CI "solution" doesn't seems right to me; I will take care of that along with #270

Hofer-Julian:
Again, I said that I DON'T see that as the solution of the CI problem (see above 😉)

bilelmoussaoui:
it's not a solution if you merge directly to master, which you should never do :-)

Issue #256, opened 2021-04-04
Properties comments


sdroege:
See https://gitlab.gnome.org/Hofer-Julian/gtk-rs-book/-/issues/14

PR #273, opened 2021-04-09
Refactor custom objects in own modules

Hofer-Julian:


book/listings/gobject_properties/3/custom_button/imp.rs
```diff
@@ -0,0 +1,83 @@
...
+                i32::MAX,
+                // Default value
+                0,
+                // The property can be read and written to
+                ParamFlags::READWRITE,
+            )]
+        });
+        PROPERTIES.as_ref()
+    }
+    fn set_property(&self, _obj: &Self::Type, _id: usize, value: &Value, pspec: &ParamSpec) {
```
bilelmoussaoui, line 52: Could use an empty line here to be consistent with the other functions in ObjectImpl

book/listings/gobject_properties/3/custom_button/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+
+use glib::{BindingFlags, Object};
+use gtk::{glib, Align, Orientation};
+use gtk::{prelude::*, BoxBuilder};
+use gtk::{Application, ApplicationWindowBuilder};
+use std::cell::RefCell;
+
+glib::wrapper! {
+    pub struct CustomButton(ObjectSubclass<imp::CustomButton>)
+        @extends gtk::Button, gtk::Widget;
```
bilelmoussaoui, line 11: Any reason for putting those into a new line?

Hofer-Julian:
Thanks for the fast review, this is just a draft PR to see if it only fails locally or also on the CI 😊

I will clean up many things later on.

book/listings/gobject_properties/3/custom_button/mod.rs
Hofer-Julian: Is there a reason not to?
I like it more, because if it is in the same line users with a smaller monitor might overlook it

bilelmoussaoui: That's way below the characters per line limit set by rustfmt i think, no preferences though :-)

sdroege: I usually put a linebreak before the @extends / @implements part as the line tends to get quite long otherwise

bilelmoussaoui: @sdroege you have tagged two random persons :p

sdroege: Well!

book/listings/gobject_subclassing/2/custom_button/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+}
+
+impl CustomButton {
+    pub fn new() -> Self {
+        glib::Object::new(&[]).expect("Failed to create Button")
+    }
+
+    pub fn with_label(label: &str) -> Self {
+        let button = Self::new();
+        button.set_label(label);
```
sdroege, line 18: This is an anti-pattern btw. Better to do

```suggestion
        glib::Object::new(&[("label", label)]).expect("Failed to create Button")
```

book/listings/gobject_subclassing/2/custom_button/imp.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Trait shared by all widgets
+impl WidgetImpl for CustomButton {}
+
+// Trait shared by all buttons
+impl ButtonImpl for CustomButton {
+    fn clicked(&self, button: &Self::Type) {
+        *self.number.borrow_mut() += 1;
+        button.set_label(&self.number.borrow().to_string())
```
sdroege, line 35: Wouldn't it make more sense to prefix the label passed to the constructor here, for example?

sdroege (commented):
Seems good to me otherwise

book/listings/gobject_subclassing/2/custom_button/mod.rs
Hofer-Julian: Thanks, changed it accordingly.

book/listings/gobject_subclassing/2/custom_button/imp.rs
Hofer-Julian: I see your point,
Actually I never need to pass a label, apart from a single time at the subclassing chapter. So I just removed `with_label` for the other CustomButton appearances.

Hofer-Julian:
Ideally this one would be merged before I start with https://github.com/gtk-rs/gtk4-rs/issues/256
Maybe someone can have a final look.

PR #281, opened 2021-04-10
gdk-x11: manual impl get_xdisplay and get_xsreen

### LINKED ISSUE #170: get_xdisplay() is missing from gdk4-x11::X11Display
Hi, 

I want to move a window on X11, so I need to get access to the respective X11 display and window. 
Following the GDK4 documentation this API call should be used: https://developer.gnome.org/gdk4/stable/gdk4-X-Window-System-Interaction.html#gdk-x11-display-get-xdisplay

Gir.toml mentions it in line 55: https://github.com/gtk-rs/gtk4-rs/blob/master/gdk4-x11/Gir.toml#L55
But the source doesn't contain the API call: https://github.com/gtk-rs/gtk4-rs/blob/master/gdk4-x11/src/auto/x11_display.rs

I'm not sure if this is really a bug, or if I need to generate something. A hint how to fix it manually would also be welcome.

gtk4-rs, gdk4-x11

alatiera:
Close #170

### DIFF
--- a/gdk4-x11/src/x11_display.rs
+++ b/gdk4-x11/src/x11_display.rs
@@ -1,0 +1,17 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::X11Display;
+use glib::translate::ToGlibPtr;
+use x11::xlib;
+
+impl X11Display {
+    #[doc(alias = "gdk_x11_display_get_xdisplay")]
+    pub unsafe fn get_xdisplay(&self) -> *mut xlib::Display {
+        ffi::gdk_x11_display_get_xdisplay(self.to_glib_none().0)
+    }
+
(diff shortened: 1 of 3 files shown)


gdk4-x11/src/x11_display.rs
```diff
@@ -0,0 +1,16 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::X11Display;
```
bilelmoussaoui, line 3: Needs to be added to lib.rs

alatiera: That explains why it always was compiling no matter what I did!

bilelmoussaoui (approved):

sdroege (approved):

PR #285, opened 2021-04-10
rename get getters

bilelmoussaoui:
depends on https://github.com/gtk-rs/gir/pull/1021

cc @fengalin 
I have noticed few properties having their getter removed, is that normal? there were also some ones that didn't got renamed. I will have another look tomorrow and point them to see whether there's something actionable or not. Otherwise seems to be working fine!


gdk4/src/auto/clipboard.rs
```diff
@@ -298,21 +298,6 @@ impl Clipboard {
         }))
     }
 
-    pub fn get_property_local(&self) -> bool {
```
bilelmoussaoui, line 301: Like this property

gdk4/src/auto/content_formats.rs
```diff
@@ -31,7 +31,7 @@ impl ContentFormats {
     }
 
     #[doc(alias = "gdk_content_formats_new_for_gtype")]
-    pub fn new_for_gtype(type_: glib::types::Type) -> ContentFormats {
+    pub fn for_gtype(type_: glib::types::Type) -> ContentFormats {
```
bilelmoussaoui, line 34: should probably be `for_type`

gdk4/src/auto/content_serializer.rs
```diff
@@ -24,12 +24,12 @@ impl ContentSerializer {
     }
 
     #[doc(alias = "gdk_content_serializer_get_gtype")]
-    pub fn get_gtype(&self) -> glib::types::Type {
+    pub fn gtype(&self) -> glib::types::Type {
```
bilelmoussaoui, line 27: `type_` to be consistent with the rest

gdk4/src/auto/device.rs
```diff
@@ -118,22 +118,8 @@ impl Device {
         unsafe { from_glib(ffi::gdk_device_has_bidi_layouts(self.to_glib_none().0)) }
     }
 
-    pub fn get_property_has_bidi_layouts(&self) -> bool {
```
bilelmoussaoui, line 121: This one got removed as well

gdk4/src/auto/display.rs
```diff
@@ -175,22 +175,8 @@ impl Display {
         }
     }
 
-    pub fn get_property_composited(&self) -> bool {
```
bilelmoussaoui, line 178: And this one

gdk4/src/auto/clipboard.rs
fengalin: This one is found to be duplicate of [`is_local`](https://github.com/gtk-rs/gtk4-rs/blob/8107dfccc9c8a3939f1eaf45b23911dfc05d986e/gdk4/src/auto/clipboard.rs#L47). Either it's fine, or you need to use `rename` or `bypass_auto_rename` in `Gir.toml` to control what should get generated.

gdk4/src/auto/device.rs
fengalin: Same as for `is_local`. See https://github.com/gtk-rs/gtk4-rs/blob/8107dfccc9c8a3939f1eaf45b23911dfc05d986e/gdk4/src/auto/device.rs#L117.

gdk4/src/auto/display.rs
fengalin: Same as... you got the idea :) https://github.com/gtk-rs/gtk4-rs/blob/8107dfccc9c8a3939f1eaf45b23911dfc05d986e/gdk4/src/auto/display.rs#L134

gdk4/src/auto/clipboard.rs
bilelmoussaoui: Aha, Indeed. I will have a look at the other ones

gdk4/src/auto/content_formats.rs
fengalin: That would be tricky. It would be a combination of 2 Gir rules: this one is the detection of the `new_for` pattern [that was added yesterday](https://github.com/fengalin/gir/blob/4384773b3def43311991166f93c7091ebae05f70/src/analysis/functions.rs#L565), which is different from detecting a substitution for `get_gtype`. I think this one should be `renamed` manually in `Gir.toml`.

gtk4/src/subclass/editable.rs
```diff
@@ -21,12 +21,12 @@ pub trait EditableImpl: WidgetImpl {
...
 
-    fn get_text(&self, editable: &Self::Type) -> GString {
-        self.parent_get_text(editable)
+    fn text(&self, editable: &Self::Type) -> GString {
+        self.parent_text(editable)
     }
 
-    fn get_delegate(&self, editable: &Self::Type) -> Option<Editable> {
-        self.parent_get_delegate(editable)
+    fn delegate(&self, editable: &Self::Type) -> Option<Editable> {
```
fengalin, line 28: One of the issues with not using the migration tools ([fn def](https://crates.io/crates/fix-getters-def) and [call sites](https://crates.io/crates/fix-getters-calls)) is that you don't take advantage of the heuristics for deciding whether a function is a getter for sure.

This function is not a getter in the sense: "a function that retrieves current value of a field". After the rename, it reads more like it performs an imperative `delegate` action. I think we lost a meaning in the process. There are many examples like this in the manual updates. You also take the risk of missing some [useful substitutions](https://github.com/fengalin/fix-getters/blob/26c950680699cdb8a096ff454bd2f858c018a0a7/rules/src/function.rs#L81) or not renaming things like `get_mut_structure` as the more common `get_structure_mut` among other cases I identified when migrating `gtk-rs`.

I strongly advise to follow this workflow:

- Regen with https://github.com/gtk-rs/gir/pull/1021.
- Check the result and update `Gir.toml` if needed.
- Run the migrations tools (you can install them using `cargo`):
  - `fix-getters-def`
  - `fix-getters-calls`
  - `rust fmt`
- Compile and fix the remaining candidates on a case by case basis.

An additional benefit for this is that application developers will be able to use the same heuristics (using the [call sites](https://crates.io/crates/fix-getters-calls) migration tool) to ease their application migration process. Not using these is like throwing my work away :). The point in these tools was that `sed` was not enough to do the migration properly.

I can do the work on `gtk4-rs` if you don't have time to go through this. Just let me finish the on-going migration of `gstreamer-rs` & `gst-plugins-rs`.

fengalin (changes requested):

gtk4/src/subclass/editable.rs
bilelmoussaoui: Well, the thing is it's a getter. The semantics are a bit different here because it's a virtual function but at the end, you must return the Editable implementation you want to "delegate" and at the end you do return an internal field. 

I will regenerate using your tool, but I believe this function should be renamed nonetheless.

fengalin: IMHO, it's not a substitute to using a `struct` field, in the sense of calling a function to read a field that you could read directly if the field was `pub`. In this case, you *retrieve* a value by indicating a type, that's why I believe we should keep `get` here.

fengalin: Oh, I just got your point about the sublassing part. So it might not be the best example. But I think we should get an agreement on the range for which we decide to remove the `get` prefix. I expected a debate on this, but it didn't happen.

bilelmoussaoui: Makes sense, at least for this one as after looking at the source code of Editable. The `delegate` is not a property per see but just an internal thing with a getter method to read it.

fengalin: Please, tell me what you think about the general selection rules I listed there: https://github.com/fengalin/fix-getters#get-functions-selection.

bilelmoussaoui:
This can be closed now, thanks to @fengalin's PR! :D

PR #288, opened 2021-04-11
[docs] Adds an introduction for the gtk4 crate

### LINKED ISSUE #255: Write an introduction for gtk4 crate
for src/lib.rs so that it appears when you open the gtk4 docs

sophie-h:
Based on gtk-rs/gtk-rs#431

Closes #255

### DIFF
--- a/gtk4/src/lib.rs
+++ b/gtk4/src/lib.rs
@@ -1,6 +1,164 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-// TODO: Introduction
+//! # Rust GTK 4 bindings
+//!
+//! This library contains safe Rust bindings for [GTK 4](http://www.gtk.org), a
+//! multi-platform GUI toolkit. It is a part of [Gtk-rs](http://gtk-rs.org/).
+//!
+//! The library is a work in progress. Expect breaking changes between releases.
+//! Most of this documentation is generated from the C API.
+//! Until all parts of the documentation have been reviewed there will be incongruities
+//! with the actual Rust API.
+//!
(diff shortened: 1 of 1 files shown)


gtk4/src/lib.rs
```diff
@@ -1,6 +1,161 @@
...
+//!
+//! ## Library versions
+//!
+//! By default this crate provides only GTK 4.0 APIs. You can access additional
+//! functionality by selecting one of the `v4_2`, `v4_4`, etc. features.
+//!
+//! `Cargo.toml` example:
+//!
+//! ```toml
+//! [dependencies.gtk4]
```
bilelmoussaoui, line 135: I tend to prefer renaming the package to gtk here and use that instead of doing `gtk4::`, especially that gsk/gdk are re-exported as  `gdk` `gsk` and not `gdk4`, `gsk4`. Maybe do that here as well?

bilelmoussaoui:
Otherwise looks good to me, thanks a lot!

gtk4/src/lib.rs
sophie-h: I agree. Changed it.

What do you think about the `gtk4` in the code examples above?

bilelmoussaoui: should probably be renamed to use gtk as well, like everywhere else

sophie-h: Renamed and tried to explain it

sophie-h:
Doctests will pass after #289 is merged

bilelmoussaoui:
I will go ahead and merge it, improvements can be done in separate PR if someone has any. Thanks!

Issue #236, opened 2021-03-26
GTK should be marked as initialized after `parent_startup()`


YaLTeR:
Right now it's marked as initialized only after the entire `startup()` in a `GtkApplication` subclass finishes, however you might reasonably want to do some GTK stuff after calling `parent_startup()`:

```rust
    fn startup(&self, self_: &Self::Type) {
            self.parent_startup(self_);

            let provider = gtk::CssProvider::new();
            provider.load_from_resource("/org/gnome/gitlab/YaLTeR/VideoTrimmer/style.css");
            gtk::StyleContext::add_provider_for_display(
                &gdk::Display::get_default().unwrap(),
                &provider,
                gtk::STYLE_PROVIDER_PRIORITY_APPLICATION,
            );
    }
```

GTK is initialized in `parent_startup()`: https://gitlab.gnome.org/GNOME/gtk/-/blob/cc64dbafbc8158d5116bf2afe2d422da48b77137/gtk/gtkapplication.c#L262

cc @bilelmoussaoui

sdroege:
The same change will also be needed for gtk3

YaLTeR:
To reproduce, add this to the [application subclass example](https://github.com/gtk-rs/gtk4-rs/blob/cb5896a15fe712ed3568b9a3e20d1b02d150faa7/examples/src/bin/application_subclass.rs#L25).

```rust
        fn startup(&self, application: &Self::Type) {
            self.parent_startup(application);

            let _ = gtk::Button::new();
        }
```

sdroege:
The problem here is

```rust
unsafe extern "C" fn application_startup<T: ObjectSubclass>(ptr: *mut gio::ffi::GApplication)
where
    T: GtkApplicationImpl,
{
    let instance = &*(ptr as *mut T::Instance);
    let imp = instance.get_impl();
    let wrap: Borrowed<gio::Application> = from_glib_borrow(ptr);
    imp.startup(wrap.unsafe_cast_ref());
    crate::rt::set_initialized();
}
```

It would only be initialized after the `startup` implementation of the subclass. Somehow we need to hook into the `startup` of `GtkApplication` directly, the one that is called as part of `parent_startup` above.

sdroege:
See https://github.com/gtk-rs/gtk4-rs/pull/833

sdroege:
This is fixed in 0.4.3 now

PR #289, opened 2021-04-11
[gtk] Update to gtk3 handling of Application

### LINKED ISSUE #287: [gtk] Make ApplicationBuilder::build() return type consistent
`gtk::ApplicationBuilder::build()` returns a `Result` in gtk4 (manual implementation) and an `Application` in gtk3 (auto implementation.)

I would suggest to have something like `build()` that panics as the default, because everyone writes `expect()` there anyways. And then maybe a `try_build()` with a `Result`.

sophie-h:
Closes #287

### DIFF
--- a/gtk4/src/auto/application.rs
+++ b/gtk4/src/auto/application.rs
@@ -10,6 +10,7 @@ use glib::signal::connect_raw;
 use glib::signal::SignalHandlerId;
 use glib::translate::*;
 use glib::StaticType;
+use glib::ToValue;
 use std::boxed::Box as Box_;
 use std::fmt;
 use std::mem::transmute;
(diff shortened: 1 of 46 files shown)


gtk4/src/auto/application.rs
```diff
@@ -22,6 +23,101 @@ glib::wrapper! {
     }
 }
 
+impl Application {
+    #[doc(alias = "gtk_application_new")]
+    pub fn new(application_id: Option<&str>, flags: gio::ApplicationFlags) -> Application {
+        assert_initialized_main_thread!();
```
sdroege, line 29: This should also call `register_startup_hook()` or not?

sophie-h: I don't think so because [`gtk_application_new`](https://developer.gnome.org/gtk3/stable/GtkApplication.html#gtk-application-new) should do everything for us as of docs.

sdroege: It does not call https://github.com/gtk-rs/gtk4-rs/blob/8107dfccc9c8a3939f1eaf45b23911dfc05d986e/gtk4/src/rt.rs#L69 though

sophie-h: This sounds like a problem :upside_down_face:

sophie-h: So we want to manual version but should no longer return a `Result`

sdroege: I think so. This is all a bit annoying :)

sdroege (approved):

sophie-h:
Is this change somehow related to #236? I am not able to parse that issue in my head rn.

sophie-h:
> Is this change somehow related to #236?

I guess only in the way, that we wouldn't need the manual implementation of `new` anymore if this issue is fixed.

sdroege:
> Is this change somehow related to #236? I am not able to parse that issue in my head rn.

Related, yes. The problem there is more difficult though: GTK is only actually initialized in the class handler of `GtkApplication::startup` (i.e. what `parent_startup` in that example code would execute). Before that it must be forbidden to call any GTK functions (unless it was initialized in another way), but we didn't find a way yet to guarantee that.

bilelmoussaoui (approved):

PR #291, opened 2021-04-11
Address Sebastian's property comments

### LINKED ISSUE #256: Properties comments

Hofer-Julian:
- Clarifying usage of properties
- Introducing setter and getter methods
- Explain unwraps
- Briefly explain what `glib::Value`s are
- Explain that `once_cell` will join stable std
- Go more into depth with powers of property binding

Fixes #256


book/listings/gobject_properties/4/main.rs
```diff
@@ -0,0 +1,87 @@
...
+                .get::<i32>()
+                .expect("The property needs to be of type `i32`.")
+                .expect("The property needs to be `Some`.");
+            let incremented_number = number + 1;
+            Some(incremented_number.to_value())
+        })
+        // How to transform "number" from `button_2` to "number" of `button_1`
+        .transform_from(|_, value| {
+            let number = value
+                .get::<i32>()
```
sdroege, line 49:
```suggestion
                .get_some::<i32>()
```

And then you can omit an `unwrap()`. Also above

book/listings/gobject_properties/2/main.rs
```diff
@@ -22,18 +21,22 @@ fn on_activate(application: &Application) {
...
-        .bind_property("state", &switch_2, "state")
-        .flags(BindingFlags::BIDIRECTIONAL)
-        .build();
-    // ANCHOR_END: bind_state
+    // Set and then immediately obtain state
+    switch.set_property("state", &true).unwrap();
+    let current_state = switch
+        .get_property("state")
+        .expect("The property needs to exist and be readable.")
+        .get::<bool>()
```
sdroege, line 33:
```suggestion
        .get_some::<bool>()
```

book/listings/gobject_properties/4/main.rs
```diff
@@ -0,0 +1,87 @@
...
+    // ANCHOR_END: buttons
+
+    // ANCHOR: bind_numbers
+    // Assure that "number" of `button_2` is always 1 higher than "number" of `button_1`
+    button_1
+        .bind_property("number", &button_2, "number")
+        // How to transform "number" from `button_1` to "number" of `button_2`
+        .transform_to(|_, value| {
+            let number = value
+                .get::<i32>()
```
sdroege, line 40:
```suggestion
                .get_some::<i32>()
```

book/listings/gobject_properties/4/main.rs
```diff
@@ -0,0 +1,87 @@
...
+        .build();
+    // ANCHOR_END: bind_numbers
+
+    // ANCHOR: connect_notify
+    // The closure will be called whenever the property "number" of `button_1` gets changed
+    button_1.connect_notify_local(Some("number"), move |button, _| {
+        let number = button
+            .get_property("number")
+            .expect("The property needs to exist and be readable.")
+            .get::<i32>()
```
sdroege, line 65:
```suggestion
            .get_some::<i32>()
```

book/src/gobject_properties.md
```diff
@@ -1,35 +1,48 @@
...
 
 <span class="filename">Filename: listings/gobject_properties/1/main.rs</span>
 
 ```rust ,no_run
 {{#rustdoc_include ../listings/gobject_properties/1/main.rs:switch}}
 ```
+Alternatively, we can use the general [`get_property`](http://gtk-rs.org/docs/glib/object/trait.ObjectExt.html#tymethod.get_property) and [`set_property`](http://gtk-rs.org/docs/glib/object/trait.ObjectExt.html#tymethod.set_property) methods.
+Because they can be used for properties of different types, they operate with the polymorphic [`glib::Value`](http://gtk-rs.org/docs/glib/value/struct.Value.html).
+The underlying type of `glib::Value` can be a boolean, integer, float or a GObject.
+It can even be optional and corresponds then to `Option<T>`.
```
sdroege, line 18: Only for some types though: all types that are nullable, so objects, strings, boxed types

book/src/gobject_properties.md
```diff
@@ -1,35 +1,48 @@
...
+All of that, can be determined at runtime, which is why the checks have to happen at runtime as well.
```
sdroege, line 19: Maybe add a note here that this is conceptually like a Rust enum

```rust
enum Value {
    Bool(bool),
    Int(i32),
    String(String),
    ...
```

book/src/gobject_properties.md
```diff
@@ -39,21 +52,26 @@ Whenever we now click on one of the two switches, the other one gets toggled as
 
 We can also add properties to custom GObjects.
 We can demonstrate that, by binding the `number` of our `CustomButton` to a property.
+For that, we need to be able to [lazily evaluate](https://en.wikipedia.org/wiki/Lazy_evaluation) expressions.
+`once_cell` does that for us and is already part of `std` of Rust nightly.
```
sdroege, line 56: It provides us with many other types too. The one we use here is called `Lazy` :)

book/src/gobject_properties.md
```diff
@@ -39,21 +52,26 @@ Whenever we now click on one of the two switches, the other one gets toggled as
...
 
 ```toml
 [dependencies]
 once_cell = "1"
 ```
 
-Then we have to define the property within the `ObjectImpl` implementation.
+Now we define the "number" property within the `ObjectImpl` implementation.
+We can immediately take advantage of this new property by binding the "label" property to it.
+It even converts the integer value of "number" to the string if "label".
```
sdroege, line 68: "if label"?

sdroege (commented):
Looks good otherwise

book/src/gobject_properties.md
bilelmoussaoui: probably meant of "label"

bilelmoussaoui:
Looks good to me as well! thanks

Hofer-Julian:
I addressed the comments and attempted to write a small introduction chapter about `glib::Value`

book/src/gobject_values.md
```diff
@@ -0,0 +1,33 @@
+# Generic Values
+
+Some GObject-related functions rely on generic values for their arguments or return parameters.
+Enumerators in C are not as powerful as the one in Rust, which is why [`glib::Value`](http://gtk-rs.org/docs/glib/value/struct.Value.html) is used for this.
```
sdroege, line 4: Enums / enumerations would be the correct word. An enumerator would be a different word for an iterator

sdroege (approved):
Looks good otherwise

book/src/gobject_signals.md
```diff
@@ -13,15 +13,16 @@ In our "Hello World" example we [connected](../docs/gtk4/trait.ButtonExt.html#ty
 {{#rustdoc_include ../listings/gobject_signals/1/main.rs:callback}}
 ```
 
-If we wanted to, we could have connected to it with the general (but much more verbose) `connect_local` method.
+If we wanted to, we could have 
+connected to it with the generic `connect_local` method.
```
bilelmoussaoui, line 17: It would be nice to link to ObjectExt::connect_local here

book/src/gobject_values.md
Hofer-Julian: TIL, thanks!

book/src/gobject_signals.md
Hofer-Julian: Good point!

Hofer-Julian:
Should be fine now.

PR #293, opened 2021-04-11
CI: auto-deploy docs on release as well

bilelmoussaoui:


Hofer-Julian:
Should something similar be done for the book?

bilelmoussaoui:
Yes, this is just a test build to see if what i imagined can work or not

bilelmoussaoui:
Okay, this looks good! I will go ahead and merge it so we can check if the deploy job are running as expected as well. We should probably do a 0.0 release to check if the CI does what it should and remove it afterwards but that's for later.

PR #296, opened 2021-04-13
Remove get  for getters & properties where applicable

fengalin:
## Note to application developers

Applications developers should use [`fix-getters-calls`](https://crates.io/crates/fix-getters-calls) to ease migration of their applications. If you also wish to change your get functions definition to comply with the standards implemented here, use [`fix-getters-def`](https://crates.io/crates/fix-getters-def).

## Object

This is the continuation of https://github.com/gtk-rs/gir/pull/1021 and https://github.com/gtk-rs/gtk-rs/pull/211 as an attempt to address https://github.com/gtk-rs/gir/issues/963.

This one uses the [`fix-getters` tools](https://github.com/fengalin/fix-getters), which should ease migration for applications developers. Note however, that some more `get` functions might be renamed due to the conservative approach of these tools.

I used the same conflict resolutions that I used for [`gtk-rs`](https://github.com/gtk-rs/gtk-rs/pull/211) where applicable. Please take a look at the first commit ("gir prerp. for regen..."). If we decide to change some of these conflict resolutions, I'll update `gtk-rs` too.

@bilelmoussaoui: with @sdroege, we planned to perform a second pass on `gtk-rs`, `gtk4-rs` and `gstreamer-rs` to address a couple of edge cases related to `enum` generation and subclassing (similar to the [discussion we had in your PR](https://github.com/gtk-rs/gtk4-rs/pull/285#pullrequestreview-633269711)). See https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/issues/325. The idea is to get the big chunk in shortly and proceed with these details in the coming days.


book/listings/gobject_properties/4/custom_button/imp.rs
```diff
@@ -80,7 +80,7 @@ impl WidgetImpl for CustomButton {}
 // Trait shared by all buttons
 impl ButtonImpl for CustomButton {
     fn clicked(&self, button: &Self::Type) {
-        let incremented_number = self.number.borrow().clone() + 1;
+        let incremented_number = *self.number.borrow() + 1;
```
bilelmoussaoui, line 83: Interesting, we should run clippy for the book listings as well
cc @Hofer-Julian

gdk4/src/content_deserializer.rs
```diff
@@ -5,7 +5,7 @@ use glib::translate::*;
 
 impl ContentDeserializer {
     pub fn set_value(&self, value: glib::Value) {
-        assert!(value.type_() == self.gtype(), "Type mismatch");
+        assert!(value.type_() == self.typ_(), "Type mismatch");
```
bilelmoussaoui, line 8: Should probably be type_ instead

fengalin: Indeed :)

gtk4/Gir.toml
```diff
@@ -2138,10 +2150,24 @@ manual_traits = ["WidgetExtManual"]
...
     [[object.function]]
     name = "get_name"
-    rename = "get_widget_name"
+    rename = "widget_name"
     [[object.function]]
     name = "set_name"
     rename = "set_widget_name"
+    [[object.function]]
+    name = "get_sensitive"
+    # Avoid clash with existing `is_sensitive` with a different meaning.
```
bilelmoussaoui, line 2159: What's the difference meaning of is_sensitive/is_visible here?

fengalin: See https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-get-sensitive and `gtk_widget_is_sensitive` below. I'd be glade if someone could find names that make the difference obvious :)

gtk4/Gir.toml
```diff
@@ -2138,10 +2150,24 @@ manual_traits = ["WidgetExtManual"]
...
+    [[object.function]]
+    name = "get_visible"
+    # Avoid clash with existing `is_visible` with a different meaning.
+    bypass_auto_rename = true
+    [[object.function]]
+    name = "get_hexpand"
+    rename = "does_hexpand"
+    [[object.function]]
+    name = "get_vexpand"
+    rename = "does_vexpand"
```
fengalin, line 2170: These are getters return `bool` values, so I used a conjugated verb. It didn't seem correct to use the form `widget.is_vexpand()`, so I used `widget.does_vexpand()`, but we might use `widget.vexpands()`. More compact, but less obvious too read maybe.

What do you think?

gtk4/Gir.toml
bilelmoussaoui: I think we should rather rename is_sensitive to `is_effective_sensitive` (not great i know, but any suggestions are more than welcome) as per the getter rename rules it's get_sensitive that should become is_sensitive. The is_sensitive case is similar as well. As both those functions check whether the current widget is sensitive/visible along with it's parents as well.

gtk4/Gir.toml
bilelmoussaoui: I don't like the renamed variants, probably not renaming them is better than the `does_[h|v]expand` variants I find. 

Although the `vexpands`/`hexpands` looks much better to me to me.

gtk4/Gir.toml
fengalin: I didn't use this because it would change the behaviour of existing code. Maybe we should keep `is_sensitive` / `is_visible` as is, meaning that we test whether the widget is `sensitive` / `visible` in the resulting UI, and `is_widget_...` for the value of the widget itself... or we keep `get_...` :)

gtk4/Gir.toml
fengalin: > the vexpands/hexpands looks much better to me to me

Yes, I prefer this too.

gtk4/Gir.toml
fengalin: On a second thought, `is_widget_...` isn't helpful... I guess that's why I kept `get_...`. Finally `get_property_...` would be the most explicit I guess.

fengalin: So, I kept `get_{sensitive,visible}` and improved the comments in `Gir.toml`, so that we don't need to do all the thinking again :)

fengalin:
Will force push again shortly, just noticed that `gir` submodule was added in the wrong commit. Doesn't change the code though.

gtk4/Gir.toml
bilelmoussaoui: Thanks @fengalin

fengalin: Your're welcome!

book/listings/gobject_properties/4/custom_button/imp.rs
Hofer-Julian: Good idea!
Wann open a PR or issue for that?

PR #298, opened 2021-04-15
no more boxes

### LINKED ISSUE #294: Remove all the `Option<Box<dyn Fn()>>>` usages
There are a few callbacks that are still optional. We should investigate whether those makes sense to be optional or not, if it's the case split the function into a `set_callback_name` & `unset_callback_name`

bilelmoussaoui:
Fixes #294 

There's a very weird one, [`gtk::MapListModel::new`](https://docs.gtk.org/gtk4/ctor.MapListModel.new.html), I ended up adding a new method there named `MapListModel::without_map_func`. Better name suggestions are very much welcome!

### DIFF
--- a/gtk4/Gir.toml
+++ b/gtk4/Gir.toml
@@ -1039,6 +1039,12 @@ trust_return_value_nullability = false
 name = "Gtk.FontChooser"
 status = "generate"
 trust_return_value_nullability = false
+manual_traits = ["FontChooserExtManual"]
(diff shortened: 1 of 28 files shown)


sdroege (approved):
Looks good to me

bilelmoussaoui:
Thank you, I will go ahead and merge it then. We can improve things in the future if someone complains :)

PR #300, opened 2021-04-15
Address Sebastian's signal comments

### LINKED ISSUE #258: Signals comments
- Maybe mention the difference between `connect` and `connect_local`, and that the properly typed connector functions take care of this distinction themselves. Might be just a footnote
 - `let button = args.get(0).unwrap().get::<Button>().unwrap().unwrap();` explain the `unwrap`s here like for properties, and also why not `args[0]`? It's a slice after all :) Maybe also mention that the first argument to signals is always the object that emitted the signal, and that this specific signal has no other arguments
 - "// Types of the values which will be sent to the receiver" (and below that): Receiver would be better called "signal handler" as that's the GObject terminology
 - You do it correct, but it's worth a note in the `clicked` function that you should really never ever keep a mutex (or refcell) locked/borrowed while: emitting signals, notifying property changes, anything that can cause these to happen. Because that will call into external code, and that external code might call back into your object, trying to use the same mutex/refcell and then you have a problem :)

Hofer-Julian:
Fixes #258

### DIFF
--- a/book/listings/gobject_signals/3/custom_button/imp.rs
+++ b/book/listings/gobject_signals/3/custom_button/imp.rs
@@ -33,9 +33,9 @@ impl ObjectImpl for CustomButton {
             vec![Signal::builder(
                 // Signal name
                 "max-number-reached",
-                // Types of the values which will be sent to the receiver
+                // Types of the values which will be sent to the signal handler
                 &[i32::static_type().into()],
-                // Type of the value the receiver sends back
+                // Type of the value the signal handler sends back
                 <()>::static_type().into(),
             )
             .build()]
--- a/book/listings/gobject_signals/3/custom_button/imp.rs
+++ b/book/listings/gobject_signals/3/custom_button/imp.rs
@@ -72,7 +72,7 @@ impl ObjectImpl for CustomButton {
         }
     }
 
-    fn get_property(&self, _obj: &Self::Type, _id: usize, pspec: &ParamSpec) -> Value {
+    fn property(&self, _obj: &Self::Type, _id: usize, pspec: &ParamSpec) -> Value {
         match pspec.name() {
             "number" => self.number.borrow().to_value(),
             _ => unimplemented!(),
--- a/book/listings/gobject_signals/4/custom_button/imp.rs
+++ b/book/listings/gobject_signals/4/custom_button/imp.rs
@@ -102,10 +102,11 @@ impl ButtonImpl for CustomButton {
         // If `number` reached `MAX_NUMBER`,
         // emit "max-number-reached" signal and set `number` back to 0
         if incremented_number == MAX_NUMBER {
+            let borrowed_number = self.number.borrow();
             button
-                .emit_by_name("max-number-reached", &[&incremented_number])
+                .emit_by_name("max-number-reached", &[&(*borrowed_number + 1)])
                 .unwrap();
-            button.set_property("number", &0).unwrap();
+            // Only here `borrowed_number` gets dropped
         } else {
             button.set_property("number", &incremented_number).unwrap();
         }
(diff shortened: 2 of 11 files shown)


Hofer-Julian:
>You do it correct, but it's worth a note in the clicked function that you should really never ever keep a mutex (or refcell) locked/borrowed while: emitting signals, notifying property changes, anything that can cause these to happen. Because that will call into external code, and that external code might call back into your object, trying to use the same mutex/refcell and then you have a problem :)

I tried to play a bit, to modify the existing example to nicely demonstrate this problem.
I couldn't reach anything acceptable.

@sdroege maybe you can think of an example, or alternatively a nice explanation using words only  :)

sdroege:
Call something on the object from inside the signal handler that also needs to borrow the state. Like printing the current counter value or something.

Hofer-Julian:
I gave it a try

book/src/gobject_signals.md
```diff
@@ -59,6 +59,44 @@ Whenever we now receive the "max-number-reached" signal, the accompanying number
...
+At `clicked` we can now stop setting the property after we emitted "max-number-reached".
+
+<span class="filename">Filename: listings/gobject_signals/4/custom_button/imp.rs</span>
+
+```rust,no_run
+{{#rustdoc_include ../listings/gobject_signals/4/custom_button/imp.rs:button_impl}}
+```
+
+Please also note, that we borrowed our number and bound it to `borrowed_number`.
+While it does not make much sense in our case, it is quite usual to bind locks of Mutexes or borrows of RefCells to variables.
```
sdroege, line 83: "locks" -> mutex guards?

sdroege:
Sounds good to me otherwise

book/src/gobject_signals.md
Hofer-Julian: Fixed

Hofer-Julian:
I think this one can be merged now.
Do you agree @sdroege?

sdroege (approved):

book/listings/gobject_signals/3/custom_button/imp.rs
```diff
@@ -23,27 +24,71 @@ impl ObjectSubclass for CustomButton {
...
+        match pspec.name() {
+            "number" => {
+                let input_number = value.get().unwrap().unwrap();
+                self.number.replace(input_number);
+            }
+            _ => unimplemented!(),
+        }
+    }
+
+    fn get_property(&self, _obj: &Self::Type, _id: usize, pspec: &ParamSpec) -> Value {
```
bilelmoussaoui, line 75: this needs to be renamed per the latest changes

book/listings/gobject_signals/4/custom_button/imp.rs
```diff
@@ -0,0 +1,115 @@
...
+        match pspec.name() {
+            "number" => {
+                let input_number = value.get().unwrap().unwrap();
+                self.number.replace(input_number);
+            }
+            _ => unimplemented!(),
+        }
+    }
+
+    fn get_property(&self, _obj: &Self::Type, _id: usize, pspec: &ParamSpec) -> Value {
```
bilelmoussaoui, line 75: Same here

book/listings/gobject_signals/3/custom_button/imp.rs
Hofer-Julian: Fixed

book/listings/gobject_signals/4/custom_button/imp.rs
Hofer-Julian: Fixed as well :)

bilelmoussaoui:
Thanks Julian!

PR #303, opened 2021-04-16
Get removal round 2

fengalin:
## Note to applications developers

Applications developers should use [`fix-getters-calls` 0.3.0](https://crates.io/crates/fix-getters-calls/0.3.0) to ease migration of their applications. If you also wish to change your get functions definition to comply with the standards implemented here, use [`fix-getters-def` 0.3.0](https://crates.io/crates/fix-getters-def/0.3.0).

## Object

This PR addresses two concerns identified during https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/639:

- `enum` getters generated automatically required specific renames in `Gir.toml` because they were consuming `self`.
- `get` elision was not applied in the `subclass` infrastructure.

This second round addresses other important concerns. See the details in this [gir PR](https://github.com/gtk-rs/gir/pull/1102). Also depends on https://github.com/gtk-rs/gtk-rs/pull/455.


gsk4/Gir.toml
```diff
@@ -90,6 +90,10 @@ cfg_condition = "feature = \"broadway\""
 name = "Gsk.GLShader"
 status = "generate"
 generate_builder = true
+    [[object.function]]
+    name = "get_arg_bool"
```
bilelmoussaoui, line 94: Why this one needs an explicit rename but not the other ones such as arg_int?

gtk4/src/auto/stack_page.rs
```diff
@@ -42,7 +42,7 @@ impl StackPage {
     }
 
     #[doc(alias = "gtk_stack_page_get_needs_attention")]
-    pub fn is_needs_attention(&self) -> bool {
+    pub fn needs_attention(&self) -> bool {
```
bilelmoussaoui, line 45: Nice! :) i really like the new name

fengalin: Thanks :)

fengalin: BTW, this is an automatic conversion on behalf of `fix-getters-rules` which is also automatically detected on the call sites when migrating code using `fix-getters-calls`.

gsk4/Gir.toml
fengalin: What I meant is that the return value is not an answer to "is this arg a boolean?", but the actual value of the `arg`.

bilelmoussaoui: Ah I see. Makes sense.

PR #307, opened 2021-04-19
gtk: bind BitsetIter manually

bilelmoussaoui:
cc @jsparber as you requested it few days ago

sadly the tests has to be disabled as expression's tests initialize gtk already and couldn't find a nice/simple to initialize gtk globally when tests are run or a way to run them on the main thread maybe? no idea how to handle it.

### DIFF
--- a/gtk4/src/bitset_iter.rs
+++ b/gtk4/src/bitset_iter.rs
@@ -138,40 +138,28 @@ impl<'a> ToGlibPtr<'a, *const ffi::GtkBitsetIter> for BitsetIter<'a> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::TEST_THREAD_WORKER;
 
     #[test]
     fn test_bitset_iter() {
-        crate::init().unwrap();
+        TEST_THREAD_WORKER
+            .push(move || {
+                let set = Bitset::new_range(0, 20);
+                let (mut iter, init_value) = BitsetIter::init_first(&set).unwrap();
+                assert_eq!(init_value, 0);
+                assert_eq!(iter.next(), Some(1));
+                assert_eq!(iter.previous(), Some(0));
 
-        let set = Bitset::new_range(0, 20);
-        let (mut iter, init_value) = BitsetIter::init_first(&set).unwrap();
-        assert_eq!(init_value, 0);
-        assert_eq!(iter.next(), Some(1));
-        assert_eq!(iter.previous(), Some(0));
-
-        let set2 = Bitset::new_range(0, 3);
-        let (mut iter, init_val) = BitsetIter::init_last(&set2).unwrap();
-        assert_eq!(init_val, 2);
-        assert_eq!(iter.previous(), Some(1));
-        assert_eq!(iter.previous(), Some(0));
-        assert_eq!(iter.previous(), None);
-        assert!(!iter.is_valid());
-        assert_eq!(iter.next(), Some(1));
-        assert!(iter.is_valid());
-    }
-
-    #[test]
-    fn test_bitset_iter_dropped_bitset() {
-        crate::init().unwrap();
-
-        let set = Bitset::new_range(0, 20);
-        let (mut iter, init_value) = BitsetIter::init_first(&set).unwrap();
-        unsafe {
-            ffi::gtk_bitset_unref(set.to_glib_none().0);
-        };
-        assert_eq!(init_value, 0);
-        assert_eq!(iter.next(), Some(1));
-        assert_eq!(iter.previous(), Some(0));
-        assert!(iter.is_valid());
+                let set2 = Bitset::new_range(0, 3);
+                let (mut iter, init_val) = BitsetIter::init_last(&set2).unwrap();
+                assert_eq!(init_val, 2);
+                assert_eq!(iter.previous(), Some(1));
+                assert_eq!(iter.previous(), Some(0));
+                assert_eq!(iter.previous(), None);
+                assert!(!iter.is_valid());
+                assert_eq!(iter.next(), Some(1));
+                assert!(iter.is_valid());
+            })
+            .expect("Failed to schedule a test call");
     }
 }
--- a/gtk4/src/bitset_iter.rs
+++ b/gtk4/src/bitset_iter.rs
@@ -1,0 +1,177 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::Bitset;
+use glib::translate::*;
+use std::marker::PhantomData;
+
+#[derive(Copy, Clone)]
+pub struct BitsetIter<'a>(ffi::GtkBitsetIter, PhantomData<&'a Bitset>);
+
+impl<'a> BitsetIter<'a> {
+    #[doc(alias = "gtk_bitset_iter_init_at")]
+    pub fn init_at(set: &'a Bitset, target: u32) -> Option<(Self, u32)> {
+        assert_initialized_main_thread!();
+        unsafe {
+            let mut iter = std::mem::MaybeUninit::uninit();
+            let mut value = std::mem::MaybeUninit::uninit();
+            let found: bool = from_glib(ffi::gtk_bitset_iter_init_at(
+                iter.as_mut_ptr(),
+                set.to_glib_none().0,
+                target,
+                value.as_mut_ptr(),
+            ));
+
+            if found {
+                Some((
+                    BitsetIter(iter.assume_init(), PhantomData),
+                    value.assume_init(),
+                ))
+            } else {
+                None
+            }
+        }
+    }
+
+    #[doc(alias = "gtk_bitset_iter_init_first")]
+    pub fn init_first(set: &'a Bitset) -> Option<(Self, u32)> {
+        assert_initialized_main_thread!();
+        unsafe {
+            let mut iter = std::mem::MaybeUninit::uninit();
+            let mut value = std::mem::MaybeUninit::uninit();
+            let found: bool = from_glib(ffi::gtk_bitset_iter_init_first(
+                iter.as_mut_ptr(),
+                set.to_glib_none().0,
+                value.as_mut_ptr(),
+            ));
+            if found {
+                Some((
+                    BitsetIter(iter.assume_init(), PhantomData),
+                    value.assume_init(),
+                ))
+            } else {
+                None
+            }
+        }
+    }
+
+    #[doc(alias = "gtk_bitset_iter_init_last")]
+    pub fn init_last(set: &'a Bitset) -> Option<(Self, u32)> {
+        assert_initialized_main_thread!();
+        unsafe {
+            let mut iter = std::mem::MaybeUninit::uninit();
+            let mut value = std::mem::MaybeUninit::uninit();
+            let found: bool = from_glib(ffi::gtk_bitset_iter_init_last(
+                iter.as_mut_ptr(),
+                set.to_glib_none().0,
+                value.as_mut_ptr(),
+            ));
+            if found {
+                Some((
+                    BitsetIter(iter.assume_init(), PhantomData),
+                    value.assume_init(),
+                ))
+            } else {
+                None
+            }
+        }
+    }
+
+    #[doc(alias = "gtk_bitset_iter_get_value")]
+    pub fn value(&self) -> u32 {
+        unsafe { ffi::gtk_bitset_iter_get_value(self.to_glib_none().0) }
+    }
+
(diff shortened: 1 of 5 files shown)


sdroege:
There are a few things in the build logs to take care of:

```
 warning: dereferencing a null pointer
  --> gtk4/src/subclass/scrollable.rs:62:22
   |
62 |         *borderptr = *std::ptr::null();
   |                      ^^^^^^^^^^^^^^^^^ this code causes undefined behavior when executed
   |
   = note: `#[warn(deref_nullptr)]` on by default

warning: dereferencing a null pointer
  --> gtk4/src/subclass/scrollable.rs:62:22
   |
62 |         *borderptr = *std::ptr::null();
   |                      ^^^^^^^^^^^^^^^^^ this code causes undefined behavior when executed
   |
   = note: `#[warn(deref_nullptr)]` on by default
```
That one is unrelated to this PR but looks quite important :)

```
 ---- expression::tests::test_expressions stdout ----
thread 'expression::tests::test_expressions' panicked at 'called `Result::unwrap()` on an `Err` value: BoolError { message: "Failed to acquire default main context", filename: "gtk4/src/rt.rs", function: "gtk4::rt", line: 103 }', gtk4/src/expression.rs:457:23
```

Also unrelated to this PR but seems worrying.

```
---- bitset_iter::tests::test_bitset_iter stdout ----
thread 'bitset_iter::tests::test_bitset_iter' panicked at 'called `Result::unwrap()` on an `Err` value: BoolError { message: "Failed to acquire default main context", filename: "gtk4/src/rt.rs", function: "gtk4::rt", line: 103 }', gtk4/src/bitset_iter.rs:149:23
```

And this is related to this PR but the same error as above.

gtk4/src/bitset_iter.rs
```diff
@@ -0,0 +1,167 @@
...
+        unsafe { ffi::gtk_bitset_iter_get_value(self.to_glib_none().0) }
+    }
+
+    #[doc(alias = "gtk_bitset_iter_is_valid")]
+    pub fn is_valid(&self) -> bool {
+        unsafe { from_glib(ffi::gtk_bitset_iter_is_valid(self.to_glib_none().0)) }
+    }
+
+    #[doc(alias = "gtk_bitset_iter_next")]
+    pub fn next(&mut self) -> Option<u32> {
```
sdroege, line 79: This looks like it should implement `Iterator`. But not `DoubleEndedIterator` because `previous()` seems to behave difference

gtk4/src/bitset_iter.rs
```diff
@@ -0,0 +1,167 @@
...
+
+use crate::Bitset;
+use glib::translate::*;
+
+#[derive(Copy, Clone)]
+pub struct BitsetIter(ffi::GtkBitsetIter);
+
+impl BitsetIter {
+    #[doc(alias = "gtk_bitset_iter_init_at")]
+    pub fn init_at(set: &Bitset, target: u32) -> Option<(Self, u32)> {
```
sdroege, line 11: These constructor names are a bit strange

sdroege (commented):
Otherwise the code looks ok :)

sdroege:
> That one is unrelated to this PR but looks quite important :)
> 
> ```
>  ---- expression::tests::test_expressions stdout ----
> thread 'expression::tests::test_expressions' panicked at 'called `Result::unwrap()` on an `Err` value: BoolError { message: "Failed to acquire default main context", filename: "gtk4/src/rt.rs", function: "gtk4::rt", line: 103 }', gtk4/src/expression.rs:457:23
> ```
> 
> Also unrelated to this PR but seems worrying.
> 
> ```
> ---- bitset_iter::tests::test_bitset_iter stdout ----
> thread 'bitset_iter::tests::test_bitset_iter' panicked at 'called `Result::unwrap()` on an `Err` value: BoolError { message: "Failed to acquire default main context", filename: "gtk4/src/rt.rs", function: "gtk4::rt", line: 103 }', gtk4/src/bitset_iter.rs:149:23
> ```
> 
> And this is related to this PR but the same error as above.

These ones might happen because you initialize GTK in each of the tests (i.e. in multiple threads at once) but it can only really be initialized once and only from the main thread

gtk4/src/bitset_iter.rs
bilelmoussaoui: No ideas how to make them any better :-)

bilelmoussaoui:
> These ones might happen because you initialize GTK in each of the tests (i.e. in multiple threads at once) but it can only really be initialized once and only from the main thread

Yes, but I don't know if it's possible to run both tests on different threads if both of them want to be run on the main thread. Should I use something specific in such case or just ignore the tests?

gtk4/src/bitset_iter.rs
sdroege: Me neither. I'm not good at coming up with good names ;)

sdroege:
> Yes, but I don't know if it's possible to run both tests on different threads if both of them want to be run on the main thread. Should I use something specific in such case or just ignore the tests?

I think you'll need to run with `--test-threads 1` in the CI. IIRC there was also a crate that enforced that from inside the tests, we might be even using that in one of our crates somewhere.

gtk4/src/bitset_iter.rs
bilelmoussaoui: Let's leave it as is then :-)

bilelmoussaoui:
> 
```
 warning: dereferencing a null pointer
  --> gtk4/src/subclass/scrollable.rs:62:22
   |
62 |         *borderptr = *std::ptr::null();
   |                      ^^^^^^^^^^^^^^^^^ this code causes undefined behavior when executed
```

Regarding this one, I suppose the way forward is to just not set a the border to null as we can't really do that due to not having ` *mut *mut ffi::GtkBorder` as an input param, but instead we have `*mut ffi::GtkBorder`

sdroege:
> Regarding this one, I suppose the way forward is to just not set a the border to null as we can't really do that due to not having ` *mut *mut ffi::GtkBorder` as an input param, but instead we have `*mut ffi::GtkBorder`

Well, I think you'd either change it to not set the `borderptr` at all on errors (valid according to the C API contract) or to set a "default" border value instead (say, `gtk4_sys::GtkBorder { 0, 0, 0, 0}`). The border is returned by value, not as pointer.

gtk4/src/bitset_iter.rs
sdroege: clippy doesn't like this still

```
error: method `next` can be confused for the standard trait method `std::iter::Iterator::next`
  --> gtk4/src/bitset_iter.rs:79:5
   |
79 | /     pub fn next(&mut self) -> Option<u32> {
80 | |         unsafe {
81 | |             let mut value = std::mem::MaybeUninit::uninit();
82 | |
...  |
92 | |         }
93 | |     }
   | |_____^
   |
   = note: `-D clippy::should-implement-trait` implied by `-D warnings`
   = help: consider implementing the trait `std::iter::Iterator` or choosing a less ambiguous method name
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#should_implement_trait

```

.github/workflows/CI.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           override: true
 
       - name: "${{ matrix.conf.name }}: tests"
-        run: xvfb-run --auto-servernum cargo test --manifest-path ${{ matrix.conf.name }}/Cargo.toml --features "${{ matrix.conf.features }}"
+        run: xvfb-run --auto-servernum cargo test --manifest-path ${{ matrix.conf.name }}/Cargo.toml --features "${{ matrix.conf.features }}" -- --test-threads 1
```
sdroege, line 35: Could we use `serial_test` like in `gio` for this instead?

gtk4/src/bitset_iter.rs
```diff
@@ -0,0 +1,171 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::Bitset;
+use glib::translate::*;
+
+#[derive(Copy, Clone)]
+pub struct BitsetIter(ffi::GtkBitsetIter);
```
sdroege, line 7:
```suggestion
pub struct BitsetIter<'a>(ffi::GtkBitsetIter, &'a Bitset);
```

It probably has to borrow the bitset or not? If the bitset disappears the iterator will probably fall apart?

sdroege:
```suggestion
pub struct BitsetIter(ffi::GtkBitsetIter, PhantomData<&'a Bitset>);
```

sdroege: Use `PhantomData` if you don't need to actually access the bitset

bilelmoussaoui: > It probably has to borrow the bitset or not? If the bitset disappears the iterator will probably fall apart? 

aha, I will give it a try and see

.github/workflows/CI.yml
bilelmoussaoui: That sadly doesn't work, despite tests being run one after another, they are still run in different threads
```
thread 'expression::tests::test_expressions' panicked at 'Attempted to initialize GTK from two different threads.', gtk4/src/rt.rs:98:9
```
and I don't think there's a way to "uninitialize"  gtk at the end of each test so that the test run after can work.

sdroege: You could maybe run a single worker thread that initializes GTK and then from each test only queue up a job there to run the test.

gtk4/src/bitset_iter.rs
bilelmoussaoui: So we need to keep a reference of the Bitset around otherwise the iterator falls apart. I went ahead and used the PhantomData as you suggested, but that means we can't implement `Iterator` due to this difference
```
    = note: expected fn pointer `fn(&mut BitsetIter<'a>) -> Option<_>`
               found fn pointer `fn(&'a mut BitsetIter<'a>) -> Option<_>`
```
any ideas?

bilelmoussaoui: Never mind, fixed it :) I have added a test case for that, but it still fails if the bitset is dropped

gtk4/src/bitset_iter.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    #[test]
+    fn test_bitset_iter_dropped_bitset() {
+        crate::init().unwrap();
+
+        let set = Bitset::new_range(0, 20);
+        let (mut iter, init_value) = BitsetIter::init_first(&set).unwrap();
+        unsafe {
+            ffi::gtk_bitset_unref(set.to_glib_none().0);
```
sdroege, line 170: Well of course this doesn't work. You're using `unsafe` code here to force dropping the bitset so the compiler can't protect you :)

If you'd try using `drop(set)` instead, the compiler would prevent you from breaking things.

sdroege (approved):
Looks good to me apart from the test

gtk4/src/bitset_iter.rs
bilelmoussaoui: aha interesting
```
error[E0505]: cannot move out of `set` because it is borrowed
   --> gtk4/src/bitset_iter.rs:169:14
    |
168 |         let (mut iter, init_value) = BitsetIter::init_first(&set).unwrap();
    |                                                             ---- borrow of `set` occurs here
169 |         drop(set);
    |              ^^^ move out of `set` occurs here
170 |         assert_eq!(init_value, 0);
171 |         assert_eq!(iter.next(), Some(1));
    |                    ---- borrow later used here
```
Nice! thanks for the tip, learned something new :)

.github/workflows/CI.yml
bilelmoussaoui: That worked fine :) thanks!

gtk4/src/lib.rs
```diff
@@ -206,6 +206,18 @@ pub const TEXT_VIEW_PRIORITY_VALIDATE: u32 = ffi::GTK_TEXT_VIEW_PRIORITY_VALIDAT
 #[macro_use]
 mod rt;
 
+#[allow(dead_code)]
+#[cfg(test)]
+pub(crate) static TEST_THREAD_WORKER: once_cell::sync::Lazy<glib::ThreadPool> =
+    once_cell::sync::Lazy::new(|| {
```
sdroege, line 212: I believe this won't work on macOS (UI things have to happen from the thread that called `main()`), but otherwise this should be OK.

sdroege (approved):

gtk4/src/lib.rs
bilelmoussaoui: interesting. well, i guess until we have a CI for macOS we can live with this :)

PR #308, opened 2021-04-19
bump proc-macro-crate

bilelmoussaoui:


sdroege:
See https://github.com/gtk-rs/gtk-rs/pull/466 for the fixes needed

bilelmoussaoui:
hmm, i will borrow those as well, thanks!

sdroege:
You will need to make the examples and book `Cargo.toml` to depend on `glib` so that the `clone!` etc macros can be used. Otherwise it can't figure out the name of the `glib` dependency (there is none).

bilelmoussaoui:
Yes, I just was still hoping there's some kind of possible workaround when a crate re-exports a dependency. It's a bit weird to only have an explicit dependency on glib when all the other crates are implicitly there.

sdroege:
> Yes, I just was still hoping there's some kind of possible workaround when a crate re-exports a dependency. It's a bit weird to only have an explicit dependency on glib when all the other crates are implicitly there.

The alternative would be to guess that `glib` will be in scope with that name if it can't be found, which seems a bit fragile. Note that you'll also need `use gtk::glib` to make this work, which is not very obvious to a user either.

PR #310, opened 2021-04-19
Remove 'work in progress' statement

sophie-h:
gtk-rs/release#147


bilelmoussaoui:
The build is failing due to the change in https://github.com/gtk-rs/gtk-rs/pull/466#issuecomment-822839520, let's figure that one first before merging. Thanks!

PR #313, opened 2021-04-22
new generator & fancy updated docs

bilelmoussaoui:
cc @MarijnS95


.github/workflows/CI.yml
```diff
@@ -87,7 +87,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: "recursive"
-      - run: python3 generator.py --yes
+      - run: python3 ./gir/generator.py --yes ../
```
MarijnS95, line 90: We created a symlink in the root of `gstreamer-rs` but this works just as well :)

bilelmoussaoui: Ah that's a much better idea actually :) Thanks!

MarijnS95 (approved):
Looks good to me :)

PR #314, opened 2021-04-23
gdk: make ContentProvider subclassable

elmarco:


elmarco:
Looking for advices to bind the async function: https://gitlab.gnome.org/GNOME/gtk/-/blob/master/gdk/gdkcontentprovider.h#L68

bilelmoussaoui:
> Looking for advices to bind the async function: https://gitlab.gnome.org/GNOME/gtk/-/blob/master/gdk/gdkcontentprovider.h#L68

I don't have any prior experience with that, i didn't find examples neither when I have tried to add `gio::LoadableIcon` subclassing support. See https://github.com/gtk-rs/gtk-rs/pull/294#discussion_r571919489 for the tip that @sdroege gave me at that time. 

Also thanks a lot! It's the only missing object that I wanted to have as subclassable for the first release!

gdk4/src/subclass/content_provider.rs
```diff
@@ -0,0 +1,248 @@
...
+            let ret = f(provider
+                .unsafe_cast_ref::<ContentProvider>()
+                .to_glib_none()
+                .0);
+
+            from_glib_full(ret)
+        }
+    }
+
+    fn parent_value(&self, provider: &Self::Type, value: &mut Value) -> Result<(), glib::Error> {
```
bilelmoussaoui, line 130:
```suggestion
    fn parent_value(&self, provider: &Self::Type) -> Result<Value, glib::Error> {
```
Or the user has to initialize the value with a specific type before?

elmarco: Yes, the value is pre-initialized with the requested type.

sdroege:
Yeah the async variants will need some experimenting to see how to best wrap them :) We have no such example yet

gdk4/src/subclass/content_provider.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    fn attach_clipboard(&self, provider: &Self::Type, clipboard: &Clipboard) {
+        self.parent_attach_clipboard(provider, clipboard)
+    }
+
+    fn detach_clipboard(&self, provider: &Self::Type, clipboard: &Clipboard) {
+        self.parent_detach_clipboard(provider, clipboard)
+    }
+
+    fn ref_formats(&self, provider: &Self::Type) -> ContentFormats {
```
sdroege, line 21: Why is this called `ref` and not just `get`?

gdk4/src/subclass/content_provider.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    fn ref_storable_formats(&self, provider: &Self::Type) -> ContentFormats {
+        self.parent_ref_storable_formats(provider)
+    }
+
+    // fn write_mime_type_async(&self, provider: &Self::Type) -> ContentFormats {
+    //     self.parent_write_mime_type_async(provider)
+    // }
+
+    fn value(&self, provider: &Self::Type, value: &mut Value) -> Result<(), glib::Error> {
```
sdroege, line 33: Do you have some example usage in C somewhere with the implementation of this and a caller of `gdk_content_provider_get_value`? This seems a bit strange and the documentation is also not very helpful

gdk4/src/subclass/content_provider.rs
elmarco: Good question, I don't think there is any strong reason judging by the users in gdk/gtk.

gdk4/src/subclass/content_provider.rs
elmarco: See for ex: https://gitlab.gnome.org/GNOME/gtk/-/blob/master/gtk/gtklabel.c#L4545

The content provider checks the expected value type and fills it.

elmarco: And https://gitlab.gnome.org/GNOME/gtk/-/blob/master/gdk/gdkclipboard.c#L782

gdk4/src/subclass/content_provider.rs
sdroege: With the new naming, these would just become `fn formats()` and `fn parent_formats()` btw.

gdk4/src/subclass/content_provider.rs
sdroege: Should this then maybe become `fn value(&self, provider, type_: glib::Type) -> Result<glib::Value, glib::Error>`? That would be nicer to use from a Rust point of view at least.

The C API seems to be just too lazy to separate the in parameter (type) and out parameter (value), so both are put into the same.

gdk4/src/subclass/content_provider.rs
elmarco: ok

gdk4/src/subclass/content_provider.rs
elmarco: Yes, that's probably more convenient.

elmarco:
@sdroege I am on the right track to bind the write_mime_async? I have troubles with https://github.com/gtk-rs/gtk4-rs/pull/314/files#diff-1efbbb9b76d2e90020ab4ccc5818ea0d21f6543f4713b5e3ef0a3d55bf7f3d9bR234 and not so sure what I am doing

gdk4/src/subclass/content_provider.rs
```diff
@@ -0,0 +1,433 @@
...
+    });
+}
+
+unsafe extern "C" fn content_provider_write_mime_type_finish<T: ContentProviderImpl>(
+    _ptr: *mut ffi::GdkContentProvider,
+    res_ptr: *mut gio::ffi::GAsyncResult,
+    error_ptr: *mut *mut glib::ffi::GError,
+) -> glib::ffi::gboolean {
+    let res: gio::AsyncResult = from_glib_none(res_ptr);
+    let t = res.downcast_ref::<gio::Task>().unwrap();
```
sdroege, line 398: This means that we should probably statically enforce that the subclass actually uses a `gio::Task` here somehow, and that it's storing a `bool` or `GError` inside it.

gdk4/src/subclass/content_provider.rs
```diff
@@ -0,0 +1,433 @@
...
+
+    fn parent_write_mime_type_future<P: IsA<gio::OutputStream> + Clone + 'static>(
+        &self,
+        provider: &Self::Type,
+        mime_type: &str,
+        stream: &P,
+        io_priority: glib::Priority,
+    ) -> Pin<Box<dyn Future<Output = Result<(), glib::Error>> + 'static>> {
+        let stream = stream.clone();
+        Box::pin(gio::GioFuture::new(provider, move |obj, send| {
```
sdroege, line 234: You have to clone the provider and everything. You can only have `'static` types inside that closure.

However the `_future` variant of the async functions should only be in the non-subclass API. It's a wrapper around the `_async` functions for user convenience.

gdk4/src/subclass/content_provider.rs
```diff
@@ -0,0 +1,433 @@
...
+}
+
+unsafe extern "C" fn content_provider_write_mime_type_finish<T: ContentProviderImpl>(
+    _ptr: *mut ffi::GdkContentProvider,
+    res_ptr: *mut gio::ffi::GAsyncResult,
+    error_ptr: *mut *mut glib::ffi::GError,
+) -> glib::ffi::gboolean {
+    let res: gio::AsyncResult = from_glib_none(res_ptr);
+    let t = res.downcast_ref::<gio::Task>().unwrap();
+    let ret = t.propagate_value();
```
sdroege, line 399: You need to make it so that it either stores a `bool` or a `GError` on errors, so that the implementor can actually return a proper error on failures.

sdroege:
> I am on the right track to bind the write_mime_async?

Doesn't look too bad :) I added some comments in the commit

gdk4/src/subclass/content_provider.rs
elmarco: Yeha, it compiles. No idea if it works though :)

Right about parent_write_mime_type_future, but I thought ContentProviderImplExt was the right place. Where else?

gdk4/src/subclass/content_provider.rs
elmarco: Isn't it enforced by the code above in `content_provider_write_mime_type_async` ?

gdk4/src/subclass/content_provider.rs
elmarco: yes, should be ok now

gdk4/src/content_provider.rs
```diff
@@ -0,0 +1,29 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::ContentProvider;
+use glib::translate::*;
+use glib::IsA;
+
+pub trait ContentProviderExtManual {
```
bilelmoussaoui, line 7: you need to add the new trait to the prelude

elmarco: right, done

gdk4/src/subclass/content_provider.rs
sdroege: Indeed

gdk4/src/subclass/content_provider.rs
sdroege: Ah so you want the user to implement the futures version and not the callback-based one. I guess that works.

sdroege: Well, implement it somewhere and make sure that it actually works :)

bilelmoussaoui: Would be nice to have it as an example as well if possible

gdk4/src/auto/content_provider.rs
```diff
@@ -118,22 +115,6 @@ impl<O: IsA<ContentProvider>> ContentProviderExt for O {
...
-            );
-            if error.is_null() {
-                Ok(())
-            } else {
-                Err(from_glib_full(error))
-            }
-        }
-    }
-
     fn ref_formats(&self) -> Option<ContentFormats> {
```
bilelmoussaoui, line 118: as the subclassing code renames those two getters, they should be renamed here as well

bilelmoussaoui: The same thing for their nullability

gdk4/src/subclass/content_provider.rs
```diff
@@ -0,0 +1,445 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::{Clipboard, ContentFormats, ContentProvider};
+
+use glib::object::IsA;
+use glib::subclass::prelude::*;
```
bilelmoussaoui, line 6:
```suggestion
use crate::subclass::prelude::*;
```
to be consistent with the changes landed at #323

gdk4/src/auto/content_provider.rs
elmarco: It conflicts with the property getters. Which one should I ignore?

elmarco: I ignored the property getters.

gdk4/src/subclass/content_provider.rs
elmarco: ok

gdk4/src/subclass/content_provider.rs
elmarco: seems to do the job

gdk4/Gir.toml
```diff
@@ -240,15 +245,38 @@ status = "generate"
...
 [[object]]
 name = "Gdk.ContentProvider"
 status = "generate"
+manual_traits = ["ContentProviderExtManual"]
     [[object.function]]
     name = "new_typed"
     ignore = true # new_for_value is what the bindings are supposed to use
+    [[object.function]]
+    name = "get_value"
+    manual = true # to return Value
```
bilelmoussaoui, line 263:
```suggestion
    manual = true # to return Value
    doc_trait_name = "ContentProviderExtManual"
```
otherwise the manual function won't get documented

gtk4/examples/content_provider.rs
```diff
@@ -0,0 +1,95 @@
+use gtk::prelude::*;
```
bilelmoussaoui, line 1: The example needs an entry in `examples/Cargo.toml` otherwise it won't be tested as part of the CI, along with a README.md explaining very briefly what the example is about. See how it's handled for the other examples

gdk4/src/auto/content_provider.rs
sdroege: Are the property getters doing exactly the same? Then good that it conflicts ;) You only want one of the two, and the non-property getter is usually the better choice.

gtk4/examples/content_provider.rs
elmarco: I didn't put in examples/ but in gtk4/examples. Why wouldn't it be built? I think it's not generally useful, and it's more of a test actually, but it requires a display connection.

bilelmoussaoui: Well, it should be moved to examples instead of gtk4/examples so it's not shipped with the crate. The examples are used as a build test for the API/bindings to ensure things are kept working.

gdk4/Gir.toml
elmarco: ok

gtk4/examples/content_provider.rs
elmarco: ok, done

bilelmoussaoui (approved):

sdroege:
> #### Conflicting files
> `gdk4/src/auto/content_formats_builder.rs`
> `gdk4/src/auto/content_provider.rs`

Needs a rebase

gdk4/src/content_formats_builder.rs
```diff
@@ -0,0 +1,48 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::{ContentFormats, ContentFormatsBuilder};
+use glib::translate::*;
+
+impl ContentFormatsBuilder {
+    #[doc(alias = "gdk_content_formats_builder_add_formats")]
+    pub fn add_formats(self, formats: &ContentFormats) -> Self {
```
sdroege, line 8: Rust builders usually don't use the `add_` prefix on functions. Also why are all these manual?

sdroege (commented):
Looks good to me otherwise

gdk4/src/subclass/content_provider.rs
```diff
@@ -0,0 +1,445 @@
...
+
+    fn storable_formats(&self, provider: &Self::Type) -> ContentFormats {
+        self.parent_storable_formats(provider)
+    }
+
+    fn write_mime_type_future<P: IsA<gio::OutputStream> + Clone + 'static>(
+        &self,
+        provider: &Self::Type,
+        mime_type: &str,
+        stream: &P,
```
sdroege, line 37:
```suggestion
        stream: &gio::OutputStream,
```

and remove `P` in all these functions. As this is passed in from C code, you will never ever have a different type here and having a concrete type is easier for user code implementing this.

gdk4/src/content_formats_builder.rs
bilelmoussaoui: If you drop the prefix, make sure to add another doc alias with the original function name `#[doc(alias ="add_formats")]`

elmarco: the functions actually add to a set of values, not a typical `member()` builder set

I made it manual mainly because I don't know how to tell gir to take self and -> Self. Is there a way?

gdk4/src/subclass/content_provider.rs
elmarco: make sense

gdk4/src/content_formats_builder.rs
bilelmoussaoui: > the functions actually add to a set of values, not a typical member() builder set

Ah I guess, let's keep them as is then

> I made it manual mainly because I don't know how to tell gir to take self and -> Self. Is there a way?

No, i don't think so

bilelmoussaoui:
Awesome contribution, thanks again! :)

PR #315, opened 2021-04-24
Cleanup leftovers from the remove-get-changes

Hofer-Julian:


book/src/settings.md
```diff
@@ -65,7 +65,7 @@ But we can make this even better.
 The `Switch` has a property "state" and `Settings` allows us to bind properties to a specific setting.
 So let us do exactly that.
 
-We can remove the `get_boolean` call before initializing the `Switch` as well as the `connect_state_set` call.
+We can remove the `boolean` call before initializing the `Switch` as well as the `connect_state_set` call.
```
bilelmoussaoui, line 68: Could use a link

Hofer-Julian: Mmmh, does it make sense if I *remove* ` boolean`?

I feel like links make the most sense if they are only introduced the first time a symbol gets introduced.
My feelings regarding that are not super strong though

bilelmoussaoui: I would definitely use a link here. Boolean sounds like you're talking about the type of data stored in the setting rather than a method name

Hofer-Julian: Good point, fixed!

PR #320, opened 2021-04-25
gtk: fix measure & add a layout manager example

### LINKED ISSUE #222: Investigate WidgetImpl::measure
The current function takes &mut, while the default WidgetExt::measure just returns `(i32, i32, i32, i32)`. We should check whether it makes sense to change `WidgetImpl::measure` signature or not.

bilelmoussaoui:
This PR fixes #222 and adds a custom layout manager example which is mostly a port from the gtk4-demo

screenshot


examples/custom_layout_manager/custom_layout.rs
```diff
@@ -0,0 +1,186 @@
+use crate::TOTAL_COLORS;
+use gtk::glib;
+use gtk::prelude::*;
+use gtk::subclass::prelude::*;
+
+const N_GRID_COLUMNS: u32 = 4;
+
+mod imp {
```
sdroege, line 8: Why not move this into a separate file instead of an inline module? :)

examples/custom_layout_manager/custom_layout.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub struct CustomLayout {
+        pub position: Cell<f64>,
+        pub child_pos: RefCell<Vec<i32>>,
+    }
+    impl Default for CustomLayout {
+        fn default() -> Self {
+            let mut positions = Vec::new();
+            for i in 0..TOTAL_COLORS {
+                positions.push(i);
+            }
```
sdroege, line 21:
```rust
let positions = (0..TOTAL_COLORS).collect::<Vec<_>>();
```
would be shorter :)

examples/custom_layout_manager/custom_layout.rs
```diff
@@ -0,0 +1,186 @@
...
+            let mut child = widget.first_child().unwrap();
+            loop {
+                if !child.should_layout() {
+                    continue;
+                }
+
+                let (child_min, child_nat, _, _) = child.measure(orientation, -1);
+                min_size = min_size.max(child_min);
+                nat_size = nat_size.max(child_nat);
+                if child.next_sibling().is_none() {
```
sdroege, line 65:
```rust
if let Some(next_child) = child.next_sibling() {
    child = next_child;
} else {
    break;
}
```

examples/custom_layout_manager/custom_layout.rs
```diff
@@ -0,0 +1,186 @@
...
+            let mut child = widget.first_child().unwrap();
+            loop {
+                if !child.should_layout() {
+                    continue;
+                }
+
+                let (child_req, _) = child.preferred_size();
+                child_width = child_width.max(child_req.width);
+                child_height = child_height.max(child_req.height);
+                if child.next_sibling().is_none() {
```
sdroege, line 101: Same here

examples/custom_layout_manager/custom_layout.rs
```diff
@@ -0,0 +1,186 @@
...
+                    &gtk::Allocation {
+                        x: x as i32,
+                        y: y as i32,
+                        width: child_width,
+                        height: child_height,
+                    },
+                    -1,
+                );
+                i += 1;
+                if child.next_sibling().is_none() {
```
sdroege, line 142: And here

examples/custom_layout_manager/custom_layout_child.rs
```diff
@@ -0,0 +1,69 @@
...
+glib::wrapper! {
+    pub struct CustomLayoutChild(ObjectSubclass<imp::CustomLayoutChild>)
+        @extends gtk::Widget;
+}
+
+impl CustomLayoutChild {
+    pub fn new(color: gdk::RGBA) -> Self {
+        let widget = glib::Object::new(&[]).expect("Failed to create CustomLayoutChild");
+        let inner = imp::CustomLayoutChild::from_instance(&widget);
+        inner.color.replace(color);
```
sdroege, line 66: This should really be a construct-only property :)

examples/custom_layout_manager/main.rs
```diff
@@ -0,0 +1,55 @@
...
+    "yellow",
+    "firebrick",
+    "aqua",
+    "purple",
+    "tomato",
+    "pink",
+    "thistle",
+    "maroon",
+];
+const TOTAL_COLORS: i32 = 16;
```
sdroege, line 27:
```suggestion
const TOTAL_COLORS: i32 = COLORS.len();
```

gtk4/src/subclass/layout_manager.rs
```diff
@@ -29,7 +29,10 @@ pub trait LayoutManagerImpl: LayoutManagerImplExt + ObjectImpl {
         self.parent_create_layout_child(layout_manager, widget, for_child)
     }
 
-    fn layout_child_type() -> glib::Type;
+    /// Only set if the child implemented LayoutChildImpl
+    fn layout_child_type() -> Option<glib::Type> {
+        None
```
sdroege, line 34: Should still call the parent one :)

sdroege (commented):
Looks good otherwise, nice example :)

examples/custom_layout_manager/custom_layout.rs
bilelmoussaoui: Because none of the examples in gtk4-rs does that and didn't want to have only one that's different than others. We should probably do that for all of them though but I wouldn't do it as part of this;

cc @Hofer-Julian as you have done something similar for gtk-rs

examples/custom_layout_manager/custom_layout.rs
bilelmoussaoui: Nice :) thanks!

examples/custom_layout_manager/custom_layout.rs
bilelmoussaoui: Oups, those are reasons to not write code that late at night :)

examples/custom_layout_manager/custom_layout_child.rs
bilelmoussaoui: Does it has to be a property though? :)

gtk4/src/subclass/layout_manager.rs
bilelmoussaoui: There's no parent one it's not a method but it sets the type on the Class struct with
```
        if let Some(type_) = T::layout_child_type() {
            klass.layout_child_type = type_.to_glib();
        }
```
Setting a `layout_child_type` to a widget instead of LayoutChild requires having `layout-child` property which is the widget; not something the user often has to do that's why it's None by default

examples/custom_layout_manager/custom_layout.rs
Hofer-Julian: I can adapt them.
Want to open an issue and assign it to me?

examples/custom_layout_manager/custom_layout_child.rs
bilelmoussaoui: Updated as we don't have an example with a boxed property :-)

examples/custom_layout_manager/custom_layout.rs
bilelmoussaoui: Done, thanks Julian! :)

examples/custom_layout_manager/custom_layout_child.rs
sdroege: > Does it has to be a property though? :)

Well, constructors should (ideally!) only call `glib::Object::new` and nothing else :)

sdroege (approved):

PR #323, opened 2021-04-25
move ExtManual traits to prelude only

bilelmoussaoui:


MarijnS95 (approved):
Nice, can't find any `Ext(Manual)?\b` in `lib.rs` anymore!

sdroege (approved):

gtk4/src/dialog.rs
```diff
@@ -1,6 +1,7 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use crate::{Dialog, DialogExt, DialogFlags, ResponseType, Widget, WidgetExt, Window};
```
sdroege, line 3: Actually it would be better to `use crate::prelude::*` here everywhere. That way all this manual code does not have to be updated if one of the functions in question is moved from an `Ext` to an `ExtManual` trait or the other way around.

gtk4/src/subclass/widget.rs
```diff
@@ -1,9 +1,9 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
+use crate::prelude::*;
 use crate::{
     AccessibleRole, BuilderScope, DirectionType, LayoutManager, Orientation, Shortcut,
     SizeRequestMode, Snapshot, StateFlags, SystemSetting, TextDirection, Tooltip, Widget,
-    WidgetExt,
 };
 use glib::prelude::*;
 use glib::subclass::{prelude::*, SignalId};
```
MarijnS95, line 9: glib prelude should already be pulled in by `crate::prelude::*`? And maybe the subclass prelude can be used as well through `crate::subclass::prelude::*`.

gtk4/src/subclass/root.rs
```diff
@@ -1,6 +1,6 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use crate::subclass::prelude::NativeImpl;
+use crate::subclass::prelude::*;
 use crate::Root;
 use glib::subclass::prelude::*;
```
MarijnS95, line 5: Should already be available from `use crate::subclass::prelude::*;`. Do check the other files, too :)

gtk4/src/subclass/widget.rs
bilelmoussaoui: Good point! let me update those :)

sdroege (approved):

sdroege:
`rustfmt` is angry

PR #324, opened 2021-04-25
examples: add a ListView: applications launcher

bilelmoussaoui:


examples/flow_box/main.rs
```diff
@@ -0,0 +1,61 @@
...
+use gtk::prelude::*;
+
+fn create_color_button(color: &'static str) -> gtk::Button {
+    let button = gtk::Button::new();
+    let drawing_area = gtk::DrawingAreaBuilder::new()
+        .content_height(24)
+        .content_width(24)
+        .build();
+    drawing_area.set_draw_func(move |_, cr, _width, _height| {
+        let rgba = gdk::RGBA::from_str(color).unwrap();
```
sdroege, line 14: This should probably happen outside the closure, once. Instead of inside the closure for each frame that is drawn :)

sdroege (commented):
Looks good to me otherwise

examples/flow_box/main.rs
bilelmoussaoui: Oh indeed.

sdroege (approved):

bilelmoussaoui:
I will go ahead and merge it despite the docs job failing, till rustup gets back to normal

PR #325, opened 2021-04-26
Accomodate changes in glib::Value

Hofer-Julian:
While the listings already build, the book itself had to be adapted.

### DIFF
--- a/book/listings/gobject_values/1/main.rs
+++ b/book/listings/gobject_values/1/main.rs
@@ -1,18 +1,12 @@
 use gtk::prelude::*;
 
 fn main() {
+    // Convert `i32` to `Value`
     let number_value = 10.to_value();
-    let string_value = "Hello!".to_value();
 
-    // Retrieve `i32` from `number_value`
+    // Retrieve `i32` again from `Value`
     let number = number_value
         .get::<i32>()
         .expect("The value needs to be of type `i32`.");
     assert_eq!(number, 10);
-
-    // Retrieve `String` from `string_value`
-    let string = string_value
-        .get::<String>()
-        .expect("The value needs to be of type `String`.");
-    assert_eq!(string, "Hello!");
 }
--- a/book/listings/gobject_values/1/main.rs
+++ b/book/listings/gobject_values/1/main.rs
@@ -9,6 +9,8 @@ fn main() {
     let integer = integer_value
         .get::<i32>()
         .expect("The value needs to be of type `i32`.");
+
+    // Check if the retrieved value is correct
     assert_eq!(integer, 10);
     // ANCHOR_END: i32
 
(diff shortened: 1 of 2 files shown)


book/src/gobject_values.md
```diff
@@ -5,25 +5,24 @@ Enums in C are not as powerful as the one in Rust, which is why [`glib::Value`](
...
+    bool,
+    i8,
+    i32,
+    u32,
+    i64,
+    u64,
+    f32,
+    f64,
+    String,
+    Object,
```
bilelmoussaoui, line 18: The old thing was correct, with this you can't really hold any value

sdroege (commented):
Otherwise looks good to me

book/src/gobject_values.md
Hofer-Julian: Yeah, I know.
I wanted to remove it in order to not introduce additional complexity.
Probably it is more confusing now, will change it back.

Hofer-Julian: I want to get rid of the Option though, any objections?
@sdroege @bilelmoussaoui

sdroege: Why? You can still store an `Option<String>` and that's what it is internally, but you can get a `String` from it that simply returns an `Err` if it happens to be `None`.

Hofer-Julian: I still feel this is detail because of the implementation in C.
You will not want to store Option<String> more often than Option<i32>.
But you only can store Option<String>, so this is definitely surprising.

Just mentioning String and Object seems to cover most use cases and the rest can be handled by the docs in my opinion.

sdroege: Maybe just skip `String` and `Object` then to not confuse :)

Hofer-Julian: I will need them later though :)
I pushed another commit which re-adds the original setting and explains it a bit more.

book/listings/gobject_values/1/main.rs
```diff
@@ -1,18 +1,36 @@
...
     assert_eq!(string, "Hello!");
+    // ANCHOR_END: string
+
+    // ANCHOR: string_none
+    // Store `Option<String>` as `Value`
+    let string_none_value = None::<String>.to_value();
+
+    // Retrieve `String` from `Value`
+    let string_none = string_none_value
+        .get::<Option<String>>()
```
sdroege, line 32: Maybe like this it's not clear that you could also do `string_value.get::<Option<String>>()`

book/src/gobject_values.md
```diff
@@ -17,16 +17,32 @@ enum Value <T> {
...
+If you are sure that your boxed type, will never be `None`, you can just access it the same way as with the `i32` above.
+
+<span class="filename">Filename: listings/gobject_values/1/main.rs</span>
+
+```rust,no_run
+{{#rustdoc_include ../listings/gobject_values/1/main.rs:string}}
+```
+
+In case your `Value` might be `None`, just call `get::<Option<T>>` instead.
+If you do that, it will give you the chance to check the output instead of panicking immediately.
```
sdroege, line 42: It's not panicking. It's returning an `Err` just like with type mismatches :)

Hofer-Julian: In my code example it would panic, because I call expect. 
I agree that the current wording is confusing though.

book/listings/gobject_values/1/main.rs
Hofer-Julian: Yeah maybe, suggestions are welcome.

sdroege: You could add that line to the example

Hofer-Julian: Done

book/src/gobject_values.md
Hofer-Julian: I gave it another try

sdroege (approved):

PR #329, opened 2021-04-27
gtk: bind functions::test_list_all_types

bilelmoussaoui:


gtk4/src/functions.rs
```diff
@@ -25,3 +25,12 @@ pub fn show_about_dialog<P: IsA<Window>>(parent: Option<&P>, properties: &[(&str
     about_dialog.set_transient_for(parent);
     about_dialog.show();
 }
+
+#[doc(alias = "gtk_test_list_all_types")]
```
bilelmoussaoui, line 29: I suppose this can't be generated automatically due to the `zero-terminated=1` annotation; although, not sure if it's really the case. see https://gitlab.gnome.org/GNOME/gtk/-/blob/master/gtk/gtktestutils.c#L152-158

gtk4/Gir.toml
```diff
@@ -440,6 +440,9 @@ trust_return_value_nullability = false
     [[object.function]]
     name = "enumerate_printers"
     cfg_condition = "target_os = \"linux\""
+    [[object.function]]
+    name = "test_list_all_types"
+    manual = true
```
sdroege, line 445: Is this function actually useful without the whole `gtk_test` API?

gtk4/src/functions.rs
sdroege: It's probably terminated with `G_TYPE_INVALID`, i.e. `0`.

But if there's a length parameter we should simply ignore the `zero-terminated` annotation

gtk4/Gir.toml
bilelmoussaoui: There are few missing testing API functions yes, but I plan to tackle those one after another :) some of them can't have simple bindings because they use some weird macros and have to be re-implemented manually.

bilelmoussaoui: After looking at the missing test functions, there are only 4 of them. `test_init` is the most important one from the list, the three others are accessibility related and require a manual macro re-implementation for them to be useful. I will go ahead and merge this one for now and figure out `test_init` in the future as I can't really test it locally due to the default build of gtk4/glib on Fedora

Issue #36, opened 2020-11-04
Port C examples


bilelmoussaoui:
It would be nice to port most if not all the examples in the upstream gtk repository to Rust as a replacement of the current examples that are mostly a copy paste of the gtk3 ones.

grantshandy:
I can do this, the upstream C `gtk` examples are fairly simple.

Some are formatted like a guide, (`application1`-`application9`), and I will place these in their own directory, each file is a separate bin. I will also add the upstream `README` files from each directory and add screenshots like the other examples.

I don't have much time to do this, so it may take around a week and a half to complete.

bilelmoussaoui:
@grantshandy Thanks for having a look at this! I would rather port one example after another to simplify the review process for us and we can get them merged much sooner.

grantshandy:
Sure that sounds great, I'll have each one as a separate pull request.

bilelmoussaoui:
Perfect! let me know if I can help somehow :)

mairacanal:
Hi! I have interest in working on this task, but I have a couple of questions. Is there interest in porting the application* examples? Also, examples such as `basics` are called `hello-world` on the C repo, should I rename them to make them more similar to the C examples?

I am excited to get to work on this, thanks!

Hofer-Julian:
@mairacanal great that you want to look into this.

> Also, examples such as basics are called hello-world on the C repo, should I rename them to make them more similar to the C examples?

IMO there is no need to change existing names to be closer to the GTK ones

>  Is there interest in porting the application* examples? 

Any example that is concise and isn't already covered by existing examples would be good to port.

bilelmoussaoui:
I think we are doing quiet good in terms of examples nowadays, more are always welcome though. But the main purpose of this issue is no longer that relevant so closing it.

PR #338, opened 2021-04-30
Add grid_packing example from the upstream C repository.

grantshandy:
This pull request is apart of #36, and is an adaptation of the [grid_packing](https://gitlab.gnome.org/GNOME/gtk/-/blob/master/examples/grid-packing.c) example from the upstream C repository.


examples/grid_packing/main.rs
```diff
@@ -0,0 +1,41 @@
+use gtk::glib::clone;
+use gtk::prelude::*;
+
+fn build_ui(application: &gtk::Application) {
+    // Create a new window, and set its title
+    let window = gtk::ApplicationWindow::new(application);
```
bilelmoussaoui, line 6: I would use ApplicationWindowBuilder & set a larger window default size, see other examples

examples/grid_packing/main.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    // Here we construct the grid that is going contain our buttons
+    let grid = gtk::Grid::new();
+
+    // Pack the grid in the window
+    window.set_child(Some(&grid));
+
+    // Create the first button and put it into the grid at (0, 0)
+    let button_1 = gtk::Button::with_label("Button 1");
+    button_1.connect_clicked(move |_| println!("Hello World"));
```
bilelmoussaoui, line 17:
```suggestion
    button_1.connect_clicked(move |_| eprintln!("Hello World"));
```

examples/grid_packing/main.rs
```diff
@@ -0,0 +1,41 @@
...
+    grid.attach(&button_1, 0, 0, 1, 1);
+
+    // Create the second button and put it into the grid at (1, 0)
+    let button_2 = gtk::Button::with_label("Button 2");
+    button_2.connect_clicked(move |_| println!("Hello World"));
```
bilelmoussaoui, line 22:
```suggestion
    button_2.connect_clicked(move |_| eprintln!("Hello World"));
```

examples/grid_packing/main.rs
```diff
@@ -0,0 +1,41 @@
+use gtk::glib::clone;
+use gtk::prelude::*;
+
+fn build_ui(application: &gtk::Application) {
+    // Create a new window, and set its title
+    let window = gtk::ApplicationWindow::new(application);
+    window.set_title(Some("Grid Packing"));
+
+    // Here we construct the grid that is going contain our buttons
+    let grid = gtk::Grid::new();
```
bilelmoussaoui, line 10: I would add some margin around the grid, so that the demo looks slightly nicer

examples/grid_packing/main.rs
grantshandy: Like in the original C example, the buttons and the grid don't expand with the size of the window. Should I add that too if we're going for a window that's bigger than the widgets inside?

bilelmoussaoui:
Also make sure to update the main readme in examples/README.md with the new example

examples/grid_packing/main.rs
grantshandy: "Hello World" is not an error, why print it to `stderr`?

examples/grid_packing/main.rs
bilelmoussaoui: Yeah, there's no need to expand the grid :) it would just look ugly

examples/grid_packing/main.rs
bilelmoussaoui: Oups, my bad

examples/grid_packing/main.rs
grantshandy: So you want me to have a larger window, but not have the widgets inside expand? It'll probably look something like this:
Screenshot from 2021-04-30 14-44-21

Are you sure we want this?

bilelmoussaoui: Set the valign/halign on the grid to center, so it's centered on the window

grantshandy: Okay I'll do that

bilelmoussaoui: And probably set column/row spacing so that the items don't look so close to each other. That would demo another aspect of GtkGrid I guess

grantshandy:
@bilelmoussaoui It should all be done now.

examples/grid_packing/main.rs
```diff
@@ -0,0 +1,55 @@
...
+    // Create the quit button, make it expand and put it into the grid at (0, 1)
+    let quit_button = gtk::Button::with_label("Quit");
+    quit_button.connect_clicked(clone!(@weak window => move |_| window.destroy()));
+
+    grid.attach(&quit_button, 0, 1, 2, 1);
+
+    window.show();
+}
+
+fn main() {
```
bilelmoussaoui, line 47: Other examples puts the main function on the top of the file, let's keep that consistent :)

examples/grid_packing/README.md
```diff
@@ -0,0 +1,5 @@
+# Grid Packing
+
+This example demonstrates how to attach widgets to a grid.
```
bilelmoussaoui, line 3: replace grid with `gtk::Grid`

bilelmoussaoui:
I left few nitpicks :) otherwise looks good to me! thanks

grantshandy:
@bilelmoussaoui Added the `GridBuilder`, spacing, and changed the example README, everything should look good here.

examples/grid_packing/main.rs
bilelmoussaoui: Make sure to update the screenshot afterwards

grantshandy: Column and row spacing as well as an updated screenshot should be there now

examples/grid_packing/main.rs
```diff
@@ -0,0 +1,54 @@
...
+        .halign(gtk::Align::Center)
+        .valign(gtk::Align::Center)
+        .row_spacing(6)
+        .column_spacing(6)
+        .build();
+
+    // Add the grid in the window
+    window.set_child(Some(&grid));
+
+    // Create the first button, make it expand and put it into the grid at (0, 0)
```
bilelmoussaoui, line 35:
```suggestion
    // Create the first button and put it into the grid at (0, 0)
```
You don't really make any of the buttons expand? the same thing about the two other buttons

grantshandy:
That should fix it

bilelmoussaoui:
Thanks!

PR #344, opened 2021-05-01
gdk: fix ContentProvider::write_mime_type_async

elmarco:
The spawned closure doesn't have a reference to the stream, we must
therefore give it a clone() to ensure it has a valid reference when
called. I believe the other objects are being referenced implictely by
the task, or the closure, for the instance. I am not so clear about the
`Borrowed<GString>`, it might need a similar treatement.


sdroege (commented):
Yeah this is all wrong. You must not move `from_glib_borrowed()` return values out of its scope, that's causing use-after-free and similar issues.

That is, you must not use it at all here and use `from_glib_none()` (which basically does that same as cloning) for everything that is moved into the async block.

sdroege (approved):

PR #346, opened 2021-05-01
gtk: add getters/setters for Border

bilelmoussaoui:
Instead of having to go through DerefMut/Deref


bilelmoussaoui:
The setters are still not great to use when subclassing Scrollable/Range. Maybe a Builder pattern for Border? :)

sdroege (approved):

PR #347, opened 2021-05-02
gtk: add FontChooser/TreeModelFilter subclassing support

bilelmoussaoui:
split out of #169 as the tree model stuff needs more thinking

part of #56

### DIFF
--- a/gtk4/src/subclass/font_chooser.rs
+++ b/gtk4/src/subclass/font_chooser.rs
@@ -1,0 +1,375 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::subclass::prelude::*;
+use crate::FontChooser;
+use glib::translate::*;
+use glib::{Cast, GString, IsA, ObjectExt, Quark};
+use once_cell::sync::Lazy;
+use pango::{FontFace, FontFamily, FontMap};
+
+#[derive(Debug)]
+pub struct FilterCallback {
+    filter_func: ffi::GtkFontFilterFunc,
+    user_data: glib::ffi::gpointer,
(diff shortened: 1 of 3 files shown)


gtk4/src/subclass/font_chooser.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    fn font_face(&self, font_chooser: &Self::Type) -> Option<FontFace> {
+        self.parent_font_face(font_chooser)
+    }
+
+    fn font_size(&self, font_chooser: &Self::Type) -> i32 {
+        self.parent_font_size(font_chooser)
+    }
+
+    fn set_filter_func(&self, font_chooser: &Self::Type, callback: Option<&FilterCallback>) {
```
sdroege, line 57:
```suggestion
    fn set_filter_func(&self, font_chooser: &Self::Type, callback: Option<FilterCallback>) {
```

Somehow the impl needs to store it, which is not possible if you just pass around a reference here.

gtk4/src/subclass/font_chooser.rs
```diff
@@ -0,0 +1,375 @@
...
+                    .to_glib_none()
+                    .0)
+            } else {
+                // No font size is selected
+                -1
+            }
+        }
+    }
+
+    fn parent_set_filter_func(&self, font_chooser: &Self::Type, callback: Option<&FilterCallback>) {
```
sdroege, line 136: And this needs to consume the callback (i.e. also no reference) so that you don't get a double free

sdroege (commented):
Seems good otherwise

PR #349, opened 2021-05-02
Regen with latest gir

bilelmoussaoui:


gdk4-x11/src/auto/enums.rs
```diff
@@ -23,11 +23,11 @@ impl fmt::Display for X11DeviceType {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(
             f,
-            "X11DeviceType::{}",
+            "Self::{}",
```
MarijnS95, line 26: Wait what?!?!

I guess I'll have to fix that :sweat_smile:

PR #361, opened 2021-05-05
book: Scalable lists chapter

Hofer-Julian:
- Add the scalable lists chapter
- Change RefCell to Cell as requested by Bilal (that also means that part of the signals chapter made no sense anymore)


book/listings/scalable_lists/1/main.rs
```diff
@@ -0,0 +1,96 @@
...
+            .unwrap()
+            .downcast::<IntegerObject>()
+            .unwrap();
+
+        let label = list_item.child().unwrap().downcast::<Label>().unwrap();
+        integer_object
+            .bind_property("number", &label, "label")
+            .flags(BindingFlags::SYNC_CREATE)
+            .build();
+    });
```
Hofer-Julian, line 51: @bilelmoussaoui maybe you can give me a hint, so I don't go in the wrong direction.
Goal: List of integers, if I double click on one of them it increases by 1.
Questions:
- Is binding properties the way to go?
- If yes, what is the typical way to unbind them again (I guess in the unbind step)?

book/src/saving_window_state.md
```diff
@@ -20,6 +20,10 @@ First, we create one and add methods for getting and setting the window state.
 
 ```rust,no_run
 {{#rustdoc_include ../listings/saving_window_state/1/custom_window/mod.rs:mod}}
+
+# // Please ignore this line
+# // It is only there to make rustdoc happy
```
bilelmoussaoui, line 25: why rustdoc is even run on the book ?

book/listings/scalable_lists/2/main.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        let old_number = integer_object
+            .property("number")
+            .expect("The property needs to exist and be readable.")
+            .get::<i32>()
+            .expect("The property needs to be of type `i32`.");
+
+        integer_object
+            .set_property("number", old_number + 1)
+            .unwrap();
```
bilelmoussaoui, line 91: add a method that increments the the number in your IntegerObject

book/listings/scalable_lists/2/integer_object/imp.rs
```diff
@@ -0,0 +1,69 @@
...
+use gtk::prelude::*;
+use gtk::subclass::prelude::*;
+use once_cell::sync::Lazy;
+
+use std::cell::RefCell;
+
+// Object holding the state
+#[derive(Default)]
+pub struct IntegerObject {
+    number: RefCell<i32>,
```
bilelmoussaoui, line 12:
```suggestion
    number: Cell<i32>,
```
for Copy types :-)

book/src/saving_window_state.md
Hofer-Julian: I think it just mimics the behavior of rustdoc. It wraps every code block without main in a new main.

book/listings/scalable_lists/2/main.rs
Hofer-Julian: Yeah, that's probably better

book/listings/scalable_lists/2/integer_object/imp.rs
Hofer-Julian: I had that originally in the Subclassing chapter, but changed it back so I won't introduce another Cell type. Is there a good reason why I should introduce Cell within this book?

book/listings/scalable_lists/1/integer_object/imp.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn property(&self, _obj: &Self::Type, _id: usize, pspec: &ParamSpec) -> Value {
+        match pspec.name() {
+            "number" => self.number.borrow().to_value(),
+            _ => unimplemented!(),
+        }
+    }
+
+    fn constructed(&self, obj: &Self::Type) {
+        self.parent_constructed(obj);
+    }
```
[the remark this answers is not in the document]
Hofer-Julian: Ah, that's not necessary?
Good to know!

book/listings/scalable_lists/1/integer_object/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+glib::wrapper! {
+    pub struct IntegerObject(ObjectSubclass<imp::IntegerObject>);
+}
+
+impl IntegerObject {
+    pub fn new() -> Self {
+        Object::new(&[]).expect("Failed to create IntegerObject")
+    }
+
+    pub fn from_integer(number: i32) -> Self {
```
sdroege, line 17:
```suggestion
    impl From<i32> for IntegerObject { ... }`
```

book/listings/scalable_lists/1/integer_object/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+        let integer_object = Self::new();
+        integer_object.set_property("number", number).unwrap();
+        integer_object
```
sdroege, line 20:
```suggestion
        Object::new(&[("number", &number)]).unwrap();
```

And do you actually need the `pub fn new()` constructor?

book/listings/scalable_lists/1/main.rs
```diff
@@ -0,0 +1,87 @@
...
+        let integer_object = IntegerObject::from_integer(number);
+        model.append(&integer_object);
+    }
+    // ANCHOR_END: model
+
+    // ANCHOR: factory_setup
+    let factory = SignalListItemFactory::new();
+    factory.connect_setup(move |_, list_item| {
+        let label = Label::new(None);
+        list_item.set_child(Some(&label));
```
sdroege, line 41: Shouldn't this already set the label value to the item's value?

book/src/gobject_memory_management.md
```diff
@@ -92,9 +91,7 @@ Yes! That is exactly what the [`Rc`](https://doc.rust-lang.org/std/rc/struct.Rc.
 The `Rc` counts the number of strong references created via `Clone::clone` and released via `Drop::drop`, and only deallocates it when this number drops to zero.
 We call every object containing a strong reference a shared owner of the value.
 If we want to modify the content of our [`Rc`](https://doc.rust-lang.org/std/rc/struct.Rc.html),
-we can use the [`RefCell`](https://doc.rust-lang.org/std/cell/struct.RefCell.html) type.
-`RefCell` then checks Rust's borrow rules during run time, namely that there can only be one mutable borrow at a time or multiple immutable borrows.
-`RefCell` panics if these rules are violated.
```
sdroege, line 97: I think this is not a good idea. `Cell` is much more limited and at least I rarely can use it. It would be useful to at least mention the constraints here (must be `Copy`) and the alternative (`RefCell`).

book/src/gobject_memory_management.md
```diff
@@ -111,7 +108,7 @@ We can improve that by using the `glib::clone!` macro.
 {{#rustdoc_include ../listings/gobject_memory_management/2/main.rs:callback}}
 ```
 
-Just like `Rc<RefCell<T>>`, GObjects are reference-counted and mutable.
+Just like `Rc<Cell<T>>`, GObjects are reference-counted and mutable.
```
bilelmoussaoui, line 111: This should stay as RefCell, most of the times you don't have a Copy type

Hofer-Julian: Sure, there was obvious a miscommunication in our last discussion.
I will then revert everything back to RefCell, because I don't want to use Cell AND RefCell without need.

book/src/gobject_memory_management.md
Hofer-Julian: I will revert this change

book/listings/scalable_lists/1/integer_object/mod.rs
Hofer-Julian: Huh, didn't know that this was possible :)
And no, in that case I don't need `new`.

book/listings/scalable_lists/1/integer_object/mod.rs
Hofer-Julian: I was contemplating if I should use the `From` trait here.
If you want to use it, then this decides it :)

book/src/gobject_memory_management.md
bilelmoussaoui: Well, it's not a need, it's about "correctness". The book should use RefCell where it fits & Cell where it does.

Hofer-Julian: Cell is correct in all instances I've used it. So is RefCell.

However, RefCell is useful in more cases.

Hofer-Julian: @sdroege any opinions on:
1) everything RefCell
2) Cell where "it makes sense"?

If we choose 2, we definitely need a discussion where it makes senses, and how to explain/motivate it to the reader.

sdroege: Use `Cell` where you can, use `RefCell` elsewhere, and where needed use `Mutex`, etc. Like you'd do in a proper application. You use the tool that fits your usage best.

Hofer-Julian:
TODO as discussed in the Matrix room:
- [ ] Address the comments
- [ ] Add this link (https://doc.rust-lang.org/1.30.0/book/first-edition/choosing-your-guarantees.html#cell-types)
- [ ] Add an example with RefCell & String in the memory management chapter

book/src/gobject_signals.md
```diff
@@ -59,43 +67,6 @@ Whenever we now receive the "max-number-reached" signal, the accompanying number
 {{#rustdoc_include ../listings/gobject_signals/3/main.rs:signal_handling}}
 ```
 
-Because the signal handlers can be at completely different places of the code than the emitters, we have to be extra careful when dealing with RefCells and Mutexes.
```
sdroege, line 62: I think it's not ideal that this paragraph goes away. This is a common problem that people run into once they have to use `RefCell`s :) Here it's nicely explained why and what to do.

sdroege (commented):
Otherwise seems fine to me

book/src/gobject_signals.md
Hofer-Julian: Glad to hear that you feel that it nicely explains this topic.
I, however, was not really satisfied to begin with.
With the `Cell` changes I don't find it acceptable anymore.

It is true however, that something similar should be included again, which is why I opened https://github.com/gtk-rs/gtk4-rs/issues/392

bilelmoussaoui: The thing is you're trying to over-simplify it to make it very newcomers friendly, which is not often a great idea.

book/listings/scalable_lists/1/integer_object/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+use gtk::glib;
+
+// ANCHOR: integer_object
+glib::wrapper! {
+    pub struct IntegerObject(ObjectSubclass<imp::IntegerObject>);
+}
+
+impl From<i32> for IntegerObject {
+    fn from(number: i32) -> Self {
+        Object::new(&[("number", &number)]).unwrap()
```
bilelmoussaoui, line 13: I would add a proper constructor to IntegerObject & use it in the From implementation.

Hofer-Julian: That's what I had before and @sdroege asked me to change it 😉

book/listings/scalable_lists/4/main.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    let factory = SignalListItemFactory::new();
+    // ANCHOR: factory_setup
+    factory.connect_setup(move |_, list_item| {
+        // Create label
+        let label = Label::new(None);
+        list_item.set_child(Some(&label));
+
+        // Create expression describing `list_item->item->number`
+        let list_item_expression = ConstantExpression::new(list_item);
```
bilelmoussaoui, line 43: Do you need the ConstantExpression here?

Hofer-Julian: Not sure.
I ported this nearly 1:1 from the gtk4-demo.

book/src/scalable_lists.md
```diff
@@ -0,0 +1,179 @@
+# Scalable Lists
+
+Sometimes you have more elements than you can display at once.
+The typical solution is to put them into a [`ScrolledWindow`](../docs/gtk/struct.ScrolledWindow.html) so you can scroll through the complete list.
+Howevee, in the most trivial case you would still create one widget per element.
```
bilelmoussaoui, line 5:
```suggestion
However, in the most trivial case you would still create one widget per element.
```
not sure how codespell didn't catch this one

book/src/scalable_lists.md
```diff
@@ -0,0 +1,179 @@
...
+Because creating new widgets takes a bit of time, this can become a problem if you have many element.
+Just think of the nearly infinite amount of posts in your social media timeline.
+
+This is why we use scalable lists instead.
+Here, a model holds our data, filters them and describes its order.
+The list item factory defines how the data transforms into widgets.
+The view specifies how the widgets are then arranged.
+Howvever, the most important part is that GTK only has to create slightly more widgets than we can currently look at.
```
bilelmoussaoui, line 13:
```suggestion
However, the most important part is that GTK only has to create slightly more widgets than we can currently look at.
```
same

book/src/scalable_lists.md
Hofer-Julian: Good catch.
Codespell is extremely limited.
On the plus side, that keeps the number of false positives to a minimum as well.

book/src/scalable_lists.md
```diff
@@ -0,0 +1,179 @@
...
+
+In order to interact with our `ListView`, we connect to its "activate" signal.
+
+<span class="filename">Filename: listings/scalable_lists/2/main.rs</span>
+
+```rust,no_run
+{{#rustdoc_include ../listings/scalable_lists/2/main.rs:list_view_activate}}
+```
+
+Now every time we activate an element, for example by double-clicking on it, the corresponding "number" property of the `IntegerObject` in the model will be increased by 1.
```
bilelmoussaoui, line 105:
```suggestion
Now every time we activate an element, for example by double-clicking on it, the corresponding `number` property of the `IntegerObject` in the model will be increased by 1.
```

book/src/scalable_lists.md
```diff
@@ -0,0 +1,179 @@
...
+{{#rustdoc_include ../listings/scalable_lists/4/main.rs:activate}}
+```
+
+After our changes, the application looks like this.
+
+<div style="text-align:center"><img src="img/scalable_lists_demo_2.png"/></div>
+
+
+We now know how to handle huge amount of data with the help of models such as [`gio::ListStore`](http://gtk-rs.org/docs/gio/struct.ListStore.html) and widgets like [`ListView`](../git/docs/gtk4/struct.ListView.html), [`ColumnView`](../git/docs/gtk4/struct.ColumnView.html) or [`GridView`](../git/docs/gtk4/struct.GridView.html).
+Please note that you do not have to have thousands of element until it makes sense to use this pattern.
```
bilelmoussaoui, line 178:
```suggestion
Please note that you do not have to have thousands of element until it makes sense to use this pattern.
```
are you sure about this?

book/src/scalable_lists.md
Hofer-Julian: I try to consistently use quotes instead of backticks when I talk about properties.

book/listings/scalable_lists/1/integer_object/mod.rs
bilelmoussaoui: That's because you had both a new & new_with_number or something like that :-)

book/src/scalable_lists.md
Hofer-Julian: Why not use it when you only have ~100?

book/src/scalable_lists.md
```diff
@@ -0,0 +1,179 @@
...
+As soon as you feel the need to add a [`ScrolledWindow`](../docs/gtk/struct.ScrolledWindow.html), it is probably already reasonable to start using models and views.
```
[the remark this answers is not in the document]
Hofer-Julian: What is the advantage of using FlowBox/ListBox?

book/src/scalable_lists.md
bilelmoussaoui: why use them when a listbox can do the job and is much simpler to setup?

bilelmoussaoui: the models are usable in listbox/flowbox as well btw. It's just you lose a ton of flexibility that was added to the *View widgets. but yes, starting the chapter with Listbox/FlowBox and having a section on scalable ones where you introduce the use of the View widgets would be much better.

book/src/scalable_lists.md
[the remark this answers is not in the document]
bilelmoussaoui: well there are no advantages, just disadvantages: they don't work with a high number of data (something around > 1K?) and that number also depends on the complexity of the widgets you need to setup. If the widgets are plain labels, a simple listbox can do the job for a quiet large number of data i think. But then if you're creating very complex widgets, having to create a thousands of them will make your app freeze (no draw) until all the widgets are created which what the View widgets work around: only create widgets for a small number of elements in the model & recycle them while you scroll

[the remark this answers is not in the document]
Hofer-Julian: Just to be clear, there is one advantage which is that they are easier to use, right?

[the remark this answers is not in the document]
bilelmoussaoui: Yes especially that a listbox/flowbox don't require a model so you can just add a label for the n elements you have and display whatever data you want. It depends on the use cases I guess :)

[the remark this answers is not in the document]
Hofer-Julian: Sounds good! 😊

Hofer-Julian:
I have now added a section about `ListBox`.

book/src/lists.md
```diff
@@ -0,0 +1,210 @@
...
+```rust,no_run
+{{#rustdoc_include ../listings/lists/1/main.rs:scrolled_window}}
+```
+
+<div style="text-align:center"><img src="img/lists_list_box.png"/></div>
+
+That was easy enough.
+However, we currently create one widget per element.
+Since creating new widgets takes a bit of time, this can become a problem with many elements.
+`ListBox` can easily handle a few thousands of them, but how could we possibly deal with the infinite amount of posts in a social media timeline?
```
bilelmoussaoui, line 31: Technically it's not ListBox that can handle thousands of elements, it's the creation of thousands of widgets that can be fine if those widgets are simple enough such as labels.  otherwise seems fine :)

book/src/lists.md
```diff
@@ -0,0 +1,210 @@
...
+
+We use scalable lists instead!
+
+- The **model** holds our data, filters them and describes its order.
```
bilelmoussaoui, line 35: The model concept is not unique to scalable lists btw, as I mentioned before you can use them with a listbox just fine

book/src/lists.md
```diff
@@ -0,0 +1,210 @@
...
+- The **list item factory** defines how the data transforms into widgets.
+- The **view** specifies how the widgets are then arranged.
+
+What makes this concept scalable is that GTK only has to create slightly more widgets than we can currently look at.
+As we scroll through our elements, the widgets which become invisible will be reused.
+The following figure demonstrates how this works in practice.
+
+<div style="text-align:center"><img src="img/scalable_lists_concept.png"/></div>
+
+100 000 elements is something `ListBox` will already struggle with, so let us this to demonstrate scalable lists.
```
bilelmoussaoui, line 45:
```suggestion
100 000 elements is something `ListBox` will struggle with, so let us this to demonstrate scalable lists.
```

book/src/lists.md
```diff
@@ -0,0 +1,210 @@
...
+The other options would have been [`MultiSelection`](../docs/gtk4/struct.MultiSelection.html) or [`NoSelection`](../docs/gtk4/struct.NoSelection.html).
+Then we pass the model and the factory to the [`ListView`](../git/docs/gtk4/struct.ListView.html).
+
+<span class="filename">Filename: listings/lists/2/main.rs</span>
+
+```rust,no_run
+{{#rustdoc_include ../listings/lists/2/main.rs:selection_list}}
+```
+
+Every `ListView` has to be inside a `ScrolledWindow`, so we are adding it to one.
```
bilelmoussaoui, line 103: Maybe explicitly mention that it has to be a direct child of a scrolledwindow, not inside a viewport or a box in-between? it's a common mistake.

book/src/lists.md
Hofer-Julian: I do mention this in the conclusion.
Is this enough for you?

Hofer-Julian:
@bilelmoussaoui I've resolved all comments apart from two.
I think they are fine as well, but you might want to have another look.

PR #364, opened 2021-05-07
Update gtk4 dependency.

mehmooda:
gtk-rs has updated cairo-rs minimum supported version to 1.14


mehmooda:
I don't know how to create multi-file commits from the github webui but #365

PR #366, opened 2021-05-08
gtk: Add run_future() for native dialogs

sophie-h:
Port gtk-rs/gtk-rs#500

### DIFF
--- a/gtk4/src/native_dialog.rs
+++ b/gtk4/src/native_dialog.rs
@@ -1,0 +1,55 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::prelude::*;
+use crate::NativeDialog;
+use crate::ResponseType;
+use std::cell::Cell;
+use std::future::Future;
+use std::pin::Pin;
+
+pub trait NativeDialogExtManual {
+    // rustdoc-stripper-ignore-next
+    /// Shows the dialog and returns a `Future` that resolves to the
+    /// `ResponseType` on response.
+    ///
+    /// ```no_run
+    /// # use gtk4 as gtk;
+    /// use gtk::prelude::*;
+    ///
+    /// # async fn run() {
+    /// let dialog = gtk::FileChooserNativeBuilder::new()
+    ///    .title("Select a File")
+    ///    .build();
+    ///
+    /// dialog.run_future().await;
+    /// println!("Selected file: {:?}", dialog.file());
+    /// dialog.destroy();
+    /// # }
+    /// ```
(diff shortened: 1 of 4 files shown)


bilelmoussaoui:
hmm, seems like I should drop the x11/wayland features I guess

gtk4/src/native_dialog.rs
```diff
@@ -0,0 +1,55 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::prelude::*;
+use crate::NativeDialog;
+use crate::ResponseType;
+use std::cell::Cell;
+use std::future::Future;
+use std::pin::Pin;
+
+pub trait NativeDialogExtManual {
```
bilelmoussaoui, line 10: should be added to prelude

bilelmoussaoui: oups, never mind

gtk4/src/native_dialog.rs
```diff
@@ -0,0 +1,55 @@
...
+    // rustdoc-stripper-ignore-next
+    /// Shows the dialog and returns a `Future` that resolves to the
+    /// `ResponseType` on response.
+    ///
+    /// ```no_run
+    /// # use gtk4 as gtk;
```
bilelmoussaoui, line 16:
```suggestion
```
there's no mention of gtk4 in the docs, it's mostly a crate name

sdroege (approved):

gtk4/src/native_dialog.rs
sophie-h: What do you mean? I think we need this line for the example to compile.

bilelmoussaoui: We already assume the package is renamed from gtk4 to gtk.

sophie-h: Added your change, no the tests fail because the example cannot compile.

bilelmoussaoui: wait, was that part of the code and not just a comment?

sophie-h: Yes. The hash just hides it in the docs:

https://doc.rust-lang.org/rustdoc/documentation-tests.html#hiding-portions-of-the-example

bilelmoussaoui: aah oh I see, TIL. My bad then, please add it back :)

sophie-h: Done :)

bilelmoussaoui:
Thanks Sophie! :)

PR #369, opened 2021-05-09
gtk: add ParamSpecExpression

bilelmoussaoui:


bilelmoussaoui:
I'm not sure whether having this is any useful as the expression param_spec doesn't have anything specific that would justify having bindings for ParamSpecExpression. Seems like GStreamer's custom param spec doesn't do that neither.

sdroege (approved):
Seems fine to me. Some of the GStreamer ones allow reading the min/max/default values, for example (but that's not implemented yet).

PR #375, opened 2021-05-13
Update to fix with latest nightly compiler

matzipan:
const_fn feature in field_offset crate doesn't work on nightly compiler.


bilelmoussaoui:
Not needed, just cargo update

PR #388, opened 2021-05-20
Add doc alias check

GuillaumeGomez:
Same as gtk-rs/gtk3-rs#540.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -116,6 +116,11 @@ jobs:
         working-directory: checker
       - run: ./check_init_asserts ../${{matrix.crate}}
         working-directory: checker
+      - name: Check doc aliases
+        run: |
+          python3 doc_aliases.py ../${{matrix.crate}}
+          cd .. && git diff -R --exit-code
+        working-directory: checker
 
   checker-macros:
     name: gtk-rs macro checker


gsk4/src/text_node.rs
```diff
@@ -75,6 +75,7 @@ impl TextNode {
     //
     // Due to https://gitlab.gnome.org/GNOME/gtk/-/issues/3675 this function wasn't properly
     // exported in 4.0.0 and so requires 4.1.2 at least for it to work properly.
+    #[doc(alias = "gsk_text_node_has_color_glyphs")]
```
bilelmoussaoui, line 78: there's one already, it's just before the comment

gtk4/src/constant_expression.rs
```diff
@@ -11,6 +11,7 @@ define_expression!(
 
 impl ConstantExpression {
     #[doc(alias = "gtk_constant_expression_new")]
+    #[doc(alias = "gtk_constant_expression_new_for_value")]
```
bilelmoussaoui, line 14: new_for_value is already used on the second function `for_value`, is this one needed here too?

gtk4/src/rt.rs
```diff
@@ -90,6 +90,8 @@ pub unsafe fn set_initialized() {
 /// so will not cause the program to terminate if GTK could not be initialized.
 /// Instead, an Ok is returned if the windowing system was successfully
 /// initialized otherwise an Err is returned.
+#[doc(alias = "gtk_init")]
+#[doc(alias = "gtk_is_initialized")]
```
bilelmoussaoui, line 94:
```suggestion
```

gsk4/src/text_node.rs
GuillaumeGomez: Then it's a bug! I wondered If I should include comments in the "look back", well, that confirms it.

gtk4/src/constant_expression.rs
GuillaumeGomez: The all knowing script says so. :)

gtk4/src/rt.rs
GuillaumeGomez: I need to ignore it in the script then.

gsk4/src/text_node.rs
GuillaumeGomez: Fixed in https://github.com/gtk-rs/checker/pull/22.

PR #391, opened 2021-05-21
gtk: drop MainContext check

bilelmoussaoui:
there's no need to ensure we are initializing gtk on the main thread
if gtk was not initialized yet


sdroege:
Same is also needed for gtk3-rs btw. You create a PR?

sdroege (approved):

bilelmoussaoui:
> Same is also needed for gtk3-rs btw. You create a PR?

already done :-)

PR #403, opened 2021-05-23
book: Take advantage of the newly introduced `builder` method

Hofer-Julian:
It was in introduced in PR #401

### DIFF
--- a/book/listings/gobject_memory_management/1/main.rs
+++ b/book/listings/gobject_memory_management/1/main.rs
@@ -1,6 +1,6 @@
 use gtk::prelude::*;
 use gtk::Application;
-use gtk::{self, ApplicationWindowBuilder, ButtonBuilder, Orientation};
+use gtk::{self, ApplicationWindow, Button, Orientation};
 use std::{cell::RefCell, rc::Rc};
 fn main() {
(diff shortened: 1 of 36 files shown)


bilelmoussaoui:
Do you want doing the same thing for the examples as well?

Hofer-Julian:
(Nearly) done!

Any idea though why `Application` has no builder method?
https://gtk-rs.org/gtk4-rs/git/docs/gtk4/struct.Application.html

bilelmoussaoui:
> (Nearly) done!
> 
> Any idea though why `Application` has no builder method?
> https://gtk-rs.org/gtk4-rs/git/docs/gtk4/struct.Application.html

Please open an issue, I have an idea what can be the cause of that

bilelmoussaoui:
No need to actually, because Application's code is manual that's why. I need to investigate why it's the case though

Hofer-Julian:
> No need to actually, because Application's code is manual that's why. I need to investigate why it's the case though

Sure, I've opened https://github.com/gtk-rs/gtk4-rs/issues/405 for now.
Unless you have comments, I'd like to merge this PR then, so I can rebase the other book branch then.

PR #413, opened 2021-05-28
gtk: register actions on the subclassed widget

bilelmoussaoui:
We use Widget's class data to store the actions used by install_action which is problematic 
when you have a Widget that subclasses another Widget subclass because the class_data is already defined by the first one.

Instead, store the actions on the subclass's class_data

### DIFF
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -588,7 +588,7 @@ unsafe impl<T: WidgetImpl> IsSubclassable<T> for Widget {
             let mut data = T::type_data();
             let data = data.as_mut();
             // Used to store actions for `install_action`
-            data.set_class_data(Widget::static_type(), Actions::default());
+            data.set_class_data(<T as ObjectSubclassType>::type_(), Actions::default());
         }
 
         klass.compute_expand = Some(widget_compute_expand::<T>);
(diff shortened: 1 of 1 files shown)


gtk4/src/subclass/widget.rs
```diff
@@ -588,7 +588,7 @@ unsafe impl<T: WidgetImpl> IsSubclassable<T> for Widget {
             let mut data = T::type_data();
             let data = data.as_mut();
             // Used to store actions for `install_action`
-            data.set_class_data(Widget::static_type(), Actions::default());
+            data.set_class_data(<T as ObjectSubclassType>::type_(), Actions::default());
```
sdroege, line 591: That's also suboptimal. The class data is owned by the class, not your code :)

If you need to store multiple things in there you'll have to store a `HashMap` or so.

bilelmoussaoui: Actually it doesn't seem to be the issue at all and this change isn't relevant anymore. The code in Fractal had the wrong class_init/instance_init. It wasn't chaining up the right parent and I didn't notice it till later. Although I thought this change wasn't very probelmatic at the end , or is it ?

sdroege: Yes, it would be a problem if the subclass wants to store class data itself. You could conflict then.

bilelmoussaoui: Ah I see. I had a huge mis-undertanding how it all worked and now it's much clearer :) Thanks!

PR #419, opened 2021-05-29
gtk: add Buildable subclassing support

bilelmoussaoui:
This provides basic Buildable interface implementation support
without the custom_ tags supports

The custom_ parts of Buildable requires providing BuildableParser/BuildableParserContext
which are very C.

As those are not often used by application developers, don't provide them for now.

Part of #56 

cc @jsparber

### DIFF
--- a/gtk4/src/subclass/buildable.rs
+++ b/gtk4/src/subclass/buildable.rs
@@ -1,0 +1,396 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::prelude::*;
+use crate::subclass::prelude::*;
+use crate::{Buildable, Builder};
+use glib::translate::*;
+use glib::{Cast, GString, Object, Quark, Value};
+use once_cell::sync::Lazy;
+
+pub trait BuildableImpl: ObjectImpl {
+    fn set_id(&self, buildable: &Self::Type, id: &str) {
+        self.parent_set_id(buildable, id)
+    }
+    fn id(&self, buildable: &Self::Type) -> GString {
+        self.parent_id(buildable)
+    }
+    fn add_child(
+        &self,
+        buildable: &Self::Type,
+        builder: &Builder,
+        child: &Object,
+        type_: Option<&str>,
+    ) {
+        self.parent_add_child(buildable, builder, child, type_)
+    }
+    fn set_buildable_property(
+        &self,
+        buildable: &Self::Type,
+        builder: &Builder,
+        name: &str,
+        value: &Value,
+    ) {
+        self.parent_set_buildable_property(buildable, builder, name, value)
+    }
+    fn parser_finished(&self, buildable: &Self::Type, builder: &Builder) {
+        self.parent_parser_finished(buildable, builder)
+    }
+    fn internal_child(
+        &self,
+        buildable: &Self::Type,
+        builder: &Builder,
+        name: &str,
+    ) -> Option<Object> {
+        self.parent_internal_child(buildable, builder, name)
+    }
+    fn construct_child(&self, buildable: &Self::Type, builder: &Builder, name: &str) -> Object {
+        self.parent_construct_child(buildable, builder, name)
+    }
+    /*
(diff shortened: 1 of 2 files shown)


gtk4/src/subclass/buildable.rs
```diff
@@ -0,0 +1,370 @@
...
+        builder: &Builder,
+        child: &Object,
+        type_: Option<&str>,
+    );
+    fn set_buildable_property(
+        &self,
+        buildable: &Self::Type,
+        builder: &Builder,
+        name: &str,
+        value: Value,
```
sdroege, line 25:
```suggestion
        value: &Value,
```

Maybe?

sdroege (commented):
Looks good to me otherwise

bilelmoussaoui:
> Looks good to me otherwise

Thanks for the review :) I will wait till Julian can give it a try as he has a use case for it in Fractal

gtk4/src/subclass/buildable.rs
```diff
@@ -0,0 +1,384 @@
...
+        builder: &Builder,
+        name: &str,
+        value: &Value,
+    ) {
+        unsafe {
+            let type_data = Self::type_data();
+            let parent_iface =
+                type_data.as_ref().parent_interface::<Buildable>() as *const ffi::GtkBuildableIface;
+
+            let func = (*parent_iface)
```
jsparber, line 178: i think we should just silently do nothing when there isn't a parent implementation.  I had to override it just because `gtk::Widget`  doesn't implement it. Probably the same for other `parent_*()` functions.

jsparber: Reading a little bit more of the docs. Set `buildable_property()` falls back to `g_object_set_property()` when there is no custom implementation. 

> Sets a property of a buildable object. It is normally not necessary to implement this, g_object_set_property() is used by default. GtkWindow implements this to delay showing itself (i.e. setting the “visible” property) until the whole interface is created.

bilelmoussaoui: Should be better now :)

bilelmoussaoui: I should have looked at gtkbuilder.c first ^^

sdroege (approved):

gtk4/src/subclass/buildable.rs
```diff
@@ -0,0 +1,394 @@
...
+            // gtk::Builder falls back to using ObjectExt::set_property if the method is not implemented
+            if let Some(func) = (*parent_iface).set_buildable_property {
+                func(
+                    buildable.unsafe_cast_ref::<Buildable>().to_glib_none().0,
+                    builder.to_glib_none().0,
+                    name.to_glib_none().0,
+                    value.to_glib_none().0,
+                )
+            } else {
+                buildable.set_property(name, value).unwrap_or_else(|err| {
```
jsparber, line 192: we need to use `set_property_from_value()` here.

bilelmoussaoui: What's the difference ?:)

jsparber: `set_property()` takes a `ToValue`. It runs `value.into()`, which makes the actually set `value` into `Value<Value<T>`. Look at the implementation https://gtk-rs.org/gtk-rs-core/git/docs/glib/object/trait.ObjectExt.html#tymethod.set_property

bilelmoussaoui: Oups indeed

jsparber: I noticed because of the nice error message you made :)

sdroege: See https://github.com/gtk-rs/gtk-rs-core/pull/85

bilelmoussaoui: @jsparber should be fixed now

jsparber:
works now as excepted :)

bilelmoussaoui:
> works now as excepted :)

thanks for testing :)

PR #422, opened 2021-05-30
CI docs: use gir-rustdoc

bilelmoussaoui:
a simple utility to build the index page & provide versioned docs support

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,12 +26,16 @@ jobs:
       - working-directory: gir
         run: cargo build --release
       - run: cargo install rustdoc-stripper
+      - run: git clone https://gitlab.gnome.org/World/Rust/gir-rustdoc/ # checkout action doesn't support random urls
+      - run: python3 ./gir-rustdoc/gir-rustdoc.py --default-branch 'master' --branch 'master' pre-docs
       - run: python3 ./generator.py --embed-docs --yes ./
+      - run: python3 ./gir-rustdoc/gir-rustdoc.py --project-title 'GTK 4 Rust bindings' html-index --early-section 'Book' '<ul><li><a href="stable/book">Stable</a></li><li><a href="git/book">Development</a></li></ul>'
       - run: echo "DEST=$(if [ "$GITHUB_EVENT_NAME" == "release" ]; then echo 'stable/${{ github.event.release.tag_name }}'; else echo 'git'; fi)" >> ${GITHUB_ENV}
       - uses: actions-rs/cargo@v1
         env:
           RUSTDOCFLAGS: |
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -26,20 +30,12 @@ jobs:
       - working-directory: gir
         run: cargo build --release
       - run: cargo install rustdoc-stripper
+      - run: git clone https://gitlab.gnome.org/World/Rust/gir-rustdoc/ # checkout action doesn't support random urls
+      - run: eval $(python3 ./gir-rustdoc/gir-rustdoc.py pre-docs)
```
sophie-h, line 34: I think you want to specify `--default-branch` here, because it defaults to 'main' and github does not want to tell me the default branch.

sophie-h (changes requested):

sophie-h (commented):
You would still need something that runs on something else than the master branch for the release docs, right?

bilelmoussaoui:
> You would still need something that runs on something else than the master branch for the release docs, right?

Not really, it's the same thing :-) 
the magic happens at
```
      - run: echo "DEST=$(if [ "$GITHUB_EVENT_NAME" == "release" ]; then echo 'stable/${{ github.event.release.tag_name }}'; else echo 'git'; fi)" >> ${GITHUB_ENV}
```

sophie-h:
> Not really, it's the same thing :-)

Oh, okay :)  I got suspicious because you don't really need to call `html-index` outside of `master`. But let's just see what happens xD

bilelmoussaoui:
> > Not really, it's the same thing :-)
> 
> Oh, okay :) I got suspicious because you don't really need to call `html-index` outside of `master`. But let's just see what happens xD

What would that cause? a re-generate index.html? it  would still contain the same information :p

sophie-h:
> What would that cause? a re-generate index.html? it would still contain the same information :p

Okay, the problem is I have never worked with github CI. I might be completely off.

The idea was to have the configuration for the index page centralized at master. If you know push a fix to a release branch, would that regenerate the index page with the potentially old configuration in the release branch?

bilelmoussaoui:
> > What would that cause? a re-generate index.html? it would still contain the same information :p
> 
> Okay, the problem is I have never worked with github CI. I might be completely off.
> 
> The idea was to have the configuration for the index page centralized at master. If you now push a fix to a release branch, would that regenerate the index page with the potentially old configuration in the release branch?

Ah no, the release/stable docs are generated on release event, not when you push to other branches.

bilelmoussaoui:
OK CI is green, I will merge this to see how the result looks like and then figure out the story with the rustdocflags later :)

PR #476, opened 2021-06-23
gtk4: drop probelmatic links

bilelmoussaoui:
see the release branch

### DIFF
--- a/gtk4/sys/Cargo.toml
+++ b/gtk4/sys/Cargo.toml
@@ -7,7 +7,6 @@ edition = "2018"
 homepage = "http://gtk-rs.org/"
 keywords = ["gtk", "gtk4", "ffi", "gtk-rs", "gnome"]
 license = "MIT"
-links = "gtk-4"
 name = "gtk4-sys"
 repository = "https://github.com/gtk-rs/gtk4-rs"
 version = "0.2.0"


gtk4/sys/Cargo.toml
```diff
@@ -7,7 +7,6 @@ edition = "2018"
 homepage = "http://gtk-rs.org/"
 keywords = ["gtk", "gtk4", "ffi", "gtk-rs", "gnome"]
 license = "MIT"
-links = "gtk-4"
```
sdroege, line 10: Why is this a problem?

bilelmoussaoui: Without this change i get 
```
the package `gtk4-sys` links to the native library `gtk-4`, but it conflicts with a previous package which links to `gtk-4` as well:
package `gtk4-sys v0.1.0 (/__w/gtk4-rs/gtk4-rs/gtk4/sys)`

failed to select a version for `gtk4-sys` which could resolve this conflict
```
dropping the usage of links in the other -sys crates in gtk3-rs/gtk-rs-core fixes their release CI as well

sdroege: Which other crate lists `gtk-4` as `links`? This error should only happen if there are two or more.

bilelmoussaoui: there are none of them actually, only the -sys crates do

sdroege: That must mean that the crate is pulled in multiple times. Please debug this instead of removing the `links=` sections :) Something's very wrong.

bilelmoussaoui:
Closing as the change wasn't needed at all

PR #495, opened 2021-06-30
Implement TreeIter manually to allow instance creation and storage of objects

felinira:
This would in theory allow for TreeModel subclassing. I am not completely happy with how it turned out because FromGlibPtrBorrow copies the struct data. I did not really find a better alternative though.


bilelmoussaoui:
> This would in theory allow for TreeModel subclassing. 

Is that something you plan to work on? Most of the code is already in #169, it could use some testing & an example

felinira:
> > This would in theory allow for TreeModel subclassing.
> 
> Is that something you plan to work on? Most of the code is already in #169, it could use some testing & an example

I was looking into it at least. As TreeIter was the biggest stumbling block from that issue I concentrated on that first. I might even have a use case for custom TreeModels. Even though it is a very complex interface and definitely not "fun" to implement ^^

gtk4/src/auto/mod.rs
```diff
@@ -990,9 +990,6 @@ pub use self::recent_info::RecentInfo;
 mod text_iter;
 pub use self::text_iter::TextIter;
 
-mod tree_iter;
-pub use self::tree_iter::TreeIter;
```
sdroege, line 994: You need to `git rm` that file too :)

gtk4/src/tree_iter.rs
```diff
@@ -0,0 +1,152 @@
...
+// 0xffffffff is an invalid pointer that is distinct from the null pointer
+// This makes it unlikely this is in here by accident
+const MAGIC_NUMBER_1: usize = 0xffffffff;
+// "rust" in ASCII
+const MAGIC_NUMBER_2: usize = 0x72757374;
+
+#[repr(C)]
+#[derive(Clone)]
+#[doc(alias = "GtkTreeIter")]
+pub struct TreeIter {
```
sdroege, line 21: Why not

```rust
#[repr(transparent)]
pub struct TreeIter(ffi::GtkTreeIter);
```
?

gtk4/src/tree_iter.rs
```diff
@@ -0,0 +1,152 @@
...
+
+        // We store magic numbers in here to find out if we are allowed to deref the pointer
+        // This makes it very unlikely that we are accidentally dereferencing a valid pointer
+        let user_data2 = MAGIC_NUMBER_1 as gpointer;
+        let user_data3 = MAGIC_NUMBER_2 as gpointer;
+
+        Self::from_raw(stamp, user_data, user_data2, user_data3)
+    }
+
+    pub fn from_raw(
```
sdroege, line 45:
```suggestion
    pub unsafe fn from_raw(
```

gtk4/src/tree_iter.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    pub fn has_object(&self) -> bool {
+        self.user_data2 as usize != MAGIC_NUMBER_1 && self.user_data3 as usize != MAGIC_NUMBER_2
+    }
+
+    /// Returns the user data from user_data as a glib::Object
+    /// This dereferences the memory unconditionally.
+    /// Be sure you know what you are dealing with
+    unsafe fn object_unchecked(&self) -> glib::Object {
+        Object::clone(&from_glib_borrow(self.user_data as *mut GObject))
```
sdroege, line 79:
```suggestion
        from_glib_none(self.user_data as *mut GObject)
```

gtk4/src/tree_iter.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+impl TreeIter {
+    /// Create a TreeIter referencing a glib::Object
+    pub fn new(stamp: i32, item: Object) -> Self {
+        assert_initialized_main_thread!();
+
+        // We only really need the first user_data field for our data.
+        // We will store a reference to the glib object here.
+        let user_data = unsafe { glib::gobject_ffi::g_object_ref(item.as_ptr()) } as gpointer;
```
sdroege, line 35: This would leak the object if the iter is passed to C and freed there. Also if you clone the iter you wouldn't increase the refcount of the object, but then unref it twice later.

You can't store any owned memory in an iter, the API does not really allow such things. But for safely storing unowned/borrowed pointers you need something else.

sdroege (commented):
I don't really know how this API is supposed to be used in C when implementing your own tree model, but like this the implementation is unsound in Rust.

You'd need to find a way to not store any owned data inside the iter but without that potentially causing dangling pointers when used from safe code.

bilelmoussaoui:
What I would suggest here: 

Try to make use of the TreeModel subclassing api in #169 and see what are the use cases you have for TreeIter. That would help defining an API based on a use case. We can iter on it in the future if needed after that.

felinira:
> I don't really know how this API is supposed to be used in C when implementing your own tree model, but like this the implementation is unsound in Rust.
> 
> You'd need to find a way to not store any owned data inside the iter but without that potentially causing dangling pointers when used from safe code.

The model is expected to handle the lifetime of the TreeIter. All TreeIters get invalidated when you change the model in any way. If you use GTK_TREE_MODEL_ITERS_PERSIST as a flag a TreeIter is only invalidated when it is actually removed.

You can't really control who has references to your TreeIters, they are not GObjects so they are not reference counted which makes this unsafe to use. I should have seen it earlier that you can't really know who still has references to your TreeIters but didn't really think it through. This whole API is a giant use after free nightmare. So I do come to the (sad) conclusion that it seems to be impossible to store references in the pointer fields.

This also means the way of just taking the TreeIter references from C and transparently casting them is unsafe too. I'm not completely sure how Boxed works and if it catches these kinds of issues but if I am not mistaken even the current auto implementation is unsafe too because if you store a reference to the TreeIter anywhere and the model frees it you have use after free.

So the safest way is probably to make the struct Copy.

> What I would suggest here:
> 
> Try to make use of the TreeModel subclassing api in #169 and see what are the use cases you have for TreeIter. That would help defining an API based on a use case. We can iter on it in the future if needed after that.

I tried using the indextree crate. The problem is you need some way to store a [NodeId](https://docs.rs/indextree/4.3.1/indextree/struct.NodeId.html) or a reference to one of those. As you can't store the struct directly you probably need to create a HashMap or a Vec and store your object there and just store indexes in the TreeIter.

So after thinking about it a bit the only safe way I can think of is just storing an integer in the user_data pointer (or even using the stamp field for it and leaving user_data completely alone). Then you store all the stuff you need elsewhere as this data structure sadly is not really capable of doing anything remotely useful in a safe way.

I will redo the whole thing and just allow it to store integers. At least they can't betray you in the way pointers can^^.

gtk4/src/tree_iter.rs
felinira: RGBA in glib does it the same way but sure I can do it that way, nothing wrong with it.

sdroege:
The `stamp` probably makes sense to keep, and you could expose the pointers as `usize`s. The idea from a Rust point of view here would be that you'd have to give each of your items a unique identifier (the `usize`s) and the `stamp` would be used to notice when the iterator is stale (you'd keep a `stamp` internally too and on every change increase it, and on iterator usage check if the iterator has the latest `stamp`).

For the general memory unsafety here, can you reproduce a invalid memory access or similar with just using the existing Rust bindings around this? That would be bad and would require further investigation there, but also there's limits to what we can do. See https://github.com/gtk-rs/gtk3-rs/issues/333

gtk4/src/tree_iter.rs
sdroege: Because there the fields are public and that would be ugly to use otherwise

gtk4/src/tree_iter.rs
```diff
@@ -0,0 +1,87 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use glib::ffi::gpointer;
+use glib::translate::*;
+
+use std::{mem, ptr};
+
+#[repr(transparent)]
+#[doc(alias = "GtkTreeIter")]
+pub struct TreeIter(ffi::GtkTreeIter);
```
sdroege, line 10: As long as you don't expose pointers in a "safe" way, it should be fine to keep the iterator `Copy` and `Clone` (and ideally also derive `Debug`).

The current implementation is also `Clone`, just that it unnecessarily uses heap allocations (which is something you fix here).

PR #500, opened 2021-07-04
use AsRef<RenderNode> instead of IsRenderNode

bilelmoussaoui:
Would it be an api break to backport this to 0.1?

### DIFF
--- a/gdk4/src/event.rs
+++ b/gdk4/src/event.rs
@@ -236,6 +236,13 @@ impl fmt::Debug for Event {
     }
 }
 
+#[doc(hidden)]
+impl AsRef<Event> for Event {
+    fn as_ref(&self) -> &Self {
+        self
+    }
+}
+
 pub unsafe trait EventKind:
     StaticType + FromGlibPtrFull<*mut ffi::GdkEvent> + 'static
 {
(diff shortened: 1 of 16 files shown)


sdroege:
> Would it be an api break to backport this to 0.1?

Yes, but we should then probably also do the same for the 5-6 other cases of custom inheritance that we have here, in gtk3-rs and in cairo and glib.

sdroege:
Can the traits be removed now or are they still useful?

bilelmoussaoui:
> > Would it be an api break to backport this to 0.1?
> 
> Yes, but we should then probably also do the same for the 5-6 other cases of custom inheritance that we have here, in gtk3-rs and in cairo and glib.

At least in gtk4-rs, they are all handled properly now. I don't think I will have the time to fix others though.

bilelmoussaoui:
> Can the traits be removed now or are they still useful?

They are still useful for downcast/upcast methods

sdroege:
> At least in gtk4-rs, they are all handled properly now. I don't think I will have the time to fix others though.

No problem, can you create an issue though? :)

bilelmoussaoui:
From a quick look:

- gtk3-rs has only a `gdk::FromEvent` trait, it's not used on the APIs at all except for downcast/is_a kind of methods. No actions needed it seems
- There's none in gtk-rs-core

bilelmoussaoui:
For cairo though, there's a Deref implementation so it's more or less the same thing. Maybe we should use AsRef there as well?

bilelmoussaoui:
I will go ahead and merge this one, no backport to 0.1 but I will do a 0.2 in a week or so

PR #513, opened 2021-07-13
Add a custom buildable example

A6GibKm:
The only point which is handled in an unsatisfactory way is how we determine if a child has already been bindedn in  `BuildableImpl` via
```rust
         if buildable.first_child().is_none() {
            self.parent_add_child(buildable, builder, child, type_);
        } 
```

This is addressed in the second commit where a method was added to TemplateChild to verify if it has been bound already.


gtk4/src/subclass/widget.rs
```diff
@@ -1298,6 +1298,12 @@ where
                 .expect("Failed to retrieve template child. Please check that it has been bound.")
         }
     }
+
+    /// Determines if the child has been bound. This is primarily
+    /// useful for implementing the Buildable interface.
+    pub fn is_bound(&self) -> bool {
```
sdroege, line 1304: This should come together with a `fn try_get(&self) -> Option<T>` as otherwise you'll end up with code like

```rust
if self.foo.is_bound() {
  self.foo.stuff();
}
```

while

```rust
if let Some(foo) = self.foo.get() {
   foo.do_stuff();
}
```

would be nicer.

A6GibKm: Added
```rust
    /// Returns Some(child) if the widget has been bound.
    pub fn try_get(&self) -> Option<T> {
        unsafe { Option::<T>::from_glib_none(self.ptr) }
    }
```

examples/Cargo.toml
```diff
@@ -44,6 +44,10 @@ path = "css/main.rs"
 name = "custom_application"
 path = "custom_application/main.rs"
 
+[[bin]]
```
bilelmoussaoui, line 47: It needs an entry on the README. Along with a readme/screenshot, like other examples.

gtk4/src/subclass/widget.rs
```diff
@@ -1294,10 +1294,21 @@ where
     #[track_caller]
     pub fn get(&self) -> T {
         unsafe {
```
bilelmoussaoui, line 1296: unsafe is no longer necessary here

examples/Cargo.toml
sdroege: Yeah, especially for me, who doesn't really know GTK :) I have no idea what this example actually does

gtk4/src/subclass/widget.rs
```diff
@@ -1294,10 +1294,21 @@ where
...
-            Option::<T>::from_glib_none(self.ptr)
+            self.try_get()
                 .expect("Failed to retrieve template child. Please check that it has been bound.")
         }
     }
+
+    /// Determines if the child has been bound. This is primarily
+    /// useful for implementing the Buildable interface.
```
bilelmoussaoui, line 1303:
```suggestion
    /// useful for implementing the [`Buildable`][`crate::Buildable`] interface.
```

bilelmoussaoui:
Thanks!

PR #516, opened 2021-07-16
book: Add interface builder chapter

Hofer-Julian:
This PR adds a new chapter about the interface builder.


book/listings/interface_builder/1/window.ui
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<interface>
+  <object class="GtkApplicationWindow" id="window">
+    <property name="title">My GTK App</property>
+    <child>
+      <object class="GtkButton" id ="button">
+        <property name="label">Press me!</property>
+        <property name="margin_top">12</property>
```
bilelmoussaoui, line 8: properties should be written with a `-` instead of `_`. It's a bad habit from glade.

book/listings/interface_builder/2/window/imp.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+}
+// ANCHOR_END: subclass
+
+// ANCHOR: object_impl
+// Trait shared by all GObjects
+impl ObjectImpl for Window {
+    fn constructed(&self, obj: &Self::Type) {
+        // Call "constructed" on parent
+        self.parent_constructed(obj);
```
bilelmoussaoui, line 40: chain up at the end i suppose?

book/listings/interface_builder/2/window/window.ui
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<interface>
+  <template class="MyGtkAppWindow" parent="GtkApplicationWindow">
+    <property name="title">My GTK App</property>
+    <child>
+      <object class="GtkButton" id ="button">
+        <property name="label">Press me!</property>
+        <property name="margin_top">12</property>
```
bilelmoussaoui, line 8: same here

book/src/interface_builder.md
```diff
@@ -0,0 +1,135 @@
...
+
+<span class="filename">Filename: listings/interface_builder/1/main.rs</span>
+```rust,no_run
+{{#rustdoc_include ../listings/interface_builder/1/main.rs:build_ui}}
+```
+
+This is a bit disappointing.
+Even though we have already described the UI in the markup file, the amount of code is still pretty much the same.
+There are still cases where it is valuable to know of the existence of `gtk::Builder`.
+We will see for example that [`ShortcutsWindow`](../docs/gtk4/struct.ShortcutsWindow.html) is quite a bit easier to instantiate that way.
```
bilelmoussaoui, line 39: Not really because gtk provides a pretty handy way to do that. In general, you very often don't want GtkBuilder directly

book/src/interface_builder.md
```diff
@@ -0,0 +1,135 @@
...
+{{#rustdoc_include ../listings/interface_builder/2/main.rs:build_ui}}
+```
+Also with regard to capabilities, we get the best of both worlds.
+
+Thanks to custom widgets we can
+- keep state and part of it as properties,
+- add signals as well as
+- override behavior.
+
+Thanks to composite templates we can
```
bilelmoussaoui, line 121: It also allows splliting code into smaller widgets. It would be also nice to mention Buildable, which are the objects you can instantiate through a gtk::Builder and so in a composite template as well. the Buildable interface is what tells to gtk how to transform what you wrote in xml to an object or whatever you want.

bilelmoussaoui:
Otherwise looks fine to me

book/listings/interface_builder/2/window/imp.rs
Hofer-Julian: I wasn't sure if it belongs at the beginning or end. I'll move it if that's the correct position.

book/src/interface_builder.md
Hofer-Julian: You mean the trick with GResources?

book/src/interface_builder.md
Hofer-Julian: > It also allows splliting code into smaller widgets

Can you give an example? 


Regarding interface. You want me to add a sentence like "if your widget implements BuildableExt trait then you can use it with Builder"?

book/listings/interface_builder/2/window/imp.rs
A6GibKm: From what I have seen from libadwaita and gtk it is done at the start. But there might be a niche reason to do it at the end.

In most places in Rust I see it at the end, and it seems to be done that way just because it is the way it has been done.

book/src/interface_builder.md
A6GibKm: That *trick* is explained in GtkApplication documentation https://docs.gtk.org/gtk4/class.Application.html#automatic-resources, since the docs of the Application class should be the most visited/read ones I would refrain from calling it a trick.

book/listings/interface_builder/2/window/imp.rs
sdroege: Who does it at the end? All my code in gst-plugins-rs does it at the beginning too and that's more correct in most cases

sdroege: Also the gtk-rs/glib tests and docs. Doing it at the end has the potential of failing, depending on the parent class.

A6GibKm: > Who does it at the end? 

Apps.

sdroege: That sounds like a bad idea unless they have a special reason for that. For other virtual methods that might make sense but not for constructed. It's always a decision for each virtual method when you want to have your behaviour and when the base behaviour.

sdroege: Someone fix those then please :)

book/src/interface_builder.md
Hofer-Julian: No reason to get worked up about it.
If that's what Bilal meant, I can't use it anyways because Ressources will be only introduced later.

PR #531, opened 2021-07-25
CompositeTemplate: support setting internal-child to true

### LINKED ISSUE #530: TemplateChild doesn't allow setting internal_child
Currently, `bind_template_child_with_offset` is used with `internal-child` set to `false`. We will need some way to set it to `true`, I was thinking of something like adding another property to the `#[template(id = "some_id", internal)]`.

bilelmoussaoui:
This adds the possibility to modify the value we set in gtk_widget_class_bind_template_child_full
to something other than false, which would allow people to use internal-child widgets in their code as well

Fixes #530


bilelmoussaoui:
This seems to be still used on some old widgets that were not properly to gtk4's new way of doing this which consists of using properties for setting those internal-widgets. 

So we can probably drop the example change as it's not something people should make use of except when creating some custom widgets.

gtk4-macros/src/lib.rs
```diff
@@ -42,18 +42,20 @@ use syn::{parse_macro_input, DeriveInput};
 ///
 ///     #[derive(Debug, Default, CompositeTemplate)]
 ///     #[template(file = "test/template.ui")]
-///     pub struct MyWidget {
+///     pub struct MyDialog {
```
sdroege, line 45: This should probably all be in a separate commit

gtk4/src/subclass/widget.rs
```diff
@@ -1224,6 +1224,7 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
     unsafe fn bind_template_child_with_offset<T>(
         &mut self,
         name: &str,
+        internal: bool,
```
sdroege, line 1227: This is of course an API change, so can't be backported

bilelmoussaoui: yeah, this is something I would keep till 0.3, I don't think a lot of people would use such APIs. But it's nice to have it there someday :)

gtk4-macros/src/lib.rs
bilelmoussaoui: As the GTK developers don't want you to use the internal-child in your newly created custom widgets, I would  drop the example changes and add a section that explains the various inputs the `template` macro attribute can take.

sdroege: I meant the type renaming. That change makes sense but is independent from the other things in the commit :)

PR #532, opened 2021-07-27
examples: Add a glium GLArea example

YaLTeR:
Tested on Linux and on Windows.

I'm not sure if what I wrote in the Safety comment is enough and I'm not sure if anything GTK does with the backend would cause a glium state mismatch. Someone with a deeper knowledge of glium could make that assessment. For now this simple example seems to work fine.


examples/glium_gl_area/main.rs
```diff
@@ -0,0 +1,280 @@
...
+    application.connect_activate(build_ui);
+    application.run();
+}
+
+fn build_ui(application: &gtk::Application) {
+    let window = gtk::ApplicationWindow::new(application);
+
+    window.set_title(Some("Glium in GLArea"));
+
+    let widget = gtk::GLArea::new();
```
bilelmoussaoui, line 255: You could have made it a GLArea subclass but no strong opinions on that

examples/glium_gl_area/main.rs
```diff
@@ -0,0 +1,280 @@
...
+
+struct GLAreaBackend {
+    // GLArea will transitively hold a strong reference to this struct, so to prevent a reference
+    // cycle, this struct must hold a weak reference to the GLArea.
+    area: WeakRef<gtk::GLArea>,
+}
+
+unsafe impl glium::backend::Backend for GLAreaBackend {
+    fn swap_buffers(&self) -> Result<(), glium::SwapBuffersError> {
+        // GLArea handles this by itself.
```
sdroege, line 17: Does this maybe require signalling to the `GLArea` that its dirty and needs to redraw?

examples/glium_gl_area/main.rs
sdroege: A subclass would make conceptually more sense IMHO

examples/glium_gl_area/main.rs
YaLTeR: The intention is that you only do this within the `render` signal. I'm not sure what happens if you do it outside.

examples/glium_gl_area/main.rs
```diff
@@ -0,0 +1,280 @@
...
+fn build_ui(application: &gtk::Application) {
+    let window = gtk::ApplicationWindow::new(application);
+
+    window.set_title(Some("Glium in GLArea"));
+
+    let widget = gtk::GLArea::new();
+    let renderer = Rc::new(RefCell::new(None));
+
+    widget.connect_realize({
+        let renderer = renderer.clone();
```
sdroege, line 259: Elsewhere we use the `clone!` macro in the examples. No strong opinion (I don't like the macro), but maybe others do.

bilelmoussaoui: I don't like it neither, but it's pretty convenient also you wouldn't need a clone here if it were a GLArea subclass as those are vfuncs

examples/glium_gl_area/main.rs
```diff
@@ -0,0 +1,280 @@
...
+}
+
+fn main() {
+    // Load GL pointers from epoxy (GL context management library used by GTK)
+    {
+        #[cfg(unix)]
+        let library = libloading::os::unix::Library::this();
+        #[cfg(windows)]
+        let library =
+            libloading::os::windows::Library::open_already_loaded("libepoxy-0.dll").unwrap();
```
sdroege, line 233: I think on Linux you ideally also want to use libepoxy to make sure the same GL libraries are used as GTK is using.

examples/glium_gl_area/main.rs
```diff
@@ -0,0 +1,280 @@
...
+                    color: [1., 0., 0.],
+                },
+            ],
+        )
+        .unwrap();
+        let index_buffer =
+            IndexBuffer::new(&context, PrimitiveType::TrianglesList, &[0u16, 1, 2]).unwrap();
+        let program = program!(&context,
+            140 => {
+                vertex: "
```
sdroege, line 124: Maybe you want to put the shaders in separate files an `include!` them here? That would potentially keep the code easier to read

examples/glium_gl_area/main.rs
YaLTeR: Yeah, that's what happens with this code. The `epoxy` crate loads epoxy-specific symbols which are available in the global address space.

examples/glium_gl_area/main.rs
sdroege: Especially you wouldn't need the weak ref anymore then

examples/glium_gl_area/main.rs
```diff
@@ -0,0 +1,280 @@
...
+                    varying vec3 vColor;
+                    void main() {
+                        gl_FragColor = vec4(vColor, 1.0);
+                    }
+                ",
+            },
+
+            100 => {
+                vertex: "
+                    #version 100
```
sdroege, line 170: What's the difference between the 3 shaders apart from the version? Maybe add a comment about that

examples/Cargo.toml
```diff
@@ -7,6 +7,9 @@ edition = "2018"
 [dependencies]
 chrono = "0.4.19"
 once_cell = "1.5"
+glium = "0.30.1"
+epoxy = "0.1.0"
+libloading = "0.7.0"
```
sdroege, line 12: Maybe these can be made optional only for this example so you don't need to build glium just to run the hello world example

examples/glium_gl_area/main.rs
sdroege: Makes sense but maybe add a comment about that then :)

examples/glium_gl_area/main.rs
YaLTeR: Took me a few reads to notice too. :) I'm not experienced enough in GLSL to describe the differences, I just took them verbatim from the glium example code.

What's the GTK minimum GL requirement? Maybe we can drop some of the older GLSL versions?

examples/glium_gl_area/main.rs
sdroege: But `libloading::os::unix::Library::this()` would load whatever is currently loaded, which might not be what GTK is using? Or am I missing something?

examples/Cargo.toml
YaLTeR: I'm not sure you can specify dependencies per binary? The alternative is with a non-default feature.

sdroege:
Nice example, I like it :) I think there was an issue somewhere about having a GL example. That can be closed once this is merged.

examples/Cargo.toml
sdroege: I usually use non-default features for that. You can specify the required features per binary so the user would get a message telling what needs to be enabled.

examples/glium_gl_area/main.rs
YaLTeR: GTK is always using epoxy, and the `epoxy` crate loads specifically `epoxy_` symbols as far as I can tell: https://github.com/mjkoo/epoxy-rs/blob/ebb7328dbd7f6dd394ffb80e6d6b4e5696b8d800/build/epoxy_gen.rs#L226

examples/glium_gl_area/main.rs
sdroege: If that's just taken from glium then let's keep it like that, I guess.

For GTK minimum version, no idea. @bilelmoussaoui ?

examples/glium_gl_area/main.rs
sdroege: Yes but you currently only use epoxy on Windows, not elsewhere

YaLTeR: On Linux the symbols can be `dlsym`'d without opening `epoxy` specifically (which means one less string to hardcode). Not sure about macOS.

examples/glium_gl_area/main.rs
bilelmoussaoui: No idea about this. Worth asking gtk developers about

examples/glium_gl_area/main.rs
sdroege: My point is that on Linux this also depends a lot on the environment. You global `glFoo` symbol at this point might not be the one that GTK is using. This is regularly a problem with GL on platforms like the RPi or with proprietary NVIDIA drivers.

Is there anything wrong with just using epoxy unconditionally everywhere?

examples/glium_gl_area/main.rs
YaLTeR: GTK is all over the place with its shaders, with many using version 150 and above. I think I'll leave just the 140 version in this example.

YaLTeR:
Updated according to review comments.

After subclassing I managed to get rid of two auxiliary types. Following expected type signatures, I implemented `glium::backend::Backend` on the `GliumGLArea` public type (`glium` wants it by value, so an `Rc`-like public type is perfect), while the context type turned out to not be needed.

YaLTeR:
Fixed fmt, clippy fail isn't mine

examples/glium_gl_area/main.rs
sdroege: @ebassi Maybe you know what the required minimum GLSL version for GTK4 is?

sdroege:
> Fixed fmt, clippy fail isn't mine

Indeed, new clippy warning on beta @bilelmoussaoui

examples/glium_gl_area/main.rs
```diff
@@ -0,0 +1,40 @@
...
+use gtk::prelude::*;
+
+mod glium_gl_area;
+use glium_gl_area::GliumGLArea;
+
+fn main() {
+    // Load GL pointers from epoxy (GL context management library used by GTK)
+    {
+        #[cfg(unix)]
+        let library = unsafe { libloading::os::unix::Library::new("libepoxy.so.0") }.unwrap();
```
sdroege, line 12: This won't work on macOS (`libepoxy.0.dylib` AFAIU)

sdroege (commented):
Looks good to me otherwise, and so much easier with the subclass :)

examples/glium_gl_area/main.rs
YaLTeR: I don't have macOS so I can't check if it works either way..

sdroege: Maybe just add a case for macOS with that and then someone on macOS can worry about it :)

sdroege:
> Indeed, new clippy warning on beta @bilelmoussaoui

https://github.com/gtk-rs/gtk4-rs/pull/537

examples/glium_gl_area/main.rs
ebassi: The version of the shaders depend on the version of GL; currently, GTK4 uses:

 - 100 on GLES
 - 110 on GL2 legacy
 - 130 on GL3 legacy
 - 150 on GL3.2+

In practice, the version of GLSL for the shaders *inside your application* depends on the GL context you're either creating or using—i.e. if you support multiple versions of GL then you should load different shaders.

If you only care about recent GL, as you should, then going for GLSL 1.50 is perfectly fine; anything else will error out, and you can catch that error and fall back to something else.

sdroege:
if you rebase the clippy warning should be gone

examples/glium_gl_area/main.rs
YaLTeR: Thanks!

YaLTeR: Do you mind if I include this whole comment into that example?

bilelmoussaoui:
Thanks @YaLTeR

PR #534, opened 2021-07-27
gdk: add helpers for checking which display is being used

### LINKED ISSUE #235: Unsure where to look to determine the in use backend
Hello, I've looked around and I don't see an equivalent to the `GDK_IS_X11/WAYLAND_DISPLAY` macro, wondering if I am just blindly missing it or if it has not been covered yet in the bindings.

Knowing these bindings are still relatively new, it wouldn't surprise me if those macros or their equivalent functions have not been handled yet.

bilelmoussaoui:
I'm not sure whether this approach makes sense, but I couldn't come with anything better :) 

Fixes #235

### DIFF
--- a/gdk4/src/display.rs
+++ b/gdk4/src/display.rs
@@ -1,5 +1,6 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
+use crate::prelude::*;
 use crate::{keys::Key, Display, Event, KeymapKey, ModifierType};
 use glib::translate::*;
 use glib::IsA;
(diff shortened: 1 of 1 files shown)


elmarco:
`display.downcast::<gdk_wl::WaylandDisplay>().is_ok()`?

bilelmoussaoui:
> `display.downcast::<gdk_wl::WaylandDisplay>().is_ok()`?

That works sure, but what about the win32/broadway/macos ones?

bilelmoussaoui:
That also requires adding gdk_x11/gdk_wayland dependencies

sdroege:
Checking the type name seems like a better approach here for avoiding complicated dependencies.

gdk4/src/display.rs
```diff
@@ -25,6 +26,26 @@ pub trait DisplayExtManual: 'static {
 
     #[doc(alias = "gdk_display_put_event")]
     fn put_event<P: AsRef<Event>>(&self, event: &P);
+
+    /// Equivalent to the C macro `GDK_IS_WAYLAND_DISPLAY`
+    #[doc(alias = "GDK_IS_WAYLAND_DISPLAY")]
+    fn is_wayland(&self) -> bool;
```
sdroege, line 32: Instead of a function for each, you could also add an enum and a single function. Not sure what's better.

bilelmoussaoui: That's a possibility yes. Then you would need to have similar methods in the enum at the end no?

sdroege: Why? You could do `display.backend() == gdk::Backend::Wayland` etc

PR #537, opened 2021-07-30
gdk4: Fix clippy warning for collecting an iterator into a Vec to jus…

sdroege:
…t use it as iterator again


bilelmoussaoui:
Thanks!

PR #544, opened 2021-08-04
add gif_paintable example

VanillaBrooks:
This is an example of rendering individual frames of a gif at it's encoded framerate. It is heavily based on @bilelmoussaoui's work in the incomplete #152. 

It makes some improvements on what was requested in #152 as well as removing the use of `FileChooserNative` in favor of a command line parameter since I thought it was making things significantly less clear. The code also works as expected now for the test cases that I ran, so that should also be an improvement.

### DIFF
--- a/examples/gif_paintable/main.rs
+++ b/examples/gif_paintable/main.rs
@@ -1,0 +1,239 @@
+// A custom GDK paintable capable of rendering a GIF
+// The paintable makes uses of the awesome image
+// crate to read a gif file and transform it to a Vec<Frame>
+// which are then rendered by the paintable at different snapshots
+
+use gtk::{gdk, glib, graphene};
+use gtk::prelude::*;
+use gtk::subclass::prelude::*;
+use gtk::Application;
+use gtk::ApplicationWindow;
+use gtk::Picture;
+
+use image::{gif::GifDecoder, AnimationDecoder};
+
+use std::io::{Cursor, Read};
+use std::time::Duration;
+use std::ops::Deref;
+
+pub struct Frame {
+    pub width: i32,
+    pub height: i32,
+    pub texture: gdk::Paintable,
+    pub frame_duration: Duration,
+}
+
+impl Frame {
+    pub fn new(f: image::Frame) -> Self {
+        // transform the fractional time into a `Duration`
+        let (numerator, denominator) = f.delay().numer_denom_ms();
+        let frame_duration = Duration::from_millis((numerator as f64 / denominator as f64) as u64);
+
+        let image = f.into_buffer();
+
+        let samples = image.into_flat_samples();
+        let (stride, width, height) = samples.extents();
+        
+        // glib::Bytes must contain gtk_stride * height number of bytes.
+        // since each pixel in the gif frame contains `stride` bytes 
+        // of information we must map this stride to a value that gdk 
+        // understands (see https://gtk-rs.org/gtk4-rs/git/docs/gdk4/struct.MemoryTexture.html)
+        let gtk_stride = stride * width;
+
+        let width = width as i32;
+        let height = height as i32;
+
+        let bytes = glib::Bytes::from(samples.as_slice());
+
+        let texture = gdk::MemoryTexture::new(
+            width,
+            height,
+            gdk::MemoryFormat::R8g8b8a8,
+            &bytes,
+            gtk_stride,
+        );
+
+        Frame {
+            width,
+            height,
+            texture: texture.upcast(),
+            frame_duration,
+        }
+    }
+}
+
+mod imp {
+    use super::*;
+    use glib::subclass;
+    use std::cell::{Cell, RefCell};
+
+    pub struct GifPaintable {
+        pub frames: RefCell<Option<Vec<Frame>>>,
+        pub next_frame: RefCell<Option<gdk::Paintable>>,
+        pub width: Cell<Option<i32>>,
+        pub height: Cell<Option<i32>>,
+        pub current_idx: Cell<usize>,
+    }
+
+    #[glib::object_subclass]
+    impl ObjectSubclass for GifPaintable {
+        const NAME: &'static str = "GifPaintable";
+        type Type = super::GifPaintable;
+        type ParentType = glib::Object;
+        type Instance = subclass::basic::InstanceStruct<Self>;
+        type Class = subclass::basic::ClassStruct<Self>;
+        type Interfaces = (gdk::Paintable,);
+
+        fn new() -> Self {
+            Self {
+                next_frame: RefCell::new(None),
+                frames: RefCell::new(None),
+                width: Cell::new(Some(0)),
+                height: Cell::new(Some(0)),
+                current_idx: Cell::new(0),
+            }
+        }
+    }
+
+    impl ObjectImpl for GifPaintable {}
+
+    impl PaintableImpl for GifPaintable {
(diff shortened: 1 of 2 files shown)


examples/gif_paintable/main.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+}
+
+fn main() {
+    let gif = std::env::args().into_iter().skip(1).next().expect("Expected a path to a gif as a command line argument");
+    let gif_path = std::path::PathBuf::from(&gif);
+
+    if gif_path.extension().unwrap() != "gif" {
+        panic!("Provided argument `{}` did not have a `.gif` file extension.", gif);
+    }
```
VanillaBrooks, line 215: I ran into an error with this portion of code. When running the example with a command line parameter like this I got:

```
(gif_paintable:168121): GLib-GIO-CRITICAL **: 21:27:55.092: This application can not open files.
```

as far as I can tell I need to edit the `gio::ApplicationFlags` as well as connecting to a signal but I have not had any luck in suppressing this error. Any advice here?

examples/gif_paintable/main.rs
```diff
@@ -0,0 +1,239 @@
+// A custom GDK paintable capable of rendering a GIF
+// The paintable makes uses of the awesome image
+// crate to read a gif file and transform it to a Vec<Frame>
+// which are then rendered by the paintable at different snapshots
```
VanillaBrooks, line 4: From the previous PR [here](https://github.com/gtk-rs/gtk4-rs/pull/152/files#r553270088) (I left unfixed):



> That sounds like a bad idea, especially if the GIF has loops. You can easily use up all memory like this :)
>
> It would be better to decode on demand.

I'm not too familiar with the GTK memory model here but I cant see how this would eat up a ton of memory. The file bytes are read in once and passed by reference to the decoder.

 Is there some cloning going on somewhere that would make this an issue? I can't say I expect folks to load in multi-gigabyte gifs.

sdroege: If you have a GIF with looping, you'd presumably get infinite frames from the decoder

VanillaBrooks: Ah I thought you were referring to looping in the UI which is done by indexing the frames. I actually didn't know that gifs had an internal looping mechanism. However, I think this example should prove that the decoder only decodes into the frames a single time: 
```
use image::codecs::gif;
use image::AnimationDecoder;

/// encode a single pixel to a gif that repeats infinitely
fn make_gif() -> Vec<u8> {
    let mut bytes = Vec::new();
    {
        let mut encoder = gif::GifEncoder::new(&mut bytes);
        encoder.set_repeat(gif::Repeat::Infinite).unwrap();
        encoder
            .encode(&[0, 0, 0, 0], 1, 1, image::ColorType::Rgba8)
            .unwrap();
    }
    bytes
}

/// decode a gif into the constituent frames
fn decode_to_frames(bytes: Vec<u8>) -> Vec<image::Frame> {
    let decoder = gif::GifDecoder::new(std::io::Cursor::new(bytes)).unwrap();
    let frames = decoder.into_frames();
    let frames = frames.collect_frames().expect("error decoding gif");
    frames
}

fn main() {
    let bytes = make_gif();
    let frames = decode_to_frames(bytes);
    assert_eq!(frames.len(), 1);
}

```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0418f0575168ea1d956c7044c836e69c

sdroege: That's good, so this handles looping better than the gdkpixbuf animation API where the equivalent code would just use up all your system's memory :)

sdroege: How would the application know from the `Vec<Frame>` that these have to be repeated though?

VanillaBrooks: I don't see any methods on the `GifDecoder` [here](https://docs.rs/image/0.23.14/image/codecs/gif/struct.GifDecoder.html), or any methods on the underlying `gif` crate [here](https://docs.rs/gif/0.11.2/gif/struct.Decoder.html) so I believe we are hand-waving away the looping issues. 

The code provided loops infinitely anyway.

examples/gif_paintable/main.rs
```diff
@@ -0,0 +1,239 @@
...
+        pub current_idx: Cell<usize>,
+    }
+
+    #[glib::object_subclass]
+    impl ObjectSubclass for GifPaintable {
+        const NAME: &'static str = "GifPaintable";
+        type Type = super::GifPaintable;
+        type ParentType = glib::Object;
+        type Instance = subclass::basic::InstanceStruct<Self>;
+        type Class = subclass::basic::ClassStruct<Self>;
```
bilelmoussaoui, line 84:
```suggestion
```

bilelmoussaoui:
Well, the issue with making it take a file as CLI argument is that things are not easily discoverable. I also think it's fine to assume that someone that wants to load gif files to know how to use the file choose to open a file. I would re-add the file chooser and drop the CLI option and split the files into a main.rs & a gif_paintable that contains only the paintable code.

examples/gif_paintable/main.rs
```diff
@@ -0,0 +1,239 @@
...
+        // invalidate the contents so that the new frame will be rendered
+        self.invalidate_contents();
+
+        // setup a callback to this function once the frame has finished so that
+        // we can play the next frame
+        let update_next_frame_callback = glib::clone!(@weak self as paintable => move || {
+            paintable.setup_next_frame();
+        });
+
+        glib::timeout_add_local_once(next_frame.frame_duration, update_next_frame_callback);
```
awused, line 198: I saw this PR by coincidence, and I've implemented something similar myself.

Using frame_duration directly as the delay here will cause the gif to drift slightly behind its full speed, and that'll be worse if any work is being done on the main thread that can delay the callback even slightly. The time that this frame was intended to be displayed should be stored and compared against the current time to correct for drift when scheduling the next frame.

If the application does take too long between frames then you have to choose between slowing down the gif or skipping frames, but that's probably down to taste.

examples/gif_paintable/main.rs
```diff
@@ -0,0 +1,239 @@
...
+    pub width: i32,
+    pub height: i32,
+    pub texture: gdk::Paintable,
+    pub frame_duration: Duration,
+}
+
+impl Frame {
+    pub fn new(f: image::Frame) -> Self {
+        // transform the fractional time into a `Duration`
+        let (numerator, denominator) = f.delay().numer_denom_ms();
```
awused, line 29: There's a From implementation to convert from Delay into Duration that also handles fractions of milliseconds.

Some gifs also have 0ms delays set between frames and, by convention to match browsers, they should be treated as 100ms delays.

VanillaBrooks: Thats good advice. I knew there had to be some way to get a floating point representation but I didnt see a method for it. Also didn't know about the 0ms delays either.

bilelmoussaoui:
Any updates on this one? :)

VanillaBrooks:
Ah yeah, pretty much forgot about this one. 

I remember the recommendation for the file picker over the CLI added some extra complexity to the examples that I didnt like when I was learning GTK.

Do you think the implementation should focus on correctness from the above notes, or should I just leave mentions of the subtleties @bilelmoussaoui ? Imo the example should focus on how a paintable works, not the intricacies of GIF

sdroege:
Superseded by https://github.com/gtk-rs/gtk4-rs/pull/1131

PR #547, opened 2021-08-06
Fix examples/basics.

blame2020:
This patch corrects the name space from gtk to gtk4.


bilelmoussaoui:
No this is wrong, we rename the package in Cargo.toml like explained in the docs.
```toml
gtk = {package = "gtk4", version = "0.2"}
```

Issue #548, opened 2021-08-12
DrawingArea draw func should be FnMut instead of Fn


AaronErhardt:
It seems that the only way to draw on a DrawingArea is to set a callback using `DrawingAreaExt::set_draw_func()`. But then the problem is that you can't pass any data to the drawing function. The original C interface at least has a gpointer to pass data to the function but gtk4-rs just uses this function signature `Fn(&DrawingArea, &Context, i32, i32)`. Of course you could capture the data in a closure but that'd need to be immutable or cloned which is not very convenient. Or you could probably define your own GObject just for storing information, but that doesn't sound like a good idea if you want to process the data from "normal" Rust code as well.

Maybe I've overlooked some details and there's actually an alternative solution to draw on a DrawingArea. Technically GLArea does have a `GLAreaExt::context()` method to receive the context and I wonder why this isn't available for DrawingArea as well.

sdroege:
You pass the data into the closure by capturing the variables somehow, just like with signal handlers and other callbacks

```rust
let x = your_stuff;
area.set_draw_func(move |...| {
    // do things with x
});
```

AaronErhardt:
Yeah, I know...

> Of course you could capture the data in a closure but that'd need to be immutable or cloned which is not very convenient

sdroege:
I think the draw func here could be `FnMut()`, which would solve part of your problem. If you want to send a PR for that, please go ahead. (CC @bilelmoussaoui ). Let's re-purpose this issue for that.

That's however not the case for signal handlers and there you'd have to make use of interior mutability, or otherwise if they were `FnMut()` it would be unsound.

AaronErhardt:
I don't know whether FnMut helps me here. Anyway, I spent the whole day trying to get my code to work and in the end I just realized that - unless I missed some details - GTK has nothing comparable to what I want. Just something similar to a canvas in JavaScript. There's nothing you can simply draw on without a callback.

Yet eventually I had a look at relm which implemented this DrawHandler hack for gtk3-rs: [the hack](https://github.com/antoyo/relm/blob/master/src/drawing.rs) + [example application](https://github.com/antoyo/relm/blob/master/relm-examples/examples/drawing.rs)

I didn't want to use it at first because I thought there must have been something built into GTK to properly implement this but after suffering for hours without success I guess that's the only available solution.

[...]

PR #552, opened 2021-08-17
Implement DrawingArea::set_draw_func() manually to allow FnMut closur…

sdroege:
…es instead of just Fn

Fixes https://github.com/gtk-rs/gtk4-rs/issues/548

### DIFF
--- a/gtk4/src/drawing_area.rs
+++ b/gtk4/src/drawing_area.rs
@@ -3,12 +3,19 @@
 use crate::DrawingArea;
 use glib::object::IsA;
 use glib::translate::*;
+use std::cell::RefCell;
 use std::ptr;
 
 pub trait DrawingAreaExtManual: 'static {
     #[doc(alias = "gtk_drawing_area_set_draw_func")]
     #[doc(alias = "set_draw_func")]
     fn unset_draw_func(&self);
+
+    #[doc(alias = "gtk_drawing_area_set_draw_func")]
+    fn set_draw_func<P: FnMut(&DrawingArea, &cairo::Context, i32, i32) + 'static>(
+        &self,
+        draw_func: P,
+    );
 }
 impl<O: IsA<DrawingArea>> DrawingAreaExtManual for O {
     fn unset_draw_func(&self) {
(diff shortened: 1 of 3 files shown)


sdroege:
This allows to keep some state inside the closure that is updated on every call, which is probably not very uncommon.

gtk4/src/drawing_area.rs
```diff
@@ -21,4 +28,43 @@ impl<O: IsA<DrawingArea>> DrawingAreaExtManual for O {
...
+        }
+        let draw_func = Some(draw_func_func::<P> as _);
+        unsafe extern "C" fn destroy_func<
+            P: FnMut(&DrawingArea, &cairo::Context, i32, i32) + 'static,
+        >(
+            data: glib::ffi::gpointer,
+        ) {
+            let _callback: Box<RefCell<P>> = Box::from_raw(data as *mut _);
+        }
+        let destroy_call3 = Some(destroy_func::<P> as _);
```
GuillaumeGomez, line 59: Why the `3`?

gtk4/src/drawing_area.rs
```diff
@@ -21,4 +28,43 @@ impl<O: IsA<DrawingArea>> DrawingAreaExtManual for O {
...
+        let super_callback0: Box<RefCell<P>> = draw_func_data;
```
GuillaumeGomez, line 60: Why the `0`?

gtk4/src/drawing_area.rs
sdroege: That's from the autogenerated code from gir. I didn't want to change it more than necessary.

FWIW, you named those variables like that :)

gtk4/src/drawing_area.rs
sdroege: See above

gtk4/src/drawing_area.rs
GuillaumeGomez: It's weird to have the count starting at 3. :laughing: 

Well, please clean it up a bit, it's manual code now. :p

sdroege: The reason for keeping it as similar as possible is so that it's easier to autogenerate it again later once we have support for overriding the closure trait

GuillaumeGomez (approved):

PR #568, opened 2021-09-01
Wait for tests in the worker thread to finish

ArekPiekarz:
This is a followup of https://github.com/gtk-rs/gtk3-rs/issues/616#issuecomment-892921651, where I noticed the synchronized tests in gtk4-rs may actually not run, because we didn't wait for the worker thread to finish its tasks. This commit also deduplicates the way the synchronization primitive was used.

### DIFF
--- a/gtk4/src/lib.rs
+++ b/gtk4/src/lib.rs
@@ -58,9 +58,17 @@ pub const TEXT_VIEW_PRIORITY_VALIDATE: u32 = ffi::GTK_TEXT_VIEW_PRIORITY_VALIDAT
 #[macro_use]
 mod rt;
 
-#[allow(dead_code)]
 #[cfg(test)]
-pub(crate) static TEST_THREAD_WORKER: once_cell::sync::Lazy<glib::ThreadPool> =
+fn test_synced(function: impl FnOnce() + Send + 'static) {
+    skip_assert_initialized!();
+    TEST_THREAD_WORKER
+        .push(function)
+        .expect("Failed to schedule a test call");
+    while TEST_THREAD_WORKER.unprocessed() > 0 {}
+}
+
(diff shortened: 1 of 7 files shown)


ArekPiekarz:
Do I understand correctly that the failure comes from https://github.com/gtk-rs/checker/blob/master/check_init_asserts which checks if a test function begins with `assert_initialized_main_thread`, `assert_not_initialized` or `skip_assert_initialized`? If yes, it would mean that the function `test_synced` contained in this commit would need to be added to the `check_init_asserts` script, right?

gtk4/src/lib.rs
```diff
@@ -58,9 +58,16 @@ pub const TEXT_VIEW_PRIORITY_VALIDATE: u32 = ffi::GTK_TEXT_VIEW_PRIORITY_VALIDAT
 #[macro_use]
 mod rt;
 
-#[allow(dead_code)]
 #[cfg(test)]
-pub(crate) static TEST_THREAD_WORKER: once_cell::sync::Lazy<glib::ThreadPool> =
+fn test_synced(function: impl FnOnce() + Send + 'static) {
+    TEST_THREAD_WORKER
```
sdroege, line 63:
```suggestion
    skip_assert_initialized!();
    TEST_THREAD_WORKER
```

sdroege (approved):

bilelmoussaoui (approved):

GuillaumeGomez:
Looks good to me as wel, just waiting for @sdroege's comment to be fixed.

gtk4/src/lib.rs
ArekPiekarz: Done.

bilelmoussaoui:
Thanks !

PR #570, opened 2021-09-05
Improve todo_app_1.md

tronta:
I tried to make this a clearer. What do you think?


Hofer-Julian:
Hi @tronta, thank you for your PR.

In my opinion it does not make it really clearer, also "enable the macro to find the correct widget definition for the struct" is not completely correct.

In any way, if templates are not explained good enough, then the explanation should probably be extended in the interface builder chapter anyways :)

PR #572, opened 2021-09-10
property name "margin-end" set

JXAZZE:
altered the second "margin-start" to "margin-end" for both objects

(this is my first contribution ever ... I don't know anything about the process. Hope it's okay what I've done)


Hofer-Julian:
Nice catch.
Can you please fix the same mistake here?
https://github.com/gtk-rs/gtk4-rs/blob/master/book/listings/todo_app/2/todo_row/todo_row.ui

JXAZZE:
I've changed the other file too.

Hofer-Julian:
Thanks

PR #582, opened 2021-09-19
Add more default object implementations

AaronErhardt:
Updates gtk4-rs to use the latest gir version with more default implementations: https://github.com/gtk-rs/gir/pull/1222

I wonder if I did this correctly though. Because when I updated the git submodules for the latest version of gir it seems to have updated the gir-files as well which results in a few more changes than I wanted to have in this PR.

Maybe this is fine anyway. Otherwise, how am I supposed to update gtk4-rs then? Only update the git submodule for gir?


sdroege:
> Otherwise, how am I supposed to update gtk4-rs then? Only update the git submodule for gir?

Yes, please only update the `gir` submodule for now :)

AaronErhardt:
Looks much better now :)

gtk3-rs and gtk-core-rs PRs are coming soon.

sdroege (approved):

Issue #594, opened 2021-09-27
[BUG] Builder without constructor method


AaronErhardt:
**Bug description**

Maybe I'm wrong, yet [`ConstraintLayoutChild`](https://docs.gtk.org/gtk4/class.ConstraintLayoutChild.html) is a class that doesn't have a constructor but still has a builder in gtk4-rs. 

It feels weird to have a builder but no real constructor method, especially because you can just use `Widget::builder.build()` as a workaround to get a new widget. It seems that this is caused by the design of the C library but still this is translated awkwardly by gir, I think.

I've also noticed that the same if true for

+ `FixedLayoutChild`
+ `GridLayoutChild`
+ `DragIconBuilder`
+ `DragIcon`
+ `OverlayLayoutChild`
+ `ShortcutsGroup`
+ `ShortcutsSection`
+ `ShortcutsShortcut`
+ `ShortcutsWindow`

For some of these widgets it might make sense to add a constructor method for some the builder is probably unnecessary...

sdroege:
Is it actually valid to create new instances of these yourself or should that be reserved to GTK?

AaronErhardt:
I don't have a lot of experience with the C library but I think most of these objects have no constructor method and are not supposed to be constructed manually.

sdroege:
CC @bilelmoussaoui

PR #595, opened 2021-09-27
Add more manual default implementations

AaronErhardt:
Adds `Default` implementations for the widgets that weren't covered by this previous PR https://github.com/gtk-rs/gtk4-rs/pull/582 because they had a manual `new` function.

The widgets affected by this PR are:

+ Application
+ DropDown
+ EntryBuffer
+ FileChooserDialog
+ IconPaintable
+ MessageDialog


gtk4/src/application.rs
```diff
@@ -43,3 +43,10 @@ impl Application {
         }
     }
 }
+
+impl Default for Application {
+    fn default() -> Self {
+        glib::object::Object::new::<Self>(&[])
+            .expect("Can't construct Application object with default parameters")
```
sdroege, line 50: You need to call `Self::register_startup_hook()` here too

gtk4/src/icon_paintable.rs
```diff
@@ -0,0 +1,10 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::IconPaintable;
+
+impl Default for IconPaintable {
```
sdroege, line 5: Why does this need a manual `Default` impl if it doesn't have a manual `new()`?

sdroege (commented):
Seems ok to me otherwise, thanks!

gtk4/src/icon_paintable.rs
AaronErhardt: You're right. Maybe IconPaintable should be on the list of #594 then? It's generally weird because the builder of IconPaintable has more options than its constructor. You can only construct it from a file but build it from a file and from an icon name...

AaronErhardt: I guess let's keep IconPaintable as it is for now.

sdroege:
@bilelmoussaoui ?

AaronErhardt:
Btw. a point release for this would be great. Without `Default` implementation for `MessageDialog` Relm4 CI will constantly fail because I don't want to implement a temporary workaround only for `MessageDialog`.

PR #601, opened 2021-10-01
book: Start to improve generic values chapter

Hofer-Julian:


Hofer-Julian:
@sdroege As discussed in the Matrix room, I've added variants to the generic value chapter.
Maybe you could have a look before we merge it :)

PR #603, opened 2021-10-03
Update submodules

A6GibKm:


A6GibKm:
So were exactly in the CI I need to add libclang?

gdk4-wayland/sys/Cargo.toml
```diff
@@ -9,13 +9,13 @@ keywords = ["gdk4", "gdk4-wayland", "ffi", "gtk-rs", "gnome"]
 license = "MIT"
 name = "gdk4-wayland-sys"
 repository = "https://github.com/gtk-rs/gtk4-rs"
-version = "0.3.0"
+version = "0.2.0"
```
bilelmoussaoui, line 12: is this wanted? i don't think so

bilelmoussaoui:
> So were exactly in the CI I need to add libclang?

I would include it in the Dockerfile

gdk4-wayland/sys/Cargo.toml
A6GibKm: Ugh... idk how did that went thru

Issue #616, opened 2021-10-14
[BUG] vfuncs measure and size_allocate facing a binding bug


Ungedummt:
**Bug description**

It seems like the vfuncs `measure` and `size_allocate` are not called or something like that. Even debug and print messages are not shown. The Widget is shown as normal but without any scale applied.

![grafik](https://user-images.githubusercontent.com/53372671/137291529-792f1574-b6d6-4046-866f-562c54413e3e.png)

**Expected Behaviour**

That the `measure` and `size_allocate` funcs allocating the right amount of space based on the scale factor.

C example:
![grafik](https://user-images.githubusercontent.com/53372671/137291713-b4d6780b-4cb4-4076-86e5-fe17e0479d5a.png)
With the scale 2 (Yes the scale is inverted)

![grafik](https://user-images.githubusercontent.com/53372671/137291886-52d9bfad-27ae-4ca9-b8a7-577ffefb5540.png)
With the scale 0.5

**Context**

I tried to port an C example code to Rust but faced that problem. (The C code is written by Benjamin Otte btw.)
[Here](https://gitlab.gnome.org/Ungedummt/demo-transformer) is a minimal example in Rust. The C code is in `c-demo.c`.
My goal:
![Bildschirmfoto von 2021-10-14 11-05-31-cropped](https://user-images.githubusercontent.com/53372671/137289139-7afd6ce0-07c2-43ad-8dd0-6d57512ee85c.jpg)
![Bildschirmfoto von 2021-10-13 00-10-10-cropped(2)](https://user-images.githubusercontent.com/53372671/137289160-fde32d0c-8666-4798-9781-9f6107e10b8a.jpg)

Some discusions in the rust gtk matrix room:
![grafik](https://user-images.githubusercontent.com/53372671/137292438-02ebbad7-3b49-4c39-a085-a9214ca13c92.png)

sdroege:
Closing as discussed on Matrix. The problem here is that this inherited from `adw::Box` instead of `gtk::Widget`. The former uses a layout manager, so the virtual methods are not called but instead deferred to the layout manager.

PR #634, opened 2021-10-21
Add an example for a widget that scales its child

zophiana:
This is a follow up to the discussions on the rust gnome matrix chat, lately. It is related to the issue #616 I opened. And it does the thing I wanted, like described in the issue.

BTW: credits go to Benjamin Otte (for writing the c equivalent) and Bilal Elmoussaoui & Sebastian Dröge (for helping me on the rust side)


bilelmoussaoui:
You also need to add the example to the main readme in examples/readme.md

PR #674, opened 2021-10-29
bump edition & min rust version

bilelmoussaoui:


sdroege (changes requested):
Now only need to update girl and regenerate for the SYS crates

PR #675, opened 2021-10-30
Implement FromIterator/Extend for StringList

bilelmoussaoui:


sdroege (approved):

Hofer-Julian (approved):

Hofer-Julian:
@bilelmoussaoui This sentence has to be adapted since we don't adapt `new` anymore.
https://github.com/gtk-rs/gtk4-rs/blob/b18674c03b23ca170cdd20adc7b90e2a9bed4767/book/src/list_widgets.md?plain=1#L236

Also please tag me or request me as reviewer if you modify the book :)

bilelmoussaoui:
> @bilelmoussaoui This sentence has to be adapted since we don't use `new` anymore. https://github.com/gtk-rs/gtk4-rs/blob/b18674c03b23ca170cdd20adc7b90e2a9bed4767/book/src/list_widgets.md?plain=1#L236

Any suggestions on how to write that? at worst I can drop the book commit or you can push on top of my branch with the changes you like


> Also please tag me or request me as reviewer if you modify the book :)

Sure

Hofer-Julian:
> Any suggestions on how to write that? at worst I can drop the book commit or you can push on top of my branch with the changes you like

Note that we can create `StringList` directly from an iterator over strings.

book/src/list_widgets.md
```diff
@@ -233,8 +233,7 @@ First, we add a bunch of strings to our model.
 {{#rustdoc_include ../listings/list_widgets/6/main.rs:string_list}}
 ```
 
-Note that [`StringList::new`](../docs/gtk4/struct.StringList.html#method.new) directly takes `&[&str]`.
-This means we do not have to create a custom GObject for our model anymore.
+Note that we can create a StringList directly from an iterator over strings. This means we do not have to create a custom GObject for our model anymore.
```
Hofer-Julian, line 236:
```suggestion
Note that we can create a `StringList` directly from an iterator over strings.
This means we do not have to create a custom GObject for our model anymore.
```

I typically start new sentences in new lines.
Tends to make git diff happier :)

book/src/list_widgets.md
```diff
@@ -233,7 +233,7 @@ First, we add a bunch of strings to our model.
 {{#rustdoc_include ../listings/list_widgets/6/main.rs:string_list}}
 ```
 
-Note that [`StringList::new`](../docs/gtk4/struct.StringList.html#method.new) directly takes `&[&str]`.
+Note that we can create a StringList directly from an iterator over strings.
```
Hofer-Julian, line 236:
```suggestion
Note that we can create a `StringList` directly from an iterator over strings.
```

Plus backticks (as seen in the former suggestion ;-)

book/listings/list_widgets/6/main.rs
```diff
@@ -27,17 +27,12 @@ fn build_ui(app: &Application) {
...
 
     // ANCHOR: string_list
-    // Create a `Vec<String` with number from 0 to 100_000
-    let vector: Vec<String> = (0..=100_000)
+    // Create a `StringList` with number from 0 to 100_000
+    // `StringList` implements FromIterator<String>
+    let model: StringList = (0..=100_000)
         .into_iter()
         .map(|number| format!("Task number {}", number))
         .collect();
```
sdroege, line 35: The usual pattern is `let foo = ....collect::<Type>();`. You generally use the turbofish instead of type ascriptions.

Hofer-Julian: Do you happen to have references for that?
In that case quite a few places have to be changed in the book I think.

sdroege: Yes, but I think it's also what's considered more idiomatic in Rust in general

PR #679, opened 2021-11-01
gdk-x11: generate missing functions

bilelmoussaoui:


gdk4-x11/Gir.toml
```diff
@@ -89,6 +89,19 @@ name = "Gdk.Surface"
...
+status = "generate"
+    [[object.function]]
+    name = "x11_free_compound_text"
+    ignore = true
+    [[object.function]]
+    name = "x11_free_text_list"
+    ignore = true
+    [[object.function]]
+    name = "x11_lookup_xdisplay"
+    ignore = true # TODO: manually implementation needed
```
sdroege, line 103: Why? That looks like the other functions to me :)

sdroege (approved):

gdk4-x11/Gir.toml
bilelmoussaoui: it takes a xlib::Display, which doesn't implement the glib traits. so you can't do xdisplay.to_glib_none ;-)

sdroege: I see

PR #680, opened 2021-11-01
examples: Add new widget that squeezes its child

Ungedummt:
This is another example related to scaling which can be very handy.


examples/squeezer_bin/squeezer_bin/imp.rs
```diff
@@ -0,0 +1,70 @@
+use gtk::glib;
+use gtk::gsk;
+use gtk::prelude::*;
+use gtk::subclass::prelude::*;
+
+#[derive(Debug)]
+pub struct SqueezerBin {
+    pub keep_aspect_ratio: bool,
```
bilelmoussaoui, line 8: This should be wrapped by a Cell and made into a real property of the widget. See other examples.

This way you could make a checkbox and bind it's active state to the keep-aspect-ratio property.

I think that would make the example more interesting:)

examples/squeezer_bin/squeezer_bin/imp.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+#[glib::object_subclass]
+impl ObjectSubclass for SqueezerBin {
+    const NAME: &'static str = "SqueezerBin";
+    type ParentType = gtk::Widget;
+    type Type = super::SqueezerBin;
+    type Interfaces = ();
```
bilelmoussaoui, line 16:
```suggestion
```
Not needed

examples/squeezer_bin/squeezer_bin/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+    pub struct SqueezerBin(ObjectSubclass<imp::SqueezerBin>) @extends gtk::Widget;
+}
+
+impl SqueezerBin {
+    #[allow(clippy::new_without_default)]
+    pub fn new() -> Self {
+        glib::Object::new(&[]).unwrap()
+    }
+
+    pub fn get_child_size(&self) -> ((i32, i32), (i32, i32)) {
```
bilelmoussaoui, line 16: Does this has to be public?

bilelmoussaoui (changes requested):

examples/squeezer_bin/squeezer_bin/imp.rs
Ungedummt: Yes, good idea

examples/squeezer_bin/squeezer_bin/mod.rs
Ungedummt: Dunno, I try it

Ungedummt: No

examples/squeezer_bin/squeezer_bin/imp.rs
Ungedummt: Done looks like this:
grafik
grafik

Ungedummt:
So updated the screenshots; From my side it is ready

Ungedummt:
@bilelmoussaoui should the property `keep-aspect-ratio` be a gobject one? Or is the current state ok?

bilelmoussaoui:
> @bilelmoussaoui should the property `keep-aspect-ratio` be a gobject one? Or is the current state ok?

A proper gobject property yes so that you can watch when it changes and modify the widget accordantly

Ungedummt:
So updated it, don't now if it is right way to do this but it works. Feel free to mark some mistakes I made

examples/squeezer_bin/main.rs
```diff
@@ -0,0 +1,33 @@
...
+        let mode_switch = gtk::Switch::new();
+        let switch_label = gtk::Label::new(Some("keep aspect ratio"));
+        let squeezer = SqueezerBin::new();
+
+        headerbar.pack_start(&mode_switch);
+        headerbar.pack_start(&switch_label);
+
+        mode_switch
+            .bind_property("state", &squeezer, "keep-aspect-ratio")
+            .flags(glib::BindingFlags::DEFAULT)
```
bilelmoussaoui, line 24:
```suggestion
```
it is the default :-)

examples/squeezer_bin/squeezer_bin/imp.rs
```diff
@@ -0,0 +1,111 @@
...
+
+        let child = gtk::Label::new(Some("Hello World!"));
+        child.set_parent(obj);
+
+        self.parent_constructed(obj);
+    }
+
+    fn dispose(&self, obj: &Self::Type) {
+        let child = obj.first_child().unwrap();
+        child.unparent();
```
bilelmoussaoui, line 79:
```suggestion
   if let Some(child) = obj.first_child() {
      child.unparent();
   }
```

dispose might be called multiple times, so better not panic :-)

examples/squeezer_bin/squeezer_bin/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+            .unwrap()
+            .measure(gtk::Orientation::Vertical, -1);
+
+        (
+            (horizontal_minimal, horizontal_natural),
+            (vertical_minimal, vertical_natural),
+        )
+    }
+
+    pub fn imp(&self) -> &imp::SqueezerBin {
```
bilelmoussaoui, line 42: is this needed? there is already a self.imp() provided for you.

examples/squeezer_bin/squeezer_bin/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub fn new() -> Self {
+        glib::Object::new(&[]).unwrap()
+    }
+
+    pub fn keep_aspect_ratio(&self) -> bool {
+        self.imp().keep_aspect_ratio.get()
+    }
+
+    pub fn set_keep_aspect_ratio(&self, keep_aspect_ratio: bool) {
+        self.imp().keep_aspect_ratio.set(keep_aspect_ratio);
```
bilelmoussaoui, line 22: You should check whether the property changed and then notify.
Something like
```rust
let imp =  self.imp();
if imp.keep_aspect_ratio.get() == keep_aspect_ratio {
   return;
} 

// /do what you are doing now
self.notify("keep-aspect-ratio"); // this is needed because you have explicit notify.

File: examples/squeezer_bin/squeezer_bin/mod.rs (hunk 6 above)
Comment (Ungedummt) [reply examples/squeezer_bin/squeezer_bin/mod.rs]: Done

File: examples/squeezer_bin/squeezer_bin/mod.rs (hunk 7 above)
Comment (Ungedummt) [reply examples/squeezer_bin/squeezer_bin/mod.rs]: Should be done too, but not sure

**Comment by Ungedummt** [comment]
So from my side it is ready @bilelmoussaoui

File: examples/squeezer_bin/squeezer_bin/imp.rs [hunk 8]
```diff
@@ -0,0 +1,111 @@
...
+impl WidgetImpl for SqueezerBin {
+    fn size_allocate(&self, widget: &Self::Type, width: i32, height: i32, baseline: i32) {
+        let ((_, horizontal_size), (_, vertical_size)) = widget.get_child_size();
+
+        let (mut horizontal_zoom, mut vertical_zoom) = (
+            width as f32 / horizontal_size as f32,
+            height as f32 / vertical_size as f32,
+        );
+
+        if self.keep_aspect_ratio.get() {
```
bilelmoussaoui, line 92: Same here

examples/squeezer_bin/squeezer_bin/imp.rs
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    fn set_property(
+        &self,
+        obj: &Self::Type,
+        _id: usize,
+        value: &glib::Value,
+        pspec: &glib::ParamSpec,
+    ) {
```
bilelmoussaoui, line 55: In general, I tend to prefer including `glib::{Value, ParamSpec};` in scope to make this looks less aweful

bilelmoussaoui (commented):
otherwise looks good! thank you

Ungedummt:
@bilelmoussaoui done it there too

> > In general, I tend to prefer including glib::{Value, ParamSpec}; in scope to make this looks less aweful
>
>Here too?

Ungedummt:
Ping @bilelmoussaoui @sdroege ?

examples/squeezer_bin/squeezer_bin/imp.rs
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    fn constructed(&self, obj: &Self::Type) {
+        obj.set_halign(gtk::Align::Fill);
+        obj.set_valign(gtk::Align::Fill);
+        obj.set_hexpand(true);
+        obj.set_vexpand(true);
+
+        let child = gtk::Label::new(Some("Hello World!"));
```
sdroege, line 65: It would seem nicer API-wise to not have a child built-in but to have API on the bin that allows the application to set a child. Similar to https://github.com/gtk-rs/gtk4-rs/pull/740

examples/squeezer_bin/squeezer_bin/imp.rs
```diff
@@ -0,0 +1,106 @@
...
+        child.set_parent(obj);
+
+        self.parent_constructed(obj);
+    }
+
+    fn dispose(&self, obj: &Self::Type) {
+        if let Some(child) = obj.first_child() {
```
sdroege, line 72: Should maybe just store the child in the struct?

examples/squeezer_bin/squeezer_bin/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+            return;
+        }
+
+        imp.keep_aspect_ratio.set(keep_aspect_ratio);
+        self.queue_resize();
+
+        self.notify("keep-aspect-ratio");
+    }
+
+    fn get_child_size(&self) -> ((i32, i32), (i32, i32)) {
```
sdroege, line 33: Why is that not in `imp.rs`?

examples/squeezer_bin/squeezer_bin/imp.rs
```diff
@@ -0,0 +1,178 @@
...
+    keep_aspect_ratio: Cell<bool>,
+}
+
+#[glib::object_subclass]
+impl ObjectSubclass for SqueezerBin {
+    const NAME: &'static str = "SqueezerBin";
+    type ParentType = gtk::Widget;
+    type Type = super::SqueezerBin;
+
+    fn new() -> Self {
```
sdroege, line 22: Why not `#[derive(Default)]` and then get rid of the `new()` function here?

examples/squeezer_bin/squeezer_bin/imp.rs
```diff
@@ -0,0 +1,178 @@
...
+            } else {
+                horizontal_zoom = vertical_zoom;
+            }
+        }
+
+        let transform = gsk::Transform::new()
+            .scale(horizontal_zoom, vertical_zoom)
+            .unwrap();
+
+        widget.first_child().unwrap().allocate(
```
sdroege, line 111: This call and all other calls to `first_child()` could just use the `child` you've stored internally now anyway or not?

examples/squeezer_bin/squeezer_bin/imp.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub(super) fn child(&self, _obj: &super::SqueezerBin) -> Option<gtk::Widget> {
+        self.child.borrow().clone()
+    }
+
+    pub(super) fn set_child(
+        &self,
+        obj: &super::SqueezerBin,
+        widget: Option<&impl IsA<gtk::Widget>>,
+    ) {
+        let widget = widget.map(|w| w.upcast_ref());
```
sdroege, line 146:
```suggestion
        let widget = widget.map(|w| w.as_ref());
```

maybe

examples/squeezer_bin/squeezer_bin/imp.rs
```diff
@@ -0,0 +1,171 @@
...
+            } else {
+                horizontal_zoom = vertical_zoom;
+            }
+        }
+
+        let transform = gsk::Transform::new()
+            .scale(horizontal_zoom, vertical_zoom)
+            .unwrap();
+
+        widget.child().unwrap().allocate(
```
sdroege, line 104: Now the remaining question: should those `child().unwrap()` calls handle the `None` case? I assume the widget should not just randomly explode if there is no child widget right now.

Ungedummt:
@sdroege I only see my review in "Files changes" tab in GitHub, so I don't know if you would find it / get notified. GitHub seem to be a bit buggy currently :shrug: 

I was saying take a look at line 117 in `imp.rs`, how should I handle that?

examples/squeezer_bin/squeezer_bin/imp.rs
```diff
@@ -0,0 +1,169 @@
...
+            Some(&transform),
+        );
+    }
+}
+
+impl SqueezerBin {
+    pub(super) fn child_size(&self, obj: &super::SqueezerBin) -> ((i32, i32), (i32, i32)) {
+        let (horizontal_minimal, horizontal_natural, _, _) = obj
+            .child()
+            .unwrap()
```
sdroege, line 117: I think this should just get an `if let Some(child) = obj.child() { do things } else { return ((0,0), (0,0)) }` or something like that

examples/squeezer_bin/squeezer_bin/imp.rs
```diff
@@ -0,0 +1,169 @@
...
+    fn dispose(&self, obj: &Self::Type) {
+        if let Some(child) = obj.child() {
+            child.unparent();
+        }
+    }
+}
+
+impl WidgetImpl for SqueezerBin {
+    fn size_allocate(&self, widget: &Self::Type, width: i32, height: i32, baseline: i32) {
+        let ((_, horizontal_size), (_, vertical_size)) = self.child_size(widget);
```
sdroege, line 85: Similarly this here should probably just return if there's no child?

examples/squeezer_bin/squeezer_bin/imp.rs
```diff
@@ -0,0 +1,171 @@
...
+                baseline,
+                Some(&transform),
+            );
+        }
+    }
+}
+
+impl SqueezerBin {
+    pub(super) fn child_size(&self, obj: &super::SqueezerBin) -> ((i32, i32), (i32, i32)) {
+        let child = obj.child().unwrap();
```
sdroege, line 117: Why unwrap here instead of returning zeroes?

Ungedummt:
> Yes, but were should I put this function, inside the impl SqueezerBin?

@sdroege

examples/squeezer_bin/squeezer_bin/imp.rs
sdroege: > Yes, but were should I put this function, inside the impl SqueezerBin?

Anywhere you want inside `imp.rs`. Can also be a free function outside any impl block. It's just measuring the child, doesn't access anything else.

Ungedummt:
@sdroege thanks, yes makes sense; done

sdroege (approved):

PR #683, opened 2021-11-03
implement subclassing for scrolledwindow

vhdirk:


bilelmoussaoui:
I'm pretty sure it's a final type

bilelmoussaoui:
Note that in gtk4, you're expected to subclass gtk::Widget and set whatever widget as it's child widget instead of subclassing everything

sdroege:
> I'm pretty sure it's a final type

It is, there is no struct definition in the header.

vhdirk:
Thanks, I've only just started porting to gtk4. Not having to subclass every possible widget seems much easier indeed.

ids1024:
> It is, there is no struct definition in the header.

This also seems to be properly documented in the [generated API docs](https://docs.gtk.org/gtk4/class.ScrolledWindow.html) now (assuming that's always accurate): 

> ```vala
> final class Gtk.ScrolledWindow : Gtk.Widget {
>  /* No available fields */
> }
> ```

sdroege:
I think that is only listed in the docs if the new GLib 2.70 `G_TYPE_FLAG_FINAL` is used.

PR #688, opened 2021-11-07
Add a example launcher & enable flatpak builds of examples

Ungedummt:
Lately I encountered some issues with my distro and build environment, therefore I wasn't able to test my new example. So I decided to enable flatpak builds of the examples to be able to build, run & test all examples in a sandboxed environment.

I used meson to handle the "main" build and the rest was done with cargo. I copied the app launcher example and modified it, to my needs. Currently the examples: `glium_gl_area`, `content_provider` and `clipboard`, doesn't work, but I doesn't looked into it yet.

This is currently only a draft because I want some feedback. (It also depends on #680)

#### Potential additions:
 - [ ] github action interation
 - [ ] hide main window of example app launcher after launching an app
 - [ ] run launched app in another thread
 - [ ] get rid of meson?!


bilelmoussaoui:
Do we really want to have such thing?  Most of the examples are meant as nice things to look at it if you want some code examples; not as a tech demo. That's what gtk4-demo does for us already

no strong opinion though but it will get complicated to maintain and keep updated :-)

Ungedummt:
> Do we really want to have such thing? Most of the examples are meant as nice things to look at it if you want some code examples; not as a tech demo. That's what gtk4-demo does for us already
> 
> no strong opinion though but it will get complicated to maintain and keep updated :-)

Yes thats true, but I would like to build it all within flatpak simply because of the ease of use. But its only because of the issues I've had with compiling. I agree that there will be issues with maintaining this so we should not advertise it. I would rather discribe it as a dev tool and maybe another CI test.

bilelmoussaoui:
I tend to think this is really out of the scope of gtk4-rs, it adds a ton of maintability work that I am not willing to spend on it and the examples are meant as "how can I do this in gtk4-rs" rather than, what is possible to with GTK in such case you could either use gtk4-demo or someone could build something similar yourself out of the gtk4-rs tree I guess

Ungedummt:
> and the examples are meant as "how can I do this in gtk4-rs" rather than, what is possible to with GTK

Yea makes sense

PR #690, opened 2021-11-08
Nicer variants for getters that returns glib::Value

bilelmoussaoui:
Some functions that had a `value` getter, got a `value_as` variant to cast it to a different type. Not sure how that sounds :)

### DIFF
--- a/gtk4/src/expression.rs
+++ b/gtk4/src/expression.rs
@@ -2,7 +2,7 @@
 
 use crate::ExpressionWatch;
 use glib::translate::*;
-use glib::{IsA, Object, StaticType, Type, Value};
+use glib::{value::FromValue, IsA, Object, StaticType, Type, Value};
 use std::boxed::Box as Box_;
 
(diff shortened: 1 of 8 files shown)


gtk4/src/expression.rs
```diff
@@ -121,6 +121,22 @@ impl Expression {
         }
     }
 
+    // rustdoc-stripper-ignore-next
+    /// Similar to [`Self::evaluate`] but panics if the value is of a different type.
+    #[doc(alias = "gtk_expression_evaluate")]
+    pub fn evalute_as<V: for<'b> FromValue<'b> + 'static, T: IsA<Object>>(
```
sdroege, line 127:
```suggestion
    pub fn evaluate_as<V: for<'b> FromValue<'b> + 'static, T: IsA<Object>>(
```

sdroege (commented):
Looks good to me except for that one typo :)

Issue #77, opened 2020-11-24
Remnants For Composite Template Support


BrainBlasted:
After #51 is merged, there's still some work left to do:

* [x] Find a way to use the CompositeTemplate trait's `bind_template_children` fn
as `klass.bind_template_children()`
* [x] Find a way to handle `instance_init()` so that `init_template()` can be called
from it.
* [x] Add support for signals/callbacks in templates

sdroege:
> Add support for signals/callbacks in templates

For this one there is https://github.com/gtk-rs/gtk4-rs/issues/24 https://github.com/gtk-rs/gtk-rs/issues/173 https://github.com/gtk-rs/gtk-rs/issues/128

BrainBlasted:
I don't think BuilderScope is required. See https://gnome.pages.gitlab.gnome.org/gtk/gtk/GtkWidget.html#gtk-widget-class-bind-template-callback for templates

sdroege:
That only works for C function callbacks (with pointers and stuff). Bindings should go via `GClosure`s based signal API that allows to do things safely via a `&[glib::Value]` parameter instead.

BrainBlasted:
Ah, I see

bilelmoussaoui:
Not sure if something like Vala  & Python's possibility to set the path/uri to a UI file with a decorator would be possible in how things are handled from Rust I was thinking of an attribute of CompositeTemplate to automatically generate a class_init with the proper set_template & call to bind_template_children. 

Vala:
```vala
[GtkTemplate (ui = "/org/gnome/clocks/ui/alarm-day-picker-row.ui")]
public class DayPickerRow : Hdy.ActionRow {
} 
```
Python:
```python
@Gtk.Template(resource_path='/com/github/bilelmoussaoui/Authenticator/settings.ui')
class SettingsWindow(Handy.PreferencesWindow):
    __gtype_name__ = 'SettingsWindow'
```

bilelmoussaoui:
Something I have noticed while porting some of my code to use templates is that often the id is the same name you give to the field and having to always specify it as an attribute is not great? maybe we should fallback to the field name in case the id attribute wasn't passed

sdroege:
Sounds like an easy improvement, yes

felinira:
[...]

PR #691, opened 2021-11-09
Add composite template callback macro

### LINKED ISSUE #24: Figure out how to use Gtk.BuilderScope
To allow users to bind the signals from their UI file;

Quoting slomo again

> This will need some subclass defined here in gtk so that builder can automatically connect signals/etc. Best to create an issue for that.

> slomo : i think the cleanest would be a attribute proc macro that you would put on callback functions, which then creates the actual callback based on the function signature and GValues and checking that the types match up. then the only remaining problem is to prevent people from defining arbitrary functions manually as that would allow (intentionally) causing memory corruptions from safe code, i.e. would be unsound

> slomo:  the GValue stuff can be done more or less automatically once i finished https://github.com/gtk-rs/gtk-rs/issues/48 . that would work very similar to how i want to handle GObject properties more conveniently Improve glib::Value usage by merging FromValue/FromValueOptional traits · Issue #48 · gtk-rs/gtk-rs

> slomo: Currently we have pub trait FromValueOptional<'a>: StaticType + Sized { unsafe fn from_value_optional(_: &'a Value) -> Option<Self>; } pub trait FromValue<'a>: ...

> slomo:  and that change is needed so that you can autogenerate code from/to GValues without knowing anything about the Rust types involved other than their name

> slomo: and GValues are relevant in your case too because that's how the signal machinery works. and the builder scope wants you to provide GClosures

> slomo:  having such an attribute proc macro could also make glib::Object::connect() more convenient btw, i think that would be useful

> slomo: makes sense?

jf2048:
Fixes #24, also fixes part 3 of #77. Depends on gtk-rs/gtk-rs-core#338.

While getting to this work, I noticed that calling `Builder::current_object` while inside `create_closure` caused GTK to crash because `from_glib_none` has consumed the floating reference. I changed `current_object` to a manual implementation that calls `g_object_ref` via `clone`, and that fixes it. Long-term it seems that `BuilderScopeImpl` should probably be an unsafe trait since all its methods are able to get access to the builder while the object is being constructed.

### DIFF
--- a/examples/composite_template/ex_application_window/ex_application_window.ui
+++ b/examples/composite_template/ex_application_window/ex_application_window.ui
@@ -17,27 +17,125 @@
     <child>
       <object class="GtkBox">
         <property name="orientation">vertical</property>
-        <property name="valign">center</property>
-        <property name="spacing">6</property>
-        <property name="margin-start">12</property>
-        <property name="margin-end">12</property>
-        <property name="margin-top">12</property>
-        <property name="margin-bottom">12</property>
         <child>
-          <object class="GtkLabel" id="label">
-            <property name="label">Composite Template</property>
-            <style>
-              <class name="large-title"/>
-            </style>
+          <object class="GtkBox">
+            <property name="orientation">vertical</property>
+            <property name="valign">center</property>
+            <property name="vexpand">1</property>
+            <property name="spacing">6</property>
+            <property name="margin-start">12</property>
+            <property name="margin-end">12</property>
+            <property name="margin-top">12</property>
+            <property name="margin-bottom">12</property>
+            <child>
+              <object class="GtkLabel" id="label">
+                <property name="label">Composite Template</property>
+                <style>
+                  <class name="large-title"/>
+                </style>
+              </object>
+            </child>
+            <child>
+              <object class="GtkLabel" id="subtitle_label">
+                <property name="wrap">True</property>
+                <property name="justify">center</property>
+                <style>
+                  <class name="dim-label"/>
+                </style>
+              </object>
+            </child>
           </object>
         </child>
         <child>
-          <object class="GtkLabel" id="subtitle_label">
-            <property name="wrap">True</property>
-            <property name="justify">center</property>
-            <style>
-              <class name="dim-label"/>
-            </style>
+          <object class="GtkGrid">
+            <property name="halign">center</property>
+            <property name="vexpand">1</property>
+            <property name="column-spacing">6</property>
+            <property name="row-spacing">6</property>
+            <child>
+              <object class="GtkEntry" id="entry_a">
+                <property name="text">Hello</property>
+                <property name="placeholder-text">First</property>
+                <layout>
+                  <property name="column">0</property>
+                  <property name="row">0</property>
+                </layout>
+              </object>
+            </child>
+            <child>
+              <object class="GtkEntry" id="entry_b">
+                <property name="text">World</property>
+                <property name="placeholder-text">Second</property>
+                <layout>
+                  <property name="column">1</property>
+                  <property name="row">0</property>
+                </layout>
+              </object>
+            </child>
+            <child>
+              <object class="GtkLabel">
+                <property name="halign">start</property>
+                <binding name="label">
+                  <closure type="gchararray" function="concat_strs">
+                    <constant type="gchararray">Result: </constant>
+                    <lookup type="GtkEntry" name="text">entry_a</lookup>
+                    <constant type="gchararray">, </constant>
+                    <lookup type="GtkEntry" name="text">entry_b</lookup>
+                  </closure>
+                </binding>
+                <layout>
+                  <property name="column">0</property>
+                  <property name="row">1</property>
+                  <property name="column-span">2</property>
+                </layout>
+              </object>
+            </child>
+            <child>
+              <object class="GtkLabel">
+                <property name="halign">start</property>
+                <binding name="label">
+                  <closure type="gchararray" function="concat_strs">
+                    <constant type="gchararray">Lengths: </constant>
+                    <closure type="gchararray" function="to_string">
+                      <closure type="guint64" function="strlen">
+                        <lookup type="GtkEntry" name="text">entry_a</lookup>
+                      </closure>
+                    </closure>
+                    <constant type="gchararray">, </constant>
+                    <closure type="gchararray" function="to_string">
+                      <closure type="guint64" function="strlen">
+                        <lookup type="GtkEntry" name="text">entry_b</lookup>
+                      </closure>
+                    </closure>
+                  </closure>
+                </binding>
+                <layout>
+                  <property name="column">0</property>
+                  <property name="row">2</property>
+                  <property name="column-span">2</property>
+                </layout>
+              </object>
+            </child>
+            <child>
+              <object class="GtkButton">
+                <property name="label">Reset First</property>
+                <signal name="clicked" handler="reset_entry" object="entry_a" swapped="true"/>
+                <layout>
+                  <property name="column">0</property>
+                  <property name="row">3</property>
+                </layout>
+              </object>
+            </child>
+            <child>
+              <object class="GtkButton">
+                <property name="label">Reset Second</property>
+                <signal name="clicked" handler="reset_entry" object="entry_b" swapped="true"/>
+                <layout>
+                  <property name="column">1</property>
+                  <property name="row">3</property>
+                </layout>
+              </object>
+            </child>
           </object>
         </child>
       </object>
--- a/examples/composite_template/ex_menu_button/ex_menu_button.ui
+++ b/examples/composite_template/ex_menu_button/ex_menu_button.ui
@@ -6,6 +6,7 @@
     </property>
     <child>
       <object class="GtkToggleButton" id="toggle">
+        <signal name="toggled" handler="toggle_toggled" swapped="true"/>
         <property name="child">
           <object class="GtkBox">
             <property name="spacing">6</property>
@@ -28,6 +29,7 @@
     </child>
     <child>
       <object class="GtkPopover" id="popover">
+        <signal name="closed" handler="popover_closed" swapped="true"/>
         <property name="child">
           <object class="GtkLabel">
             <property name="label">Hello from a custom child widget!</property>
--- a/examples/composite_template/ex_menu_button/imp.rs
+++ b/examples/composite_template/ex_menu_button/imp.rs
@@ -19,6 +19,7 @@ impl ObjectSubclass for ExMenuButton {
 
     fn class_init(klass: &mut Self::Class) {
         Self::bind_template(klass);
+        Self::bind_template_callbacks(klass);
     }
 
     fn instance_init(obj: &glib::subclass::InitializingObject<Self>) {
@@ -26,25 +27,21 @@ impl ObjectSubclass for ExMenuButton {
     }
 }
 
-impl ObjectImpl for ExMenuButton {
-    fn constructed(&self, obj: &Self::Type) {
-        self.parent_constructed(obj);
-
-        let popover = &*self.popover;
-        self.toggle
-            .connect_toggled(glib::clone!(@weak popover => move |toggle| {
-                if toggle.is_active() {
-                    popover.popup();
-                }
-            }));
-
-        let toggle = &*self.toggle;
-        self.popover
-            .connect_closed(glib::clone!(@weak toggle => move |_| {
-                toggle.set_active(false);
-            }));
+#[gtk::template_callbacks]
+impl ExMenuButton {
+    #[template_callback]
+    fn toggle_toggled(&self, toggle: &gtk::ToggleButton) {
+        if toggle.is_active() {
+            self.popover.popup();
+        }
     }
+    #[template_callback(name = "popover_closed")]
+    fn unset_toggle(&self) {
+        self.toggle.set_active(false);
+    }
+}
 
+impl ObjectImpl for ExMenuButton {
     // Needed for direct subclasses of GtkWidget;
     // Here you need to unparent all direct children
     // of your template.
--- a/gtk4-macros/src/lib.rs
+++ b/gtk4-macros/src/lib.rs
@@ -6,6 +6,7 @@
 
 mod attribute_parser;
 mod composite_template_derive;
+mod template_callbacks_attribute;
 mod util;
 
 use proc_macro::TokenStream;
@@ -17,22 +18,30 @@ use syn::{parse_macro_input, DeriveInput};
 /// The `template` attribute specifies where the template should be loaded
 /// from;  it can be a `file`, a `resource`, or a `string`.
 ///
+/// The `callbacks` attribute specifies this template has bound callbacks.
+/// See [`macro@template_callbacks`] for a description of how to use them.
+///
 /// The `template_child` attribute is used to mark all internal widgets
 /// we need to have programmatic access to.
 ///
 /// The `template_child` attribute can take two parameters:
-///     - `id` which defaults to the item name if not defined
-///     - `internal_child` whether the child should be accessible as an “internal-child”, defaults to `false`
+/// - `id` which defaults to the item name if not defined
+/// - `internal_child` whether the child should be accessible as an “internal-child”, defaults to `false`
 ///
 /// # Example
 ///
 /// Specify that `MyWidget` is using a composite template and load the
 /// template file the `composite_template.ui` file.
 ///
-/// Then, in the `ObjectSubclass` implementation you will need to call
-/// `bind_template` in the `class_init` function, and `init_template` in
-/// `instance_init` function.
+/// Then, in the [`ObjectSubclass`] implementation you will need to call
+/// [`bind_template`] in the [`class_init`] function, and [`init_template`] in
+/// [`instance_init`] function.
 ///
+/// [`ObjectSubclass`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/subclass/types/trait.ObjectSubclass.html
+/// [`bind_template`]: https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/subclass/widget/trait.CompositeTemplate.html#tymethod.bind_template
+/// [`class_init`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/subclass/types/trait.ObjectSubclass.html#method.class_init
+/// [`init_template`]: https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/prelude/trait.InitializingWidgetExt.html#tymethod.init_template
+/// [`instance_init`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/subclass/types/trait.ObjectSubclass.html#method.instance_init
 ///
 /// ```no_run
 /// # fn main() {}
@@ -89,4 +98,166 @@ pub fn composite_template_derive(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
     let gen = composite_template_derive::impl_composite_template(&input);
     gen.into()
+}
+
+/// Attribute macro for creating template callbacks from Rust methods.
+///
+/// Widgets with [`CompositeTemplate`] can then make use of these callbacks from within their
+/// template XML definition. The attribute must be applied to an `impl` statement of a struct.
+/// Functions marked as callbacks within the `impl` will be stored in a static array. Then, in the
+/// [`ObjectSubclass`] implementation you will need to call [`bind_template_callbacks`] in the
+/// [`class_init`] function.
+///
+/// Template callbacks can be specified on both a widget's public `impl` or on its private `impl`.
+/// Template callbacks from other types can also be used. Care must be taken to call
+/// `bind_template_callbacks` on each type in order to use its callbacks.
+///
+/// These callbacks can be bound using the `<signal>` or `<closure>` tags in the template file.
+/// Note that the arguments and return type will only be checked at run time when the method is
+/// invoked.
+///
+/// Methods can optionally take `&self` as a first parameter. In this case, the attribute
+/// `swapped="true"` will usually have to be set on the `<signal>` or `<closure>` tag in order to
+/// invoke the function correctly.
+///
+/// The following options are supported on the attribute:
+/// - `functions` makes all callbacks use the `function` attribute by default. (see below)
+///
+/// The `template_callback` attribute is used to mark methods that will be exposed to the template
+/// scope. It can take the following options:
+/// - `name` renames the callback. Defaults to the function name if not defined.
+/// - `function` ignores the first value when calling the callback and disallows `self`.  Useful
+/// for callbacks called from `<closure>` tags.
+/// - `function = false` reverts the effects of `functions` used on the `impl`, so the callback
+/// gets the first value and can take `self` again. Mainly useful for callbacks that are invoked
+/// with `swapped="true"`.
+///
+/// The `rest` attribute can be placed on the last argument of a template callback. This attribute
+/// must be used on an argument of type `&[glib::Value]` and will pass in the remaining arguments.
+/// The first and last values will be omitted from the slice if this callback is a `function`.
+///
+/// Arguments and return types in template callbacks have some special restrictions, similar to the
+/// restrictions on [`glib::closure`]. Each argument's type must implement [`glib::ToValue`]. The
+/// last argument can also be `&[glib::Value]` annotated with `#[rest]` as described above. The
+/// return type of a callback, if present, must implement [`glib::FromValue`]. Type-checking of
+/// inputs and outputs is done at run-time; if the argument types or return type do not match the
+/// type of the signal or closure then the callback will panic. To implement your own type checking
+/// or to use dynamic typing, an argument's type can be left as a [`&glib::Value`].
+/// This can also be used if you need custom unboxing, such as if the target type does not
+/// implement `FromValue`.
+///
+/// [`glib::closure`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/macro.closure.html
+/// [`glib::wrapper`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/macro.wrapper.html
+/// [`ObjectSubclass`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/subclass/types/trait.ObjectSubclass.html
+/// [`class_init`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/subclass/types/trait.ObjectSubclass.html#method.class_init
+/// [`bind_template_callbacks`]: https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/subclass/widget/trait.CompositeTemplateCallbacks.html#tymethod.bind_template_callbacks
+/// [`glib::FromValue`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/value/trait.FromValue.html
+/// [`glib::ToValue`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/value/trait.ToValue.html
+/// [`&glib::Value`]: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/value/struct.Value.html
+///
+/// # Example
+///
+/// ```no_run
+/// # fn main() {}
+/// use gtk::prelude::*;
+/// use gtk::glib;
+/// use gtk::CompositeTemplate;
+/// use gtk::subclass::prelude::*;
+///
+/// mod imp {
+///     use super::*;
+///
+///     #[derive(Debug, Default, CompositeTemplate)]
+///     #[template(file = "test/template_callbacks.ui")]
+///     pub struct MyWidget {
+///         #[template_child]
+///         pub label: TemplateChild<gtk::Label>,
+///         #[template_child(id = "my_button_id")]
+///         pub button: TemplateChild<gtk::Button>,
+///     }
+///
+///     #[glib::object_subclass]
+///     impl ObjectSubclass for MyWidget {
+///         const NAME: &'static str = "MyWidget";
+///         type Type = super::MyWidget;
+///         type ParentType = gtk::Box;
+///
+///         fn class_init(klass: &mut Self::Class) {
+///             Self::bind_template(klass);
+///             // Bind the private callbacks
+///             Self::bind_template_callbacks(klass);
+///             // Bind the public callbacks
+///             Self::Type::bind_template_callbacks(klass);
+///             // Bind callbacks from another struct
+///             super::Utility::bind_template_callbacks(klass);
+///         }
+///
+///         fn instance_init(obj: &glib::subclass::InitializingObject<Self>) {
+///             obj.init_template();
+///         }
+///     }
+///
+///     #[gtk::template_callbacks]
+///     impl MyWidget {
+///         #[template_callback]
+///         fn button_clicked(&self, button: &gtk::Button) {
+///             button.set_label("I was clicked!");
+///             self.label.set_label("The button was clicked!");
+///         }
+///         #[template_callback(function, name = "strlen")]
+///         fn string_length(s: &str) -> u64 {
+///             s.len() as u64
+///         }
+///     }
+///
+///     impl ObjectImpl for MyWidget {}
+///     impl WidgetImpl for MyWidget {}
+///     impl BoxImpl for MyWidget {}
+/// }
+///
+/// glib::wrapper! {
+///     pub struct MyWidget(ObjectSubclass<imp::MyWidget>) @extends gtk::Widget, gtk::Box;
+/// }
+///
+/// #[gtk::template_callbacks]
+/// impl MyWidget {
+///     pub fn new() -> Self {
+///         glib::Object::new(&[]).expect("Failed to create an instance of MyWidget")
+///     }
+///     #[template_callback]
+///     pub fn print_both_labels(&self) {
+///         let self_ = imp::MyWidget::from_instance(self);
+///         println!("{} {}", self_.label.label(), self_.button.label().unwrap().as_str());
+///     }
+/// }
+///
+/// pub struct Utility {}
+///
+/// #[gtk::template_callbacks(functions)]
+/// impl Utility {
+///     #[template_callback]
+///     fn concat_strs(#[rest] values: &[glib::Value]) -> String {
+///         let mut res = String::new();
+///         for (index, value) in values.iter().enumerate() {
+///             res.push_str(value.get::<&str>().unwrap_or_else(|e| {
+///                 panic!("Expected string value for argument {}: {}", index, e);
+///             }));
+///         }
+///         res
+///     }
+///     #[template_callback(function = false)]
+///     fn reset_label(label: &gtk::Label) {
+///         label.set_label("");
+///     }
+/// }
+/// ```
+#[proc_macro_attribute]
+#[proc_macro_error]
+pub fn template_callbacks(attr: TokenStream, item: TokenStream) -> TokenStream {
+    use proc_macro_error::abort_call_site;
+    let args = parse_macro_input!(attr as template_callbacks_attribute::Args);
+    match syn::parse::<syn::ItemImpl>(item) {
+        Ok(input) => template_callbacks_attribute::impl_template_callbacks(input, args).into(),
+        Err(_) => abort_call_site!(template_callbacks_attribute::WRONG_PLACE_MSG),
+    }
 }
--- a/gtk4/src/builder.rs
+++ b/gtk4/src/builder.rs
@@ -18,6 +18,20 @@ impl Builder {
         }
     }
 
+    #[doc(alias = "gtk_builder_get_current_object")]
+    #[doc(alias = "get_current_object")]
+    pub fn current_object(&self) -> Option<glib::Object> {
+        unsafe {
+            let ptr = ffi::gtk_builder_get_current_object(self.to_glib_none().0);
+            if ptr.is_null() {
+                None
+            } else {
+                glib::gobject_ffi::g_object_ref(ptr as *mut glib::gobject_ffi::GObject);
+                Some(from_glib_full(ptr))
+            }
+        }
+    }
+
     #[doc(alias = "gtk_builder_get_object")]
     #[doc(alias = "get_object")]
     pub fn object<T: IsA<Object>>(&self, name: &str) -> Option<T> {
--- a/gtk4/src/builder_rust_scope.rs
+++ b/gtk4/src/builder_rust_scope.rs
@@ -22,6 +22,17 @@ glib::wrapper! {
     ///     None
     /// });
     /// builder.set_scope(Some(&scope));
+    ///
+    /// // can also be used with template_callbacks
+    /// pub struct Callbacks {}
+    /// #[gtk::template_callbacks]
+    /// impl Callbacks {
+    ///     #[template_callback]
+    ///     fn button_clicked(button: &gtk::Button) {
+    ///         button.set_label("Clicked");
+    ///     }
+    /// }
+    /// Callbacks::add_callbacks_to_scope(&scope);
     /// # }
     /// ```
     pub struct BuilderRustScope(ObjectSubclass<imp::BuilderRustScope>)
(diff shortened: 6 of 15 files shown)


YaLTeR:
This is very cool!

I have a "UX" question: is the `, callbacks` part necessary on the `#[template]`? Maybe it can figure out what to do on its own?

gtk4-macros/src/attribute_parser.rs
```diff
@@ -32,14 +32,26 @@ pub fn parse_template_source(input: &DeriveInput) -> Result<TemplateSource> {
...
-    let (ident, v) = parse_attribute(meta)?;
+    let (ident, v) = parse_attribute(source)?;
 
-    match ident.as_ref() {
-        "file" => Ok(TemplateSource::File(v)),
-        "resource" => Ok(TemplateSource::Resource(v)),
-        "string" => Ok(TemplateSource::String(v)),
-        s => bail!("Unknown enum meta {}", s),
-    }
+    let callbacks = meta
```
bilelmoussaoui, line 37: can't we just assume there will always be callbacks? I can't see how useful it's

gtk4-macros/src/composite_template_callbacks_attribute.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+    let crate_ident = crate_ident_new();
+
+    let mut callbacks = vec![];
+    for item in items.iter_mut() {
+        if let syn::ImplItem::Method(method) = item {
+            let mut i = 0;
+            let mut attr = None;
+            while i < method.attrs.len() {
+                if method.attrs[i].path.is_ident("template_callback") {
```
bilelmoussaoui, line 33: what would the template_callback provides? can't we just assume every method on the trait is a template callback?

gtk4/src/builder_rust_scope.rs
```diff
@@ -0,0 +1,143 @@
...
+use glib::Closure;
+use std::collections::HashMap;
+use std::rc::Rc;
+
+glib::wrapper! {
+    pub struct BuilderRustScope(ObjectSubclass<BuilderRustScopePrivate>)
+        @implements BuilderScope;
+}
+
+#[derive(Default)]
```
bilelmoussaoui, line 18: i would put the private struct inside a mod imp;

gtk4/src/builder_rust_scope.rs
```diff
@@ -0,0 +1,143 @@
...
+pub struct BuilderRustScopePrivate {
+    callbacks:
+        std::cell::RefCell<HashMap<String, Rc<dyn Fn(&[glib::Value]) -> Option<glib::Value>>>>,
+}
```
bilelmoussaoui, line 22:
```suggestion
pub struct BuilderRustScopePrivate(std::cell::RefCell<HashMap<String, Rc<dyn Fn(&[glib::Value]) -> Option<glib::Value>>>>);
```
Also i would do include RefCell in-tree; create a Type for the GClosures you have here to make this part of code more readable

gtk4/src/builder_rust_scope.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+impl Default for BuilderRustScope {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl BuilderRustScope {
```
bilelmoussaoui, line 129:
```suggestion
impl RustBuilderScope {
```

examples/composite_template/ex_menu_button/imp.rs
```diff
@@ -26,25 +26,21 @@ impl ObjectSubclass for ExMenuButton {
...
-                    popover.popup();
-                }
-            }));
-
-        let toggle = &*self.toggle;
-        self.popover
-            .connect_closed(glib::clone!(@weak toggle => move |_| {
-                toggle.set_active(false);
-            }));
+#[gtk::composite_template_callbacks]
```
bilelmoussaoui, line 29:
```suggestion
#[gtk::template_callbacks]
```

Would call it something like this, but no strong opinions. The current name is just way too long :) any ideas @BrainBlasted @YaLTeR @Hofer-Julian ?

gtk4-macros/src/lib.rs
```diff
@@ -90,3 +94,90 @@ pub fn composite_template_derive(input: TokenStream) -> TokenStream {
...
+///         }
+///
+///         fn instance_init(obj: &glib::subclass::InitializingObject<Self>) {
+///             obj.init_template();
+///         }
+///     }
+///
+///     #[gtk::composite_template_callbacks]
+///     impl MyWidget {
+///         #[template_callback]
```
sdroege, line 151: It would probably be useful to also allow something like

```rust
#[template_callback(name = "name of the callback in the .builder file")]
fn some_other_name() {}
```

gtk4/src/builder_rust_scope.rs
```diff
@@ -0,0 +1,143 @@
+use crate::subclass::builder_scope::BuilderScopeImpl;
```
bilelmoussaoui, line 1:
```suggestion
use crate::subclass::prelude::*;
```

gtk4/src/builder_rust_scope.rs
```diff
@@ -0,0 +1,143 @@
...
+use crate::Builder;
+use crate::BuilderClosureFlags;
+use crate::BuilderError;
+use crate::BuilderScope;
+use glib::prelude::*;
+use glib::subclass::prelude::*;
```
bilelmoussaoui, line 7:
```suggestion
use crate::prelude::*;
```

the glib subclass's prelude is included in the gtk (crate) one

gtk4/src/builder.rs
```diff
@@ -18,6 +18,20 @@ impl Builder {
...
+    #[doc(alias = "gtk_builder_get_current_object")]
+    #[doc(alias = "get_current_object")]
+    pub fn current_object(&self) -> Option<glib::Object> {
+        unsafe {
+            let ptr = ffi::gtk_builder_get_current_object(self.to_glib_none().0);
+            if ptr.is_null() {
+                None
+            } else {
+                let object: Borrowed<glib::Object> = from_glib_borrow(ptr);
+                Some(object.as_ref().clone())
```
sdroege, line 30: Stupid floating references :) How I worked around that in other places is via

```rust
glib::gobject_ffi::g_object_ref(ptr as *mut glib::gobject_ffi::GObject);
Some(from_glib_full(ptr))
```

But your version also works of course

examples/composite_template/ex_menu_button/imp.rs
BrainBlasted: `#[gtk::template_callbacks]` sounds nice

gtk4/src/builder_rust_scope.rs
bilelmoussaoui: Also needs the famous header `// Take a look at the license at the top of the repository in the LICENSE file.`

gtk4-macros/src/composite_template_callbacks_attribute.rs
bilelmoussaoui: See @sdroege's comment a bit below :)

gtk4/src/builder_rust_scope.rs
```diff
@@ -0,0 +1,143 @@
...
+                let swapped = flags.contains(BuilderClosureFlags::SWAPPED);
+                let object = object.cloned().or_else(|| builder.current_object());
+                if let Some(object) = object {
+                    let object_weak = object.downgrade();
+                    let closure = if swapped {
+                        Closure::new_local(move |args| {
+                            let mut args = args.to_owned();
+                            let object = object_weak
+                                .upgrade()
+                                .expect("Object destroyed before closure invalidated");
```
sdroege, line 63: Maybe that should just cause the closure to return silently as if it was invalidated? But it can't really happen so I guess this is also fine

examples/composite_template/ex_menu_button/imp.rs
Hofer-Julian: +1 for `#[gtk::template_callbacks]`

gtk4-macros/src/composite_template_callbacks_attribute.rs
BrainBlasted: > can't we just assume every method on the trait is a template callback?

I don't think that's safe - if you did that, you'd need a separate `impl` for the callbacks and non-callbacks, right?

examples/composite_template/ex_menu_button/ex_menu_button.ui
```diff
@@ -6,6 +6,7 @@
     </property>
     <child>
       <object class="GtkToggleButton" id="toggle">
+        <signal name="toggled" handler="toggle_toggled" swapped="true"/>
```
Hofer-Julian, line 9: Do you always have to specify `swapped="true"` with the current implementation?

bilelmoussaoui: No, see the macro code

Hofer-Julian: Indeed

gtk4-macros/src/lib.rs
jf2048: This is supported, I added a use of it to the composite_template example.

jf2048:
> I have a "UX" question: is the `, callbacks` part necessary on the `#[template]`?

Not quite, but the only other option I could get to work was to make the `template_callbacks` macro mandatory for all uses of `derive(CompositeTemplate)`. Even for a widget with no template callbacks you would still need to put that on an empty impl. At least will the `callbacks` attribute it doesn't complicate the case where a template widget has no callbacks.

This is because the `template_callbacks` macro needs to generate a trait that can then be used to provide the `BuilderScope`. If that trait is not present then the `class_init` will fail to compile if it tries to build a scope. I tried to get it to work by having it figure it out on its own but I couldn't do it without the unstable [specialization](https://github.com/rust-lang/rust/issues/31844) feature. Taking suggestions on how this could be done otherwise...

gtk4-macros/src/attribute_parser.rs
jf2048: See comment [here](https://github.com/gtk-rs/gtk4-rs/pull/691#issuecomment-964806920)

gtk4/src/builder_rust_scope.rs
jf2048: I think `RustBuilderScope` is a little easier to understand but I left it as `BuilderRustScope` for consistency with `BuilderCScope`. Maybe someone will want (unsafe) bindings for that at some point, for speed reasons? Not sure.

gtk4/src/builder_rust_scope.rs
jf2048: I think this would not ever happen unless there was a serious internal bug in glib or in gtk-rs-core, so that's why I had it panic, maybe there should be a note there that this is an internal assert?

gtk4-macros/src/composite_template_callbacks_attribute.rs
bilelmoussaoui: I don't think anyone has an Impl of the private trait though

gtk4/src/builder_rust_scope.rs
bilelmoussaoui: The BuilderCScope is not meant for language bindings and I don't think anyone is going to write such binding, not sure if it's even possible. No strong opinions here but the RustBuilderScope is an internal thing that the user will not have to deal with it directly

gtk4-macros/src/attribute_parser.rs
bilelmoussaoui: I think it's fine to introduce a breaking change in the next release, we have few others in the pipeline (from gtk-rs-core at least)

bilelmoussaoui: cc @sdroege , what do you think?

sdroege: I don't understand what this one is about :) But breaking changes are fine if it improves usability, yes.

jf2048: Just for clarity: this would not be breaking for gtk-rs-core, it would only be a breaking change for users of gtk4-rs that use `derive(CompositeTemplate)`.

sdroege: Yeah that's not really a problem. There are other breaking changes for the next release already :)

sdroege: @bilelmoussaoui Explained this to me now. I think not requiring `callbacks` and always going into that code path is the desired behaviour for the next release. If you have callbacks in your UI file and they're not handled then that should complain loudly instead of being silently ignored like in the past.

jf2048: I'm not sure I understand. This won't affect that, what it means is that you would always have to have an empty `impl` with the `template_callbacks` macro on it. We still have no way to complain about missing callbacks unless gtk4-rs does some parsing of the builder XML itself.

sdroege: I guess then we're all talking about different things :)

What @bilelmoussaoui (and I in extension) understood is that if there's a callback inside the UI file that is not mapped to a Rust function then this would cause a panic. That would seem useful. Previously we just ignored callbacks but now we can handle them, so existing code that has callbacks inside the UI file (which were ignored before!) would now panic.

----

But your concern is about requiring `#[gtk::template_callbacks]` on the impl block always now, even if no callbacks are used. That seems suboptimal to me.

Would it be possible to derive the new trait from the existing derive macro, always? And then `#[gtk::template_callbacks]` would only call something on that trait impl to pass the callbacks to it that are then used when creating the scope?

jf2048: >Previously we just ignored callbacks but now we can handle them, so existing code that has callbacks inside the UI file (which were ignored before!) would now panic.

Has that been reported upstream? It seems that the `GtkBuilder` would want to at least throw a warning in that case. I think it would actually be possible to do XML parsing in the proc macro, if we were ok with bringing an extra dependency in there. That way we could check for missing callbacks at compile time, not just panic at run time. That might be neat.

>Would it be possible to derive the new trait from the existing derive macro, always? And then #[gtk::template_callbacks] would only call something on that trait impl to pass the callbacks to it that are then used when creating the scope?

I don't know how this could be done without having `#[gtk::template_callbacks]` generate an additional trait, which is the same problem...

bilelmoussaoui: > Has that been reported upstream? It seems that the GtkBuilder would want to at least throw a warning in that case. I think it would actually be possible to do XML parsing in the proc macro, if we were ok with bringing an extra dependency in there. That way we could check for missing callbacks at compile time, not just panic at run time. That might be neat. But type checking is another story.

My understanding is that you would run `create_closure` from the builder scope even if the closure doesn't exists which can panic?

jf2048: Hmm, I just checked and I do get a critical error logged from the builder if a handler doesn't exist. `create_closure` isn't supposed to panic on lookup failure, it just returns a `GError` to the builder which is then supposed to handle. I seem to remember Vala having something that would check signals at compile time so maybe that could be copied? (Actually it seems to be [here](https://gitlab.gnome.org/GNOME/vala/-/blob/88801eb4266a8c2d33176739c7f24b6b6cedd774/codegen/valagtkmodule.vala#L376)).

sdroege: > I don't know how this could be done without having #[gtk::template_callbacks] generate an additional trait, which is the same problem...

Me neither. Let's keep it this way for now until someone finds a way :)

sdroege: > Vala having something that would check signals at compile time so maybe that could be copied?

Let's worry about that as a second step IMHO, after this one is merged.

A6GibKm:
The `callback` argument looks out of place, if it comes to it, doing a macro `#[template_with_callbacks]` sounds better imo.

is it not possible to do something similar to what the `properties` method  (or overriding vfuncs) do where the user can simply not specify it? Or similar to how templates work where the user has to override an impl in `ObjectSubclass`
```rust
            Self::bind_template(klass);
        }
```
?

jf2048:
> is it not possible to do something similar to what the `properties` method (or overriding vfuncs) do where the user can simply not specify it?

This one only works because subclasses have to implement `ObjectImpl` anyway, so it's no problem to make a default implementation that does nothing.

>Or similar to how templates work where the user has to override an impl in ObjectSubclass

This could actually be done by making it `Self::bind_template_callbacks(klass);` or something like that. But I don't know if that is better than using a macro or an attribute.

A6GibKm:
> But I don't know if that is better than using a macro or an attribute.

In my opinion it works better, it also mirrors the C api, Anyways, in the future it can be improved to be automatic.

ids1024:
> I tried to get it to work by having it figure it out on its own but I couldn't do it without the unstable specialization feature. Taking suggestions on how this could be done otherwise...

There's a lot of black magic I'd like to try using to improve sub-classing, but would require specialization other  features that aren't stable/implemented and won't be soon. Unfortunately I don't think there's currently a way to avoid something like this.

Personally, I would find it more natural to define callbacks on the wrapper type. I might want to make a function part of the type's public API and also use it as a callback; which I think isn't too uncommon. With the current implementation I'd have to define the function for both types, with one wrapping the other.

(It's generally annoying having to deal with a separate private inner type and a wrapper type when sub-classing, but I don't know how that could be avoided.)

bilelmoussaoui:
I suppose this could use a rebase so the tests are run after merging the requires ObjectExt::watch_closure? :)

jf2048:
I've spent some time thinking about this and I'd like to rework it a little bit. My current proposal would be to drop `, callbacks` and instead have three separate macros that would all generate a `CompositeTemplateCallbacks` trait. These could then be bound in the `class_init` using the same trait method.

The three macros are distinguished by their handling of the `self` parameter on template callbacks:

- `gtk::template_callbacks_public` - For use on the public widget type, `self` is unwrapped as `Self`.
- `gtk::template_callbacks_private` - For use on the `ObjectSubclass` impl, `self` is unwrapped as `Self::Type`. 
- `gtk::template_callbacks_library` - For utility functions, using `self` is an error and the first `Value` is ignored entirely when calling the function.

Here's a sample of how this could work, note the `class_init` function:

```rs
#[gtk::template_callbacks_public]
impl MyWidget {
    #[template_callback]
    fn button_clicked(&self, button: gtk::Button) {
        // ...
    }
}

struct CallbackLibrary {}

#[gtk::template_callbacks_library]
impl CallbackLibrary {
    #[template_callback]
    fn concat(a: &str, b: &str) -> String {
        format!("{} {}", a, b)
    }
}

mod imp {
    #[derive(Debug, Default, CompositeTemplate)]
    #[template(file = "mywidget.ui")]
    pub struct MyWidget {
        // ...
    }

    #[gtk::template_callbacks_private]
    impl MyWidget {
        #[template_callback]
        fn private_button_clicked(&self, button: gtk::Button) {
            // ...
        }
    }

    #[glib::object_subclass]
    impl ObjectSubclass for MyWidget {
        const NAME: &'static str = "MyWidget";
        type Type = super::MyWidget;
        type ParentType = gtk::Widget;

            Self::bind_template(klass);
            Self::bind_template_callbacks(klass);
            Self::Type::bind_template_callbacks(klass);
            CallbackLibrary::bind_template_callbacks(klass);
        }

        fn instance_init(obj: &glib::subclass::InitializingObject<Self>) {
            obj.init_template();
        }
    }
}
```

These could also be combined into one macro using attributes if anyone thinks that would work better. It seems that the public and private distinction could also be removed if either `ValueType` was implemented on `ObjectSubclass` types, or if specialization was stable and we could use another trait that has separate impls for `ValueType` and `ObjectSubclass`.

A6GibKm:
```
             Self::bind_template_callbacks(klass);
             Self::Type::bind_template_callbacks(klass);
```
Do we need both?

sdroege:
Before merging this we should also figure out https://github.com/gtk-rs/gtk-rs-core/pull/366 .

jf2048:
> Do we need both?

Yes if we want to have callbacks on multiple different structs, with the current implementation as a trait. Possibly that could also be simplified with specialization.

bilelmoussaoui:
> The three macros are distinguished by their handling of the self parameter on template callbacks:

    gtk::template_callbacks_public - For use on the public widget type, self is unwrapped as Self.
    gtk::template_callbacks_private - For use on the ObjectSubclass impl, self is unwrapped as <Self as ObjectSubclass>::Type.
    gtk::template_callbacks_library - For utility functions, using self is an error and the first Value is ignored entirely when calling the function.

As you suggested, I would make those the same macro and have an attribute to define whether you want it to be public/private/library that defaults to public which is the most used use case.  I wonder if the last use case can be just an attribute on the the functions themselves instead of the whole trait?

jf2048:
>I wonder if the last use case can be just an attribute on the the functions themselves instead of the whole trait?

Sure, but only for `library`. The other ones only make sense for a trait since the interpretation of `self` is always the same. And maybe that name is not so good there, taking suggestions on what to call it...

bilelmoussaoui:
> >I wonder if the last use case can be just an attribute on the the functions themselves instead of the whole trait?
> 
> Sure, but only for `library`. The other ones only make sense on the impl since the interpretation of `self` is always the same. And maybe that name is not so good there, taking suggestions on what to call it...

Yes, that would be useful for things like Expressions

gtk4/src/builder_rust_scope/imp.rs
```diff
@@ -0,0 +1,120 @@
...
+                                glib::gobject_ffi::g_value_set_object(
+                                    v.to_glib_none_mut().0,
+                                    object.as_object_ref().to_glib_none().0,
+                                );
+                                v
+                            };
+                            args.push(obj_v);
+                            let len = args.len();
+                            args.swap(0, len - 1);
+                            callback(&args)
```
sdroege, line 77: For the arguments it seems fine as those are checked on our side, but for the return value is there anything along the path that makes sure that if our callback returns a wrongly typed value it doesn't lead to memory unsafety?

sdroege: See https://github.com/gtk-rs/gtk-rs-core/pull/366#discussion_r753680870 . The correct return value type can actually be checked from inside the `glib::Closure` bindings. And as said, the arguments are correctly checked already.

So I think we're actually fine here in this regard once `glib::Closure` is fixed accordingly.

jf2048:
This still needs docs, but has been updated with a new API:

- `#[gtk::template_callbacks]` - Unwraps `self` as `Self`.
- `#[gtk::template_callbacks(private)] - `Unwraps `self` as `<Self as ObjectSubclass>::Type`.
- `#[gtk::template_callbacks(functions)]` - By default, makes all callbacks use the `function` attribute. (see below)

The following are supported as arguments to `#[template_callback]`:

- `name = "func"` - Renames the function in the template.
- `function` - Ignores the first value when calling the callback, therefore doesn't allow `self`.
- `method` - Only useful with `functions`, reverts the effects of it so the callback gets the first value and can take `self` again. Mainly useful for callbacks that are invoked with `swapped="true"`.

The attribute `#[rest]` can be placed on the last argument of a template callback. This attribute must be used on an argument of type `&[glib::Value]` and will pass in the remaining arguments. The first and last values will be omitted from the slice if this callback is a `function`.

The composite_template example has been updated to make use of each of these.

gtk4/src/builder_rust_scope/imp.rs
jf2048: Now that https://github.com/gtk-rs/gtk-rs-core/pull/366 is merged, I did some testing with incorrect value types and it seems to be catching them. I can add some unit tests for it.

gtk4/src/builder_rust_scope.rs
jf2048: I want to leave it as `BuilderRustScope` just in case somebody wants to do something interesting by using a `Builder` and calling `set_scope` directly. I've seen some comments mentioning that `Builder` is still used occasionally.

gtk4-macros/src/attribute_parser.rs
jf2048: One major problem with checking the callbacks during compile time is that the path to the resource file is provided on the `CompositeTemplate` derive macro. It can't be accessed from a separate attribute macro. I don't know a good way to reconcile that without doing something weird like this:

```rust
#[derive(Default, CompositeTemplate)]
#[template(file = "my_widget.ui")]
pub struct MyWidget {
    @callbacks {
        fn my_callback(&self, /* ... */) {
            // ...
        }
    }
}
```

There is also the issue of how to type check. I don't think it can be done effectively without being able to access `GType`s from within the proc macro, which is not going to work. This may be something that is just beyond the ability of Rust macros right now.

gtk4/src/builder_rust_scope/imp.rs
sdroege: Thanks for testing!

gtk4/src/builder_rust_scope.rs
jf2048: I added a test that shows how to use this manually with `gtk::Builder`.

gtk4-macros/src/attribute_parser.rs
bilelmoussaoui: The callbacks validations at compile time is imho something nice to have in the future, not at all a blocker of this MR and it would integration somehow with the CompositeTemplate to validate that every `#[template_child]` actually exists and is of the correct type. Something for whoever wants to do that work to figure out

YaLTeR:
> - #[gtk::template_callbacks] - Unwraps self as Self.
> - #[gtk::template_callbacks(private)] - Unwraps self as <Self as ObjectSubclass>::Type.

Have you mixed these two up? Because `Self` is the private type and `<Self as ObjectSubclass>::Type` is the public type.

Anyhow, based on at least my usage, I think callbacks on the private struct will be more common, so it makes sense to make that the default, and make callbacks on the public struct require an argument.

> method - Only useful with functions, reverts the effects of it so the callback gets the first value and can take self again. Mainly useful for callbacks that are invoked with swapped="true".

I'm not sure about the naming here, `method` doesn't really tell me much and I had to look at the example to figure out what it does.

examples/composite_template/ex_application_window/ex_application_window.ui
```diff
@@ -17,27 +17,125 @@
     <child>
       <object class="GtkBox">
         <property name="orientation">vertical</property>
-        <property name="valign">center</property>
-        <property name="spacing">6</property>
-        <property name="margin-start">12</property>
-        <property name="margin-end">12</property>
-        <property name="margin-top">12</property>
-        <property name="margin-bottom">12</property>
         <child>
```
bilelmoussaoui, line 20: the diff is caused by a change in the indentation of this file; it used to use 2 spaces iirc. Maybe we should enforce that using an .editorconfig file

examples/composite_template/ex_menu_button/imp.rs
```diff
@@ -19,32 +19,29 @@ impl ObjectSubclass for ExMenuButton {
...
-                    popover.popup();
-                }
-            }));
-
-        let toggle = &*self.toggle;
-        self.popover
-            .connect_closed(glib::clone!(@weak toggle => move |_| {
-                toggle.set_active(false);
-            }));
+#[gtk::template_callbacks(private)]
```
bilelmoussaoui, line 30: As mentioned, private should probably be the default case.

gtk4-macros/src/lib.rs
```diff
@@ -90,3 +94,147 @@ pub fn composite_template_derive(input: TokenStream) -> TokenStream {
...
+/// impl MyWidget {
+///     pub fn new() -> Self {
+///         glib::Object::new(&[]).expect("Failed to create an instance of MyWidget")
+///     }
+///     #[template_callback]
+///     pub fn print_both_labels(&self) {
+///         let self_ = imp::MyWidget::from_instance(self);
+///         println!("{} {}", self_.label.label(), self_.button.label().unwrap().as_str());
+///     }
+/// }
```
bilelmoussaoui, line 209: Is this use case useful at all? Do you ever need to have the template callbacks defined on the public type's impl?

gtk4/src/builder_rust_scope/imp.rs
```diff
@@ -0,0 +1,122 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::prelude::*;
+use crate::subclass::prelude::*;
+use crate::Builder;
+use crate::BuilderClosureFlags;
+use crate::BuilderError;
+use crate::BuilderScope;
```
bilelmoussaoui, line 8:
```suggestion
use crate::{Builder, BuilderClosureFlags, BuilderError, BuilderScope};
```

gtk4/src/builder_rust_scope/imp.rs
```diff
@@ -0,0 +1,122 @@
...
+pub struct BuilderRustScope {
+    pub callbacks: RefCell<HashMap<String, Rc<Callback>>>,
+}
+
+#[glib::object_subclass]
+impl ObjectSubclass for BuilderRustScope {
+    const NAME: &'static str = "GtkBuilderRustScope";
+    type Type = super::BuilderRustScope;
+    type ParentType = glib::Object;
+    type Interfaces = (BuilderScope,);
```
bilelmoussaoui, line 28:
```suggestion
impl ObjectSubclass for BuilderRustScope {
    const NAME: &'static str = "GtkBuilderRustScope";
    type Type = super::BuilderRustScope;
    type Interfaces = (BuilderScope,);
```
it's a default nowadays

gtk4/src/subclass/widget.rs
```diff
@@ -1317,3 +1330,26 @@ where
...
+
+pub trait CompositeTemplateCallbacks {
+    const CALLBACKS: &'static [TemplateCallback];
+
+    fn bind_template_callbacks<T: WidgetClassSubclassExt>(klass: &mut T) {
+        Self::add_callbacks_to_scope(&klass.rust_template_scope());
+    }
+    fn bind_template_callbacks_prefixed<T: WidgetClassSubclassExt>(klass: &mut T, prefix: &str) {
+        Self::add_callbacks_to_scope_prefixed(&klass.rust_template_scope(), prefix);
+    }
```
bilelmoussaoui, line 1344: Is this actually useful?

gtk4/tests/builder_rust_scope.rs
```diff
@@ -0,0 +1,76 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use gtk4 as gtk;
+use gtk::prelude::*;
+use gtk::subclass::prelude::*;
+
+#[test]
+fn builder_rust_scope() {
+    gtk::init().unwrap();
```
bilelmoussaoui, line 9: that won't work, see the other tests in gtk/src/bitset_iter for a way to do it that would work properly across various tests (you can init gtk only once).

BrainBlasted:
> Anyhow, based on at least my usage, I think callbacks on the private struct will be more common, so it makes sense to make that the default, and make callbacks on the public struct require an argument.

Interesting - based on my usage it would be the opposite. While I would want to be able to access `Self`, I also need to be able to easily access the object itself when working with widgets.

YaLTeR:
Hmm, can't you go from private to public? I know you can go the other way around.

sdroege:
> Interesting - based on my usage it would be the opposite. While I would want to be able to access `Self`, I also need to be able to easily access the object itself when working with widgets.

I usually have functions taking both as parameter on the private type for that reason.

> Hmm, can't you go from private to public? I know you can go the other way around.

Yes, goes in both directions.

gtk4/src/lib.rs
```diff
@@ -58,12 +58,25 @@ pub const TEXT_VIEW_PRIORITY_VALIDATE: u32 = ffi::GTK_TEXT_VIEW_PRIORITY_VALIDAT
...
-fn test_synced(function: impl FnOnce() + Send + 'static) {
+fn test_synced<F, R>(function: F) -> R
+where
+    F: FnOnce() -> R + Send + std::panic::UnwindSafe + 'static,
+    R: Send + 'static,
+{
+    use futures_channel::oneshot;
+    use std::panic;
+
     skip_assert_initialized!();
```
bilelmoussaoui, line 69: should be moved to the top of the function, otherwise the gtk-rs-checker fails

examples/composite_template/ex_application_window/ex_application_window.ui
jf2048: It's still 2 spaces, the diff happened because this box got put a level deeper in the tree.

jf2048:
I don't feel strongly either way but there are some comments here expressing a preference for either the value type or the subclass to be a default, which are mutually exclusive. This is the least contentious naming scheme I can think of:

- `#[gtk::template_callbacks(value)]` - Goes on the wrapper type. Technically this can be placed on any type that implements `FromValue`.
- `#[gtk::template_callbacks(subclass)]` - Goes on the `ObjectSubclass` type.
- `#[gtk::template_callbacks]` - Works, but if you try to use `&self` then it throws an error asking for either the `value` or `subclass` option to be added to the impl.

Is that ok with everyone? Or is there formal way the naming can be decided?

jf2048:
>Have you mixed these two up? Because `Self` is the private type and `<Self as ObjectSubclass>::Type` is the public type.

Sorry, maybe my description is confusing. `<Self as ObjectSubclass>::Type` is what gets taken out of the `Value` and then `from_instance` is called on it to eventually resolve `self`. The public type is always stored in the `Value`.

>I'm not sure about the naming here, `method` doesn't really tell me much and I had to look at the example to figure out what it does.

Maybe it would be better to have it as `function = false`?

gtk4/tests/builder_rust_scope.rs
jf2048: Good call, I'm using that and I fixed it up to work with `#[should_panic]` as well.

gtk4/src/subclass/widget.rs
jf2048: I plan to use it in one of my projects for some callbacks that are used from a different widget class, e.g. for a callback like this:

```rust
impl MyForm {
    #[template_callback]
    pub fn save(&self) {
        // ...
    }
}
```

In the widget's own template, that would be `<signal handler="save"/>`. If a container widget wanted to also use that function and avoid naming conflicts, it could import it with a prefix and then refer to it as `<signal handler="my_form_save"/>`.

gtk4-macros/src/lib.rs
jf2048: I think it could be useful, for signal handlers that can also be called as public methods.

bilelmoussaoui:
> - `#[gtk::template_callbacks(value)]` - Goes on the wrapper type. Technically this can be placed on any type that implements `FromValue`.
>- `#[gtk::template_callbacks(subclass)]` - Goes on the `ObjectSubclass` type.
>- `#[gtk::template_callbacks]` - Works if the type of the first argument is specified manually, but if you try to use `&self` then it throws an error asking for either the `value` or `subclass` option to be added to the impl.

I don't know, the three variants seems a bit hard to figure out just from reading the "name". In most of the code I wrote using gtk-rs, I think the most common case is the second one, so why not make that one the default ?

sdroege:
I'm not writing much GTK code, but I would also use it on the `ObjectSubclass` type.

> `#[gtk::template_callbacks(value)]` - Goes on the wrapper type. Technically this can be placed on any type that implements `FromValue`.

Calling it `value` seems a bit strange, maybe `wrapper` would be better?

A6GibKm:
> I don't know, the three variants seems a bit hard to figure out just from reading the "name". In most of the code I wrote using gtk-rs, I think the most common case is the second one, so why not make that one the default ?

I would think the most common case would be `#[gtk::template_callbacks(value)]`, that gives access to self, the private class and the widget that was connected.

A6GibKm:
I ported both Lorem and Emblem to use builder scope. I had to deal with the following problem

I wanted to do
```rust
    #[gtk::template_callbacks(subclass)]
    impl MyWidget {
        #[template_callback]
        fn on_notify(&self, button: crate::MySubWidget) {
            // ...
        }
```

on 
```xml
  <template class="MyWidget">
      <!-- ... -->
                    <property name="child">
                      <object class="MySubWidget">
                        <signal name="notify::prop" handler="on_notify"/>
                      </object>
                    </property>
```

and I got the error

```
thread 'main' panicked at 'Wrong type for `self` in template callback `on_notify`: WrongValueType(ValueTypeMismatchError { actual: MySubWidget, requested: MyWidget })', src/test.rs:86:5
```

This makes me think that the macro is not properly passing `self` as the first parameter when `self` is present, and instead it is passing it as the last (builder scope pass the template class as the gdata gpointer).

BrainBlasted:
> thread 'main' panicked at 'Wrong type for `self` in template callback `on_notify`: WrongValueType(ValueTypeMismatchError { actual: MySubWidget, requested: MyWidget })', src/test.rs:86:5

What happens if you do this?

```xml
                        <signal name="notify::prop" handler="on_notify" swapped="yes/>
```

Without `swapped` being yes, GTK will be expecting `MySubWidget` to be the first argument, and the template class as the last. This is how it works in C.

A6GibKm:
> > thread 'main' panicked at 'Wrong type for `self` in template callback `on_notify`: WrongValueType(ValueTypeMismatchError { actual: MySubWidget, requested: MyWidget })', src/test.rs:86:5
> 
> What happens if you do this?
> 
> ```xml
>                         <signal name="notify::prop" handler="on_notify" swapped="yes/>
> ```
> 
> Without `swapped` being yes, GTK will be expecting `MySubWidget` to be the first argument, and the template class as the last. This is how it works in C.

That works, kinda. In that case yeah, you can have self as first parameter but the other interesting argument, the connected object, goes last which is not very ergonomic in certain signals with multiple (possibly useless) arguments

BrainBlasted:
That's exactly how it works in C.

This is just my opinion, but I don't think the _behavior_ here should be to different from C.

A6GibKm:
Well but C does not have `self` as part of the languaje, without swapped=yes, you cannot make use of it. My preference would be just to pass self as self always and the pass the rest of the arguments like C.

BrainBlasted:
In C it is extremely common to cast the last arg (`user_data`) to `Self`.

```c
static void
some_action (GSimpleAction *action,
             GVariant      *parameter,
             gpointer       user_data)
{
  MyObj *self = MY_OBJ (user_data);
  
  // ...
}
 ```

I've also rarely seen people use the type outright instead of the `gpointer` - though this is more common with `swapped=yes`:
 
 ```c
static void
some_action (GSimpleAction *action,
              GVariant      *parameter,
              MyObj         *self)
{
  // ...
}
 ```

[Here is an example](https://gitlab.gnome.org/GNOME/calls/-/blob/42b5b9d6001c243af4a03ac6b7fa41fc329e0ec0/src/calls-call-window.c#L157).

`swapped` or not doesn't change the existence of `self` - it's still `gpointer user_data`, but convention has it so that people just use `self` like they would on a member function when connecting signals as swapped.

**Comment by bilelmoussaoui** [comment]
> > * `#[gtk::template_callbacks(value)]` - Goes on the wrapper type. Technically this can be placed on any type that implements `FromValue`.
> > * `#[gtk::template_callbacks(subclass)]` - Goes on the `ObjectSubclass` type.
> > * `#[gtk::template_callbacks]` - Works if the type of the first argument is specified manually, but if you try to use `&self` then it throws an error asking for either the `value` or `subclass` option to be added to the impl.
> 
> I don't know, the three variants seems a bit hard to figure out just from reading the "name". In most of the code I wrote using gtk-rs, I think the most common case is the second one, so why not make that one the default ?



Thinking about this and after giving it a bit of try, i think we should only keep two variants: 

- The default: `[gtk::template_callbacks]` goes either on the public/private type (to be decided later)
- `#[gtk::template_callbacks(utilities)]` meant for shared functions to be used with expressions mainly

**Comment by sdroege** [comment]
> The default: `[gtk::template_callbacks]` goes either on the public/private type (to be decided later)

Well we need to decide that before merging this, or not?

> `#[gtk::template_callbacks(utilities)]` meant for shared functions to be used with expressions mainly

This would be for plain/associated functions that don't take any `&self`?

**Comment by bilelmoussaoui** [comment]
> > The default: `[gtk::template_callbacks]` goes either on the public/private type (to be decided later)
> 
> Well we need to decide that before merging this, or not?
> 

Yes, let's figure out a way to vote for this ?

 `#[gtk::template_callbacks(utilities)]` meant for shared functions to be used with expressions mainly
> 
> This would be for plain/associated functions that don't take any `&self`?

Yep

**Comment by sdroege** [comment]
> Yes, let's figure out a way to vote for this ?

There are lots of websites to set up votes, or we can just do it on Matrix.

But as people apparently want to do both variants it seems a bit restrictive to disallow the other completely. We should probably select a default but keep the other possible.

**Comment by A6GibKm** [comment]
They are equivalent, it just happens that one will always be a tiny bit more comfortable to use depending on the usecase. I would also prefer to have just one for the sake of simplicity.

**Comment by jf2048** [comment]
> My preference would be just to pass self as self (if present) always and then pass the rest of the arguments like C.

This can't be done with the way the GTK4 API is now, the callbacks are declared during class initialization before any instances exist, and are then invoked from C as part of the `GtkBuilder` where the only source of the `self` reference is from the closure itself. It's mentioned in the documentation that you usually need to place `swapped="true"` if you want to use a `self` parameter.

**Comment by Cogitri** [comment]
I've just ported Health to use composite template callbacks, that sure made my code a lot nicer! I exclusively used `#[gtk::template_callbacks(value)]`, there are only a handful of functions which don't need the wrapped type for me.  A few things that I've noticed (a few of them have already been talked about on Matrix, I just figured it'd be good to document them here):

* Signals like `GtkSpinButtton`'s `::input` can't be connect via the current macros currently since it uses an out pointer for returning a value.
* `gtk::Inhibit` currently doesn't implement `ToValue`, so one has to use `bool` instead.

**Comment by jf2048** [comment]
Thanks for the testing and feedback! Let's try to improve some of it:

> * Signals like `GtkSpinButtton`'s `::input` can't be connect via the current macros currently since it uses an out pointer for returning a value.

This can't be done safely, it can be done with unsafe, but I'm going to update the docs to clarify some of the behavior here and to match the docs for `glib::closure!`. I'll also mention that `Value` can be used as an argument type if custom unboxing is needed, e.g. for pointer types.

> * `gtk::Inhibit` currently doesn't implement `ToValue`, so one has to use `bool` instead.

Good catch, here's another PR for that: https://github.com/gtk-rs/gtk-rs-core/pull/430

> * Apparently the type of the parameter has to match exactly - specifying a superclass doesn't work

This should actually work, I think it's not working because [`UnitSpinButton`](https://gitlab.gnome.org/World/Health/-/blob/b3975ae6efe7916cf03fabd2190457be0a70f186/src/widgets/unit_spinbutton.rs#L249) is not a `SpinButton`, it's an `AdwBin`? I changed one of the unit tests to use `Widget` and then downcast it, and that seems to work.

>Same for `gtk::ResponseType` which instead requires you to use `i32` and then manually do `gtk::ResponseType::from_glib()` in the function.

Unfortunately the type in the C signal for `response` is a `G_TYPE_INT` so there's not much we can do here. For signals that use the right enum type, it should work as expected.

File: gtk4-macros/src/lib.rs [hunk 22]
```diff
@@ -17,12 +18,15 @@ use syn::{parse_macro_input, DeriveInput};
 /// The `template` attribute specifies where the template should be loaded
 /// from;  it can be a `file`, a `resource`, or a `string`.
 ///
+/// The `callbacks` attribute specifies this template has bound callbacks.
+/// See `template_callbacks` for a description of how to use them.
```
bilelmoussaoui, line 22: Could use a proper doc link here

bilelmoussaoui:
Overall I'm happy with the current situation 🙂

Would be nice to figure out a way to define what could be the default type of the callbacks between subclass/value

Cogitri:
> This should actually work, I think it's not working because UnitSpinButton is not a SpinButton, it's an AdwBin? I changed one of the unit tests to use Widget and then downcast it, and that seems to work.

Erm right, was tricked by my own bad naming :D

Thanks for the response :)

jf2048:
I made a PR to add a trait that removes the need for `(value)` and `(subclass)`: https://github.com/gtk-rs/gtk-rs-core/pull/435

Tentative changes to this patch are here: https://github.com/gtk-rs/gtk4-rs/commit/f2a7b65b595d7edd7afd49979fc34b284bd79422

bilelmoussaoui:
Thanks @jf2048 for your work & everyone for their reviews! If there's something else you would like to see improved/fixed please open a new issue / merge request.

liferooter:
I really can't wait for release with this PR merged. It's so hard to live without this feature after GTK with Python or Vala

Issue #693, opened 2021-11-10
[FEATURE REQUEST] Add convenience methods for GtkExpressions


SeaDve:
Something like this:

```rust
let list_item_expression = ConstantExpression::new(list_item);
let integer_object_expression = PropertyExpression::new(
    gtk::ListItem::static_type(),
    Some(&list_item_expression),
    "item",
);
let number_expression = PropertyExpression::new(
    IntegerObject::static_type(),
    Some(&integer_object_expression),
    "number",
);
```

could be shorten into this:
```rust
list_item.property_expression("list-item")
    .property_expression("item")
    .property_expression("number")
```

A possible implementation:
https://github.com/SeaDve/Noteworthy/blob/637f20ec9f4a9d6868f26328d8a8f8c324dd313b/src/utils.rs#L28-L71

sdroege:
@SeaDve Do you want to provide a PR for that? Seems useful to me in principle and it's more efficient to discuss around an actual implementation :)

SeaDve:
Sure

SeaDve:
I gave it a try. Should it be another trait and be included on /gtk4/src/expression.rs ?

sdroege:
If you can implement it without adding a new trait then that would be great, if it needs a new trait then that's just how it is :)

Code should be in `src/expression.rs` or `src/property_expression.rs` / `src/closure_expression.rs` I guess. Wherever it fits best.

PR #694, opened 2021-11-10
Add PropExpr and ChainExpr as convenience traits to chain GtkExpressions

SeaDve:
See #693 

the closure expression api could be futher simplified by instead of having:

```rust
.closure_expression(|args| {
    let last_modified: DateTime = args[1].get().unwrap();
    gettext!("Last edited {}", last_modified.fuzzy_display())
})
```

it could be 

```rust
.closure_expression(|last_modified| {
    gettext!("Last edited {}", last_modified.fuzzy_display())
})
```

but I'm not sure how to implement it with generics


sdroege:
> but I'm not sure how to implement it with generics

You can't but see https://github.com/gtk-rs/gtk-rs-core/pull/338

gtk4/src/expression.rs
```diff
@@ -182,6 +182,56 @@ impl glib::value::ToValueOptional for Expression {
     }
 }
 
+/// Trait containing convenience methods in creating [`gtk::PropertyExpression`] that
```
bilelmoussaoui, line 185: all the docs here should be prefixed with rustdoc-stripper-ignore-next comment; look for it on other codes ;-)

bilelmoussaoui:
also cc @jsparber as you have had to use expressions before in fractal-next

gtk4/src/expression.rs
```diff
@@ -182,6 +182,56 @@ impl glib::value::ToValueOptional for Expression {
...
+    fn property_expression(&self, property_name: &str) -> Expression;
+
+    /// Create a [`gtk::ClosureExpression`] with self as a parameter
+    fn closure_expression<F, T>(self, f: F) -> Expression
+    where
+        F: Fn(&[glib::Value]) -> T + 'static,
+        T: glib::value::ValueType;
+}
+
+impl ChainExpr for Expression {
```
sdroege, line 221:
```suggestion
impl<T: AsRef<Expression>> ChainExpr for T {
```

Maybe? That way it can be directly called on `PropertyExpression` / `ClosureExpression` etc? Or is that not needed because they're both `Deref<Target=Expression>`?

gtk4/src/expression.rs
```diff
@@ -182,6 +182,56 @@ impl glib::value::ToValueOptional for Expression {
     }
 }
 
+/// Trait containing convenience methods in creating [`gtk::PropertyExpression`] that
+/// looks up a property of a [`glib::Object`].
+pub trait PropExpr {
```
sdroege, line 187: The traits should all go into the prelude

gtk4/src/expression.rs
```diff
@@ -182,6 +182,56 @@ impl glib::value::ToValueOptional for Expression {
...
+    /// Create an expression looking up an object's property
+    fn property_expression(&self, property_name: &str) -> Expression;
+
+    /// Create an expression looking up an object's property with a weak reference.
+    fn weak_property_expression(&self, property_name: &str) -> Expression;
+}
+
+impl<T: IsA<glib::Object>> PropExpr for T {
```
sdroege, line 195: I'm not sure about the trait name. `GObjectPropertyExpressionExt` maybe?

gtk4/src/expression.rs
```diff
@@ -182,6 +182,56 @@ impl glib::value::ToValueOptional for Expression {
...
+
+    fn weak_property_expression(&self, property_name: &str) -> Expression {
+        let obj_expr = crate::ObjectExpression::new(self).upcast();
+        crate::PropertyExpression::new(T::static_type(), Some(&obj_expr), property_name).upcast()
+    }
+}
+
+/// Trait containing convenience methods to create chains of [`gtk::PropertyExpression`]
+/// or [`gtk::ClosureExpression`].
+pub trait ChainExpr {
```
sdroege, line 209: `ExpressionChainExt` maybe? But this could both also be directly added into an `impl Expression` block with exactly the same behaviour without having to define a new trait, or not?

gtk4/src/expression.rs
```diff
@@ -182,6 +182,56 @@ impl glib::value::ToValueOptional for Expression {
...
+    /// Create a [`gtk::PropertyExpression`] that looks up for `property_name`
+    /// with self as parameter
+    fn property_expression(&self, property_name: &str) -> Expression;
```
sdroege, line 212:
```suggestion
    fn property_expression(&self, property_name: &str) -> PropertyExpression;
```

Maybe? And `ClosureExpression` for the other function

gtk4/src/expression.rs
SeaDve: I think that could work too, but `closure_expression` and `property_expression` as method name is not pretty clear without docs. At least with the trait, it is signaled by importing the trait

gtk4/src/expression.rs
SeaDve: I think this won't work on ClosureExpression as its parameter needs an owned Expression, unless cloning it

gtk4/src/expression.rs
sdroege: The trait would be part of the prelude though.

So maybe we need different method names here?

gtk4/src/expression.rs
sdroege: They're just refcounted so that's not necessarily a problem

gtk4/src/expression.rs
SeaDve: also with the `PropExpr/GObjectPropertyExpressionExt` i think?

sdroege: How do you mean?

gtk4/src/expression.rs
SeaDve: what about `Expression::chain_property` and `Expression::chain_closure`?

gtk4/src/expression.rs
SeaDve: the `weak_property_expression` and the `property_expression` of the other trait should also return PropertyExpression?

gtk4/src/expression.rs
sdroege: I like that. @bilelmoussaoui ?

gtk4/src/expression.rs
sdroege: Oh, yes

gtk4/src/expression.rs
SeaDve: Done, i think :D

gtk4/src/expression.rs
SeaDve: Okay

gtk4/src/expression.rs
SeaDve: I think that's clearer

gtk4/src/expression.rs
SeaDve: Done :D

gtk4/src/prelude.rs
```diff
@@ -18,6 +18,8 @@ pub use crate::drawing_area::DrawingAreaExtManual;
 pub use crate::editable::EditableExtManual;
 pub use crate::entry::EntryExtManual;
 pub use crate::entry_buffer::EntryBufferExtManual;
+pub use crate::expression::ExpressionChainExt;
+pub use crate::expression::GObjectPropertyExpressionExt;
```
sdroege, line 22:
```suggestion
pub use crate::expression::{ExpressionChainExt, GObjectPropertyExpressionExt};
```

gtk4/src/expression.rs
```diff
@@ -182,6 +182,108 @@ impl glib::value::ToValueOptional for Expression {
...
+///     list_item.property_expression("item")
+///         .property_expression("number")
+///         .closure_expression(|args| {
+///             let number: i32 = args[1].get().unwrap();
+///             format!("Number: {}", number)
+///         })
+///         .bind(&button, "label", Some(&button));
+/// });
+/// ```
+pub trait ExpressionChainExt {
```
sdroege, line 259: Now instead of a trait, can this just be in `impl Expression` and then it works the same?

SeaDve: I agree with it, but changing the methods name to make it more clear

gtk4/src/expression.rs
```diff
@@ -182,6 +182,108 @@ impl glib::value::ToValueOptional for Expression {
...
+    fn property_expression(&self, property_name: &str) -> crate::PropertyExpression {
+        crate::PropertyExpression::new(self.as_ref().value_type(), Some(self), property_name)
+    }
+
+    fn closure_expression<F, R>(self, f: F) -> crate::ClosureExpression
+    where
+        F: Fn(&[glib::Value]) -> R + 'static,
+        R: glib::value::ValueType,
+    {
+        crate::ClosureExpression::new(&[self.as_ref().clone()], f)
```
sdroege, line 283: @bilelmoussaoui That function should probably take a `impl IntoIterator<Item = &'a Expression>` or so as argument instead of what it does now, but independent of this PR.

sdroege:
Looks all good to me otherwise

gtk4/src/expression.rs
```diff
@@ -182,6 +182,114 @@ impl glib::value::ToValueOptional for Expression {
...
+/// Trait containing convenience methods in creating [`gtk::PropertyExpression`] that
+/// looks up a property of a [`glib::Object`].
+///
+/// # Example
+///
+/// `label_expression` is a [`gtk::Expression`] that looks up at Button's `label`
+/// property.
+///
+/// ```no_run
+/// use gtk4::prelude::*;
```
Hofer-Julian, line 195:
```suggestion
/// use gtk::prelude::*;
```

We typically expect that people rename it to `gtk`

SeaDve: I think that will make the tests fail

gtk4/src/expression.rs
bilelmoussaoui: Looks good to me as well

gtk4/src/expression.rs
bilelmoussaoui: It won't, if you add a little hidden comment (see gtk4-macros) that does `use gtk4 as gtk;`

gtk4/src/expression.rs
```diff
@@ -182,6 +182,114 @@ impl glib::value::ToValueOptional for Expression {
...
+    fn weak_property_expression(&self, property_name: &str) -> crate::PropertyExpression;
+}
+
+impl<T: IsA<glib::Object>> GObjectPropertyExpressionExt for T {
+    fn property_expression(&self, property_name: &str) -> crate::PropertyExpression {
+        let obj_expr = crate::ConstantExpression::new(self);
+        crate::PropertyExpression::new(T::static_type(), Some(&obj_expr), property_name)
+    }
+
+    fn weak_property_expression(&self, property_name: &str) -> crate::PropertyExpression {
```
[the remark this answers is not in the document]
SeaDve: `property_expression_weak` or `property_expression_weak_ref` maybe?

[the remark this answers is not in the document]
SeaDve: or `property_expression` on a WeakRef<glib::Object>

gtk4/src/expression.rs
SeaDve: Renamed it to chain_property and chain_closure

gtk4/src/expression.rs
SeaDve: Done :D

gtk4/src/expression.rs
```diff
@@ -146,6 +146,69 @@ impl Expression {
...
+    ///         })
+    ///         .bind(&button, "label", Some(&button));
+    /// });
+    /// ```
+    pub fn chain_closure<F, R>(&self, f: F) -> crate::ClosureExpression
+    where
+        F: Fn(&[glib::Value]) -> R + 'static,
+        R: glib::value::ValueType,
+    {
+        crate::ClosureExpression::new(&[self.clone()], f)
```
SeaDve, line 210: I think this could now take in an owned self?

sdroege: Good idea

gtk4/src/expression.rs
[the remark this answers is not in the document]
sdroege: `property_expression_weak` seems fine to me

gtk4/src/expression.rs
SeaDve: hmm, this causes `cannot move out of dereference of PropertyExpression`

sdroege: Oh well, let's keep the clone then and add a FIXME comment

gtk4/src/expression.rs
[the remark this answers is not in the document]
SeaDve: gone with `property_expression_weak`

sdroege (approved):

sdroege:
Now we only need a useful example that makes use of expressions

SeaDve:
> Now we only need a useful example that makes use of expressions

The book has some

A6GibKm:
> The book has some

Not using these convenience traits :). Besides the examples work as documentation for gtk-rs and this will be hard already to discover feature for people who are not closely following the book.

gtk4/src/expression.rs
```diff
@@ -182,6 +246,46 @@ impl glib::value::ToValueOptional for Expression {
...
+/// ```no_run
+/// # use gtk4 as gtk;
+/// use gtk::prelude::*;
+///
+/// let button = gtk::Button::new();
+/// button.set_label("Label property");
+///
+/// let label_expression = button.property_expression("label");
+/// ```
+pub trait GObjectPropertyExpressionExt {
```
A6GibKm, line 267: Does it make sense to rename this to `PropertyExpressionExt`?

gtk4/src/expression.rs
```diff
@@ -182,6 +246,46 @@ impl glib::value::ToValueOptional for Expression {
...
+/// Trait containing convenience methods in creating [`gtk::PropertyExpression`] that
+/// looks up a property of a [`glib::Object`].
+///
+/// # Example
+///
+/// `label_expression` is a [`gtk::Expression`] that looks up at Button's `label`
+/// property.
+///
+/// ```no_run
+/// # use gtk4 as gtk;
```
A6GibKm, line 259: While this is used elsewhere too, I think just using `use gtk4::prelude::*;` works better.

gtk4/src/expression.rs
sdroege: Not, it's not an extension trait for `PropertyExpression` but one for `glib::Object`

Hofer-Julian:
> > The book has some
> 
> Not using these convenience traits :). Besides the examples work as documentation for gtk-rs and this will be hard already to discover feature for people who are not closely following the book.

Rest assured, that the book will make use of this after it has been merged 😉

But I agree, examples would still be nice

SeaDve:
I do agree too with more examples

gtk4/src/expression.rs
bilelmoussaoui: No, we do the rename everywhere else

sdroege:
@bilelmoussaoui Let's get this in for now, and examples can happen later?

There's the color example as a PR currently which uses expressions, that could hopefully be moved to this helper API

SeaDve:
I'll follow up an example tomorrow if that's okay

Hofer-Julian:
> 
> 
> I'll follow up an example tomorrow if that's okay

Even better :)

SeaDve:
Should I include it in this PR? Ill squash this one then append another commit for the example

SeaDve:
I found a possible problem with `chain_property` while working with the example. Sometimes it is needed to specify the value type. For example 
```rust
list_item
    .property_expression("item")
    .chain_property("number")
```
this won't work because a gtk::ListItem's item will evaluate into a GObject, thus it doesn't have a number property unless it is downcasted into a type that has a property of number.

To solve this, the `this_type` param for `PropertyExpression` will be exposed:
```rust
list_item
    .property_expression("item")
    .chain_property(TypeThatHasNumberProp::static_type(), "number")
```

examples/expressions/main.rs
```diff
@@ -0,0 +1,84 @@
...
+            .property_expression("item")
+            .chain_property(Note::static_type(), "metadata");
+
+        metadata_expression
+            .chain_property(Metadata::static_type(), "title")
+            .chain_closure(|args| {
+                let title: String = args[1].get().unwrap();
+                format!("Title: {}", title)
+            })
+            .bind(&title_label, "label", None::<&gtk::Widget>);
```
Hofer-Julian, line 48:
```suggestion
            .bind(&title_label, "label", gtk::NONE_WIDGET);
```

We chould use the const here.
No strong preference though.

examples/expressions/main.rs
```diff
@@ -0,0 +1,84 @@
...
+
+fn data() -> gio::ListStore {
+    let list_store = gio::ListStore::new(Note::static_type());
+
+    for i in 0..100 {
+        let metadata = Metadata::new();
+        metadata.set_property("title", format!("Note ({})", i));
+
+        let note = Note::new(&metadata);
+        list_store.append(&note);
```
Hofer-Julian, line 80: You will want to fill a vector and then add it to the list store in one go in order to prevent 100 signal emissions.
See e.g. https://github.com/gtk-rs/gtk4-rs/blob/f14d7eb1edbbd33b9de2820f5f88a9a52688be96/book/listings/list_widgets/5/main.rs#L34

examples/expressions/main.rs
```diff
@@ -0,0 +1,84 @@
...
+
+        list_item.set_child(Some(&hbox));
+    });
+
+    let model = gtk::NoSelection::new(Some(&data()));
+    let list_view = gtk::ListView::new(Some(&model), Some(&factory));
+
+    let scrolled_window = gtk::ScrolledWindow::new();
+    scrolled_window.set_child(Some(&list_view));
+    scrolled_window.set_policy(gtk::PolicyType::Never, gtk::PolicyType::Automatic);
```
Hofer-Julian, line 66: Use the builder here

examples/expressions/main.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    window.set_child(Some(&scrolled_window));
```
Hofer-Julian, line 68: You can also use the builder for window and only create it in the end

examples/expressions/main.rs
```diff
@@ -0,0 +1,84 @@
...
+            })
+            .bind(&title_label, "label", None::<&gtk::Widget>);
+
+        metadata_expression
+            .chain_property(Metadata::static_type(), "last-modified")
+            .chain_closure(|args| {
+                let last_modified: glib::DateTime = args[1].get().unwrap();
+                format!("Last Modified: {}", last_modified.format_iso8601().unwrap())
+            })
+            .bind(&last_modified_label, "label", None::<&gtk::Widget>);
```
Hofer-Julian, line 56: Same here
```suggestion
            .bind(&last_modified_label, "label", gtk::NONE_WIDGET);
```

Hofer-Julian (changes requested):
Example looks good to me otherwise

examples/expressions/main.rs
SeaDve: I think that may be replaced now with gtk::Widget::NONE, so I think replace it with this const?

examples/expressions/main.rs
SeaDve: i  think building the entire window at the end could also work

examples/expressions/main.rs
Hofer-Julian: Good catch.

It should be `gtk::Widget::NONE`
I've updated the suggestion.

examples/expressions/metadata/imp.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+#[glib::object_subclass]
+impl ObjectSubclass for Metadata {
+    const NAME: &'static str = "Metadata";
+    type Type = super::Metadata;
+    type ParentType = glib::Object;
+
+    fn new() -> Self {
+        let now = glib::DateTime::new_now_local().unwrap();
```
bilelmoussaoui, line 20: very weird constructor name; cc @sdroege , maybe we should rename it to plain `now_local`?

examples/expressions/metadata/imp.rs
```diff
@@ -0,0 +1,81 @@
...
+pub struct Metadata {
+    pub title: RefCell<String>,
+    pub last_modified: RefCell<glib::DateTime>,
+}
+
+#[glib::object_subclass]
+impl ObjectSubclass for Metadata {
+    const NAME: &'static str = "Metadata";
+    type Type = super::Metadata;
+    type ParentType = glib::Object;
```
bilelmoussaoui, line 17: I'm somehow not sure if you need to specify the ParentType in this case

gtk4/src/expression.rs
```diff
@@ -146,6 +146,29 @@ impl Expression {
...
+    // rustdoc-stripper-ignore-next
+    /// Create a [`gtk::PropertyExpression`] that looks up for `property_name`
+    /// with self as parameter. This is useful in long chains of [`gtk::Expressions`].
+    pub fn chain_property(
+        &self,
+        this_type: glib::Type,
+        property_name: &str,
+    ) -> crate::PropertyExpression {
+        crate::PropertyExpression::new(this_type, Some(self), property_name)
+    }
```
bilelmoussaoui, line 159: Can 't we make it work by passing doing something like this `chain_property<T>(property_name)` ? I think you can use the `IsA<Object>` for that :)

gtk4/src/expression.rs
```diff
@@ -146,6 +146,29 @@ impl Expression {
             ))
         }
     }
+
+    // rustdoc-stripper-ignore-next
+    /// Create a [`gtk::PropertyExpression`] that looks up for `property_name`
+    /// with self as parameter. This is useful in long chains of [`gtk::Expressions`].
```
bilelmoussaoui, line 152:
```suggestion
    /// with self as parameter. This is useful in long chains of [`gtk::Expression`]s.
```

bilelmoussaoui: Same as elsewhere

sdroege:
> I found a possible problem with `chain_property` while working with the example. Sometimes it is needed to specify the value type. For example
> 
> ```rust
> list_item
>     .property_expression("item")
>     .chain_property("number")
> ```
> 
> this won't work because a gtk::ListItem's item will evaluate into a GObject, thus it doesn't have a number property unless it is downcasted into a type that has a property of number.
> 
> To solve this, the `this_type` param for `PropertyExpression` will be exposed:
> 
> ```rust
> list_item
>     .property_expression("item")
>     .chain_property(TypeThatHasNumberProp::static_type(), "number")
> ```

You could do it like in https://github.com/gtk-rs/gtk-rs-core/pull/315 . You could list all properties of the type of the object, and from there get the type that actually defines the property.

examples/expressions/metadata/imp.rs
sdroege: You always have to specify it

examples/expressions/metadata/imp.rs
sdroege: Sounds like a good idea, you do that? There are more similar constructors in datetime

examples/expressions/main.rs
sdroege: Why, they're equivalent

examples/expressions/metadata/imp.rs
bilelmoussaoui: Hmm, looks like a possible simplification of glib::object_subclass :)

examples/expressions/main.rs
Hofer-Julian: Sure, but why have them at all if we don't even use them in our examples.

examples/expressions/metadata/imp.rs
bilelmoussaoui: I did open an issue at least https://github.com/gtk-rs/gtk-rs-core/issues/342

SeaDve:
> You could do it like in [gtk-rs/gtk-rs-core#315](https://github.com/gtk-rs/gtk-rs-core/pull/315) . You could list all properties of the type of the object, and from there get the type that actually defines the property.

It should have a syntax of `chain_property::<glib::Object>(prop_name)`, like what was stated by bilelmoussaoui?

sdroege:
That would work, yes

bilelmoussaoui:
> > That would work, yes
> 
> I think it would be simpler to use T::static_type() where T IsA<glib::Object>

yes that's the same approach :-)

gtk4/src/expression.rs
bilelmoussaoui: oh indeed; @SeaDve do you think you can provide a PR for this? otherwise I will open an issue about it :)

examples/expressions/main.rs
SeaDve: thanks

SeaDve: thanks

sdroege:
> I meant instead of finding the property like [gtk-rs/gtk-rs-core#315](https://github.com/gtk-rs/gtk-rs-core/pull/315)

That seems less user-friendly though. The user would have to know on which of the parent types the property actually is.

SeaDve:
I read upon gtk-rs/gtk-rs-core#315, but not sure how would i translate it for this

gtk4/src/expression.rs
SeaDve: Sure

bilelmoussaoui:
> > I meant instead of finding the property like [gtk-rs/gtk-rs-core#315](https://github.com/gtk-rs/gtk-rs-core/pull/315)
> 
> That seems less user-friendly though. The user would have to know on which of the parent types the property actually is.


> I read upon [gtk-rs/gtk-rs-core#315](https://github.com/gtk-rs/gtk-rs-core/pull/315), but not sure how would i translate it for this
> 
> I gave it a try, and it's giving same errors as before, the `find_property` returns `None`:
> 
> ```rust
> pub fn chain_property(&self, property_name: &str) -> crate::PropertyExpression {
>         let pspec = glib::object::ObjectClass::from_type(self.value_type())
>             .unwrap()
>             .find_property(property_name)
>             .unwrap_or_else(|| panic!("Couldn't find a property named `{}` to lookup", property_name));
> 
>         crate::PropertyExpression::new(pspec.owner_type(), Some(self), property_name)
>     }
> ```

This can't really work because you're using the value_type, which is the type of the value the expression evaluates to and not the object having such property. In other words, I think you always have to specify the type T

bilelmoussaoui:
> > This can't really work because you're using the value_type, which is the type of the value the expression evaluates to and not the object having such property. In other words, I think you always have to specify the type T
> 
> But you already have the type: it's the type of `&self`.

Ah I think you got confused by the two chain methods; the one in the GObject trait already makes uses of the `T: IsA<Object>` for that; the other one inside `impl Expression` can't do that :-)

sdroege:
I think the way it is implemented now is all good, or am I missing something?

bilelmoussaoui:
> I think the way it is implemented now is all good, or am I missing something?

Yes it's all good now, i will have a quick look at the rest of the code & merge

examples/expressions/metadata/imp.rs
bilelmoussaoui: Opened an issue upstream at https://github.com/gtk-rs/gtk-rs-core/issues/350

bilelmoussaoui:
Thanks for your contribution @SeaDve :-)

SeaDve:
thanks for the review too!

PR #702, opened 2021-11-11
symbolic paintable

bilelmoussaoui:


gtk4/src/subclass/symbolic_paintable.rs
```diff
@@ -0,0 +1,97 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::subclass::prelude::*;
+use crate::SymbolicPaintable;
+use glib::translate::*;
+use glib::Cast;
+
+pub trait SymbolicPaintableImpl: ObjectImpl {
```
sdroege, line 8: Doesn't this required Paintable?

gtk4/src/subclass/symbolic_paintable.rs
```diff
@@ -0,0 +1,97 @@
...
+    let imp = instance.impl_();
+
+    let snapshot: Borrowed<gdk::Snapshot> = from_glib_borrow(snapshotptr);
+
+    imp.snapshot_symbolic(
+        from_glib_borrow::<_, SymbolicPaintable>(paintable).unsafe_cast_ref(),
+        &snapshot,
+        width,
+        height,
+        FromGlibContainer::from_glib_none_num(colors, n_colors),
```
sdroege, line 95: Should probably just pass this on as &[
gdk::RGBA]. Just go via slice::from_raw_parts()

gtk4/src/symbolic_paintable.rs
```diff
@@ -8,7 +8,7 @@ pub trait SymbolicPaintableExtManual: 'static {
     #[doc(alias = "gtk_symbolic_paintable_snapshot_symbolic")]
     fn snapshot_symbolic(
         &self,
-        snapshot: &impl IsA<gdk::Snapshot>,
+        snapshot: &gdk::Snapshot,
```
sdroege, line 11: Why?

bilelmoussaoui: We don't do that anywhere else; ideally we would just cast and make the api uses gtk::Snapshot here really

gtk4/src/subclass/symbolic_paintable.rs
bilelmoussaoui: It does indeed

gtk4/src/subclass/symbolic_paintable.rs
```diff
@@ -0,0 +1,98 @@
...
+use glib::Cast;
+
+pub trait SymbolicPaintableImpl: PaintableImpl {
+    fn snapshot_symbolic(
+        &self,
+        paintable: &Self::Type,
+        snapshot: &gdk::Snapshot,
+        width: f64,
+        height: f64,
+        colors: [gdk::RGBA],
```
sdroege, line 16:
```suggestion
        colors: &[gdk::RGBA],
```

sdroege (approved):

PR #703, opened 2021-11-11
Use impl IntoIterator<Item = Expression> in ClosureExpression params

SeaDve:
I'm not sure if I did it correctly


sdroege (approved):

SeaDve:
it seems to not accept slices (`&[Expressions]`) anymore, but only array and others

sdroege:
Maybe `impl IntoIterator<Item = impl AsRef<Expression>>` then?

SeaDve:
And clone it inside the function?

sdroege:
It's refcounted so why not

PR #704, opened 2021-11-12
Allow slices on ClosureExpression's params

SeaDve:
Follow up for https://github.com/gtk-rs/gtk4-rs/pull/703

### DIFF
--- a/gtk4/src/closure_expression.rs
+++ b/gtk4/src/closure_expression.rs
@@ -23,7 +23,7 @@ define_expression!(
 
 impl ClosureExpression {
     #[doc(alias = "gtk_closure_expression_new")]
-    pub fn new<F, R>(params: impl IntoIterator<Item = Expression>, callback: F) -> Self
+    pub fn new<F, R>(params: impl IntoIterator<Item = impl AsRef<Expression>>, callback: F) -> Self
     where
         F: Fn(&[Value]) -> R + 'static,
         R: ValueType,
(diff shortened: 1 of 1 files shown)


gtk4/src/closure_expression.rs
```diff
@@ -23,18 +23,22 @@ define_expression!(
 
 impl ClosureExpression {
     #[doc(alias = "gtk_closure_expression_new")]
-    pub fn new<F, R>(params: impl IntoIterator<Item = Expression>, callback: F) -> Self
+    pub fn new<F, R, I>(params: impl IntoIterator<Item = I>, callback: F) -> Self
     where
         F: Fn(&[Value]) -> R + 'static,
         R: ValueType,
+        I: AsRef<Expression> + Clone,
```
sdroege, line 30: Why not `impl IntoIterator<Item = impl AsRef<Expression>>`?

SeaDve: I think it's a bit more readable, but maybe just me

sdroege: Something for @bilelmoussaoui to decide then. Looks good to me apart from this :)

sdroege: Why does it need `Clone` though? `&Expression` is `Clone`

SeaDve: The compiler used to complain about it a while ago, but somehow it's now okay

bilelmoussaoui:
Needs a cargo fmt otherwise looks good now

gtk4/src/closure_expression.rs
bilelmoussaoui: > Why not `impl IntoIterator<Item = impl AsRef<Expression>>`

I'm also in favour of using something like this btw

sdroege (approved):

PR #706, opened 2021-11-13
Update per gtk-rs-core simplifications

bilelmoussaoui:
cc @Hofer-Julian as it modifies the book snippets


Hofer-Julian (approved):
Thanks! 

I don't think that any additional book changes are required :)

PR #708, opened 2021-11-13
book: Use the new expressions API

Hofer-Julian:
@SeaDve

### DIFF
--- a/book/listings/list_widgets/4/main.rs
+++ b/book/listings/list_widgets/4/main.rs
@@ -3,8 +3,8 @@ mod integer_object;
 use gtk::gio;
 use gtk::prelude::*;
 use gtk::{
-    Application, ApplicationWindow, ConstantExpression, Label, ListView, PolicyType,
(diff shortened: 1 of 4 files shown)


book/listings/list_widgets/4/main.rs
```diff
@@ -46,21 +46,11 @@ fn build_ui(app: &Application) {
...
-            "number",
-        );
-
-        // Bind "number" to "label"
-        number_expression.bind(&label, "label", Some(&label));
+        // Bind `list_item->item->number` to `label->label`
+        list_item
+            .property_expression("item")
+            .chain_property::<IntegerObject>("number")
+            .bind(&label, "label", Widget::NONE);
```
Hofer-Julian, line 53: No idea btw what [`this_`](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/struct.Expression.html#method.bind) is good for.
Doesn't seem to make any difference though.

PR #709, opened 2021-11-14
Remove unnecessary clone

SeaDve:
Doesn't need anymore since https://github.com/gtk-rs/gtk4-rs/commit/8218d1d4e8d5d6723c8bb602ff0a0d6497bfebdd


bilelmoussaoui:
Could have dropped the fixme as well no?

sdroege:
That was dropped, or do you mean one somewhere else?

bilelmoussaoui:
Oups, checking GitHub from a phone is always a bad idea

PR #710, opened 2021-11-14
Add method to create a "this" expression from GObject types

jf2048:
cc @SeaDve

Can be useful for some expressions, makes it so the expressions example could also be simplified as:

```rust
let metadata_expression = gtk::ListItem::this_expression("item")
             .chain_property::<Note>("metadata");
// ...
 metadata_expression
            .bind(&last_modified_label, "label", Some(list_item));
```

Not sure if it's worth creating an extra example for this?

### DIFF
--- a/examples/expressions/main.rs
+++ b/examples/expressions/main.rs
@@ -36,11 +36,9 @@ fn build_ui(app: &gtk::Application) {
 
         // Instead of binding properties and manually unbinding them, you can create expressions.
         // The value can be obtained even if it is several steps away.
-        let metadata_expression = list_item
+        list_item
             .property_expression("item")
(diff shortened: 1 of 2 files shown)


sdroege (approved):

SeaDve:
> Not sure if it's worth creating an extra example for this?

If it is possible, maybe add it on the `expressions` one

jf2048:
I changed one of the expressions in the example to use a `this` binding.

SeaDve:
I think the chain on the `title_binding` could be just one chain, so maybe remove the `metadata_expression` variable and chain it directly

jf2048:
This can wait until #714, there are some conflicts because both edit the same part of the example.

PR #711, opened 2021-11-14
book: Add chapter about CSS

Hofer-Julian:
This chapter talks about CSS.

More specifically:
- style classes
- pseudo-classes
- nodes
- images
- exported colors


book/listings/css/1/main.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+fn load_css() {
+    // Load the css file and add it to the provider
+    let provider = CssProvider::new();
+    provider.load_from_data(include_bytes!("style.css"));
+
+    // Add the provider to the default screen
+    StyleContext::add_provider_for_display(
+        &Display::default().expect("Error initializing GTK CSS provider."),
```
bilelmoussaoui, line 28: That's not a gorrect error message here btw, you fail to connect to a display.

bilelmoussaoui (commented):
It seems there's also an SVG file that was added, is that one needed?

book/listings/css/1/main.rs
Hofer-Julian: Thanks, fixed!

Hofer-Julian:
> It seems there's also an SVG file that was added, is that one needed?

Yes, it is the source.
What is indeed not needed are the accompanying png files, so I removed them.

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+In the case of `gtk::Button`, this node is called `button`.
+Therefore, we create a `style.css` file with the following content:
+
+<span class="filename">Filename: listings/css/1/style.css</span>
+
+```css
+{{#rustdoc_include ../listings/css/1/style.css}}
+```
+
+Next, we load the CSS file in the startup step of the application.
```
YaLTeR, line 20: Might be a good idea to add a small section about testing the CSS in the inspector. This way you can make sure it works and then when you're loading it into the app you know that if it doesn't work then it's the loading code and not the CSS.

YaLTeR, line 20:
```suggestion
Next, we need to load the CSS file in the startup step of the application.
```

book/listings/css/1/main.rs
```diff
@@ -0,0 +1,52 @@
...
+    app.connect_activate(build_ui);
+
+    // Run the application
+    app.run();
+}
+
+fn load_css() {
+    // Load the css file and add it to the provider
+    let provider = CssProvider::new();
+    provider.load_from_data(include_bytes!("style.css"));
```
YaLTeR, line 24: For me `include_bytes!()` didn't recompile when changing the file. If that's the case, it might result in a bit of a debugging nightmare.

book/listings/css/1/main.rs
```diff
@@ -0,0 +1,52 @@
...
+    // Connect to signals
+    app.connect_startup(|_| load_css());
+    app.connect_activate(build_ui);
+
+    // Run the application
+    app.run();
+}
+
+fn load_css() {
+    // Load the css file and add it to the provider
```
YaLTeR, line 22:
```suggestion
    // Load the CSS file and add it to the provider
```

YaLTeR: and in other files too

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+As usual, the widgets are created during the activate step.
+
+<span class="filename">Filename: listings/css/1/main.rs</span>
+
+```rust ,no_run,noplayground
+{{#rustdoc_include ../listings/css/1/main.rs:main}}
+```
+
+When we now run the app, we notice that our button *and* the "close" button are red.
+Probably not what we wanted, but that is what we typed.
```
YaLTeR, line 30:
```suggestion
Probably not what we wanted, but that is what our CSS snippet does.
```

book/src/css.md
YaLTeR: So basically move the inspector blue message from below here and reword it a bit.

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+We did not specify for which button the rule should apply, so it was applied to both of them.
+
+<div style="text-align:center"><img src="img/css_1.png"/></div>
+
+>The `GtkInspector` comes in quite handy (not only) when playing with CSS.
+>Make sure that the window of your app is focused and press <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>D</kbd>.
+> A window will pop up which lets you browse and even manipulate the state of your app.
+> Open the `CSS` view and override the button color with the following snippet.
```
YaLTeR, line 38:
```suggestion
> Open the CSS view and override the button color with the following snippet.
```

YaLTeR: and also don't forget to press the unpause button

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+>```css
+>button {
+>  color: blue;
+>}
+>```
+
+## Classes
+
+[Class selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Class_selectors) are one way to limit the amount of elements a CSS rule applies to.
```
YaLTeR, line 47:
```suggestion
[Class selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Class_selectors) are one way to choose which specific elements a CSS rule applies to.
```

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+Let us look at few different scenarios where they are involved.
```
YaLTeR, line 48:
```suggestion
Let us look at few different scenarios a where CSS classes are involved.
```

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+
+
+### Classes Applied by GTK
+
+GTK adds style classes to many of its nodes, often depending on a certain condition.
+In the case of the node [`button`](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/struct.Button.html#css-nodes), the style class `.text-button` will be added when the content of the corresponding widget is just a label.
```
YaLTeR, line 54:
```suggestion
GTK adds style classes to many of its widgets, often depending on their content.
A [`gtk::Button`](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/struct.Button.html#css-nodes), for example, will get the `.text-button` style class when its content is a label.
```

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+
+<span class="filename">Filename: listings/css/2/style.css</span>
+
+```css
+{{#rustdoc_include ../listings/css/2/style.css}}
+```
+
+<div style="text-align:center"><img src="img/css_2.png"/></div>
+
+### Classes Applied Manually
```
YaLTeR, line 66:
```suggestion
### Applying Your Own Class
```

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+
+If we want to apply different rules to nodes which got the same style classes added, we will have to think of a different strategy.
+For that we add our own CSS class with [`add_css_class`](../docs/gtk4/prelude/trait.WidgetExt.html#tymethod.add_css_class).
```
YaLTeR, line 69:
```suggestion
If you need to style differently two widgets that have the same automatic GTK style classes, you can add your own CSS class using [`add_css_class()`](../docs/gtk4/prelude/trait.WidgetExt.html#tymethod.add_css_class).
```

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+
+<span class="filename">Filename: listings/css/3/main.rs</span>
+
+```rust ,no_run,noplayground
+{{#rustdoc_include ../listings/css/3/main.rs:buttons}}
+```
+
+And we create a CSS rule that applies to `button` nodes with the style class `button_1`.
```
YaLTeR, line 77:
```suggestion
Then, create a CSS rule that applies to `button` nodes with the style class `button_1`.
```

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+
+<span class="filename">Filename: listings/css/3/style.css</span>
+
+```css
+{{#rustdoc_include ../listings/css/3/style.css}}
+```
+
+This way, only the first button gets colored red.
```
YaLTeR, line 85:
```suggestion
This way, only the first button (for which we added the `button_1` style class) gets colored red.
```

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+If you want to make sure that CSS rules only apply to a hand-selected set of nodes, adding your own style classes is your best bet.
+Just make sure that you do not choose a name for your style class that GTK either adds itself or provides CSS rules for.
+This brings us to the next point.
+
+<div style="text-align:center"><img src="img/css_3.png"/></div>
+
+### CSS Rules Provided by GTK
+
+Certain styles are common enough that GTK provides its own CSS rules for.
```
YaLTeR, line 94:
```suggestion
Certain styles are common enough that GTK provides preset CSS rules for them in form of style classes.
```

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+With pseudo-classes you can specify a specific state of the CSS node.
+Let us use a single button with the style class `button_1` added to demonstrate this concept.
+
+<span class="filename">Filename: listings/css/6/window/window.ui</span>
+
+```xml
+{{#rustdoc_include ../listings/css/6/window/window.ui}}
+```
+
+By adding the pseudo-class `hover`, we say that we only want this rule to apply for `button` nodes with style class `button_1` over which we currently hover.
```
YaLTeR, line 134:
```suggestion
By adding the pseudo-class `hover`, we say that we only want this rule to apply for `button` nodes with style class `button_1` over which we currently hover the mouse pointer.
```

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+
+```xml
+{{#rustdoc_include ../listings/todo_app/3/todo_row/todo_row.ui}}
+```
+
+One way to make the rows set apart a bit is to add borders.
+But which color to use for the borders?
+Our To-Do app should represent a real world app, so we will not be satisfied with choosing yellow and calling it a day.
+Your stylesheet has colors defined for these use cases, so let us have a look.
+As of this writing, the exported colors of the default stylesheet can be found in its [source code](https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/theme/Default/_colors-public.scss).
```
YaLTeR, line 242:
```suggestion
As of this writing, the exported colors of the default stylesheet can be found in its [source code](https://gitlab.gnome.org/GNOME/gtk/-/blob/b2c227e9c57839a2a4e24462a71ae0bad9a95264/gtk/theme/Default/_colors-public.scss).
```

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+
+There we find the color `borders`, which should be used for the widget's main borders color.
```
YaLTeR, line 244:
```suggestion
There we find the color `borders`, which should be used for the widgets' border color.
```

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+Perfect!
+We can access exported colors by adding an `@` in front of their names.
```
YaLTeR, line 246:
```suggestion
We can access pre-defined colors by adding an `@` in front of their names.
```

book/listings/todo_app/3/style.css
```diff
@@ -1,3 +1,3 @@
-.todo_row {
+box.todo_row {
```
YaLTeR, line 1: I'd use just `.todo_row` for this kind of thing (one less place to change if you decide to change the base widget from gtk::Box). But the tutorial also doesn't have any examples of free-standing style classes.

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+
+Now our tasks have borders around them, and we are one step further in finishing our To-Do app.
+
+<div style="text-align:center"><img src="img/todo_app_change.png"/></div>
+
+## Conclusion
+
+We now learned how to use CSS to style our apps.
+We can make our CSS rules as specific as we want them to be.
+Even targeting nodes deep within the inheritance tree is not a problem anymore.
```
YaLTeR, line 262:
```suggestion
Even targeting nodes deep within the inheritance tree is not a problem.
```

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+
+In the following chapter, we will learn how to include CSS files, icons and basically everything that our app needs during runtime in a binary bundle.
```
YaLTeR, line 264:
```suggestion
In the following chapter we will learn how to make a binary bundle with CSS files, icons and any other resources that our app might need at runtime.
```

book/src/css.md
```diff
@@ -0,0 +1,264 @@
+# CSS
+
+When you want to modify the style of your website, you use [CSS](https://de.wikipedia.org/wiki/Cascading_Style_Sheets).
+Since so many people are already familiar with web development, it makes sense that GTK supports its own variant of CSS.
```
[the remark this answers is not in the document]
Hofer-Julian: Fair, will change that.

book/listings/css/1/main.rs
Hofer-Julian: I do remember you saying that, but it does for me and I can't find anything regarding this problem on the web.

book/src/css.md
Hofer-Julian: I guess the "a" was not intended?

YaLTeR: Yeah

book/listings/css/2/main.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+fn build_ui(app: &Application) {
+    // Create button
+    let button = Button::builder()
+        .label("Press me!")
+        .margin_top(12)
+        .margin_bottom(12)
+        .margin_start(12)
+        .margin_end(12)
```
Exalm, line 40: If you have css, why not use it for margin as well?

book/src/css.md
YaLTeR: From the programmer's perspective the style classes are added to widgets, and the nodes are css names that are set for widgets. I think the way I rephrased this is clearer

book/listings/css/2/style.css
```diff
@@ -0,0 +1,3 @@
+button.text-button {
```
Exalm, line 1: `.text-button` is a gtk style class, you really don't want to style anything like this. Add your own name or style class instead. It's too generic - and you're probably picked `.text-button` because otherwise it affects the window close button, right? :) But it's still too broad.

book/src/css.md
YaLTeR: Isn't style class used in this section to mean css class?

book/listings/css/6/style.css
```diff
@@ -0,0 +1,6 @@
+button.button_1:hover {
+  color: red;
+  background-color: yellow;
+  background-image: none;
+  transition: 1s;
```
Exalm, line 5: FWIW it's a bad idea to add transition specifically on hover. This means you don't get the same transition when transitioning _from_ hover to another state. If you change transition, you want to do that in general, not just for hover.

Also, this transition affects things such as focus-ring, you probably want to style specifically background/color instead.

book/listings/css/6/style.css
```diff
@@ -0,0 +1,6 @@
+button.button_1:hover {
+  color: red;
+  background-color: yellow;
+  background-image: none;
```
Exalm, line 4: These 2 lines are same as `background: yellow;`

book/listings/css/8/style.css
```diff
@@ -0,0 +1,4 @@
+combobox arrow {
+  color: red;
+  -gtk-icon-source: -gtk-icontheme('format-justify-fill-symbolic');
```
Exalm, line 3: Wouldn't recommend to style things like this if possible. It's not something apps should do.

book/listings/todo_app/3/style.css
Exalm: Yes, the `box` doesn't add anything here.

book/listings/todo_app/3/style.css
```diff
@@ -1,3 +1,3 @@
...
   border: 1px solid @borders;
```
Exalm, line 2: FWIW there's a standard style class for this - `.frame`.

book/src/css.md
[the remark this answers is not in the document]
Exalm: The justification was sorta that but not completely - the intention was to make it possible for designers to change style themselves. That hasn't really worked out though.

book/src/css.md
```diff
@@ -0,0 +1,266 @@
...
+
+> We will not explain every piece of CSS syntax used in this chapter.
+> If you are new to it or just need a refresher, have a look at the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Syntax).
+
+Let us say we have a button and we want to set its font color to red.
+Every type of widget has a corresponding CSS node.
```
Exalm, line 10: A widget can have multiple nodes as well. The combo box arrow from a snippet above is an example, though I don't remember if it's a widget or just a node. But e.g. scrolled window undershoot/overshoot are standalone css nodes.

book/src/css.md
```diff
@@ -0,0 +1,266 @@
...
+Next, we need to load the CSS file in the startup step of the application.
+As usual, the widgets are created during the activate step.
+
+<span class="filename">Filename: listings/css/1/main.rs</span>
+
+```rust ,no_run,noplayground
+{{#rustdoc_include ../listings/css/1/main.rs:main}}
+```
+
+When we now run the app, we notice that our button *and* the "close" button are red.
```
Exalm, line 29: Ok, nevermind my comment above I guess.

book/src/css.md
Exalm: Yes.

book/src/css.md
Exalm: What is an automatic GTK style class in this context?

book/src/css.md
Exalm: Adwaita is a stylesheet. `.suggested-action` and `.text-button` are style classes.

book/src/css.md
```diff
@@ -0,0 +1,266 @@
...
+<div style="text-align:center">
+ <video autoplay muted loop>
+  <source src="vid/css_6.webm" type="video/webm">
+Your browser does not support the video tag.
+ </video>
+</div>
+
+## Nodes
+
+In the previous examples a widget always corresponded to a single CSS node.
```
Exalm, line 155: Ok, nevermind another comment I guess. :)

Exalm: That said, not sure if mentioning GtkComboBox, well, anywhere in the book is a good idea - hopefully that widget is deprecated sooner rather than later.

book/src/css.md
YaLTeR: Title of the section right before about stuff like `.text-content`

book/src/css.md
```diff
@@ -0,0 +1,266 @@
...
+Fear not, the [Icon Library](https://apps.gnome.org/app/org.gnome.design.IconLibrary/) has you covered there.
+Scroll through the provided icons and choose one from the pre-installed system icons section.
+It's possible to embed icons from other sections too, we'll learn about that in a follow-up chapter.
+
+For now, we will take the icon with the semantic name "format-justify-fill-symbolic".
+In the screenshot below you see that this corresponds to three parallel lines.
+However, on a different system this icon might look different or not exist at all.
+This is why it is typically better to embed the icons you use.
+
+GTK provides its own parameter for loading icons in buttons and expanders: `-gtk-icon-source`.
```
Exalm, line 211: I don't think it works in buttons, does it? It works with a few very specific widgets:

- `GtkBuiltinIcon` (GtkCheckButton check etc, and likely this arrow as well)
- `GtkSpinner`
- `GtkTextHandle` (touch text selection)
- `GtkTreeView` somwhewhere - likely the arrow in headers
- `GtkPaned` handle

Maybe a few more, but - it's private and e.g. `GtkImage` doesn't support it. Chances are it will get deprecated before GTK 5.

book/src/css.md
Exalm: `.text-button` you mean? Sure but there's `add your own CSS class using add_css_class()` in the same sentence. :)

book/src/css.md
Hofer-Julian: Yes, I was confused

book/listings/css/2/style.css
Hofer-Julian: Now that you’ve read the text, do you agree that I don‘t really recommend it, but mostly try to make a point that GTK adds style classes?

Exalm: Yep.

book/src/css.md
```diff
@@ -0,0 +1,266 @@
...
+```css
+{{#rustdoc_include ../listings/todo_app/3/style.css}}
+```
+
+Now our tasks have borders around them, and we are one step further in finishing our To-Do app.
+
+<div style="text-align:center"><img src="img/todo_app_change.png"/></div>
+
+## Conclusion
+
```
Hofer-Julian, line 261: Mention here a gain the complete CSS syntax `css-name#name.class-1.class-2`.
Otherwise, I expect that people will be confused about the difference between `css-name` and `name`.

book/src/css.md
```diff
@@ -0,0 +1,264 @@
...
+Indeed, the resulting app features a `ComboBoxText` with a red arrow.
+
+<div style="text-align:center"><img src="img/css_7.png"/></div>
+
+## Images
+
+Do we really want to settle with an arrow?
+Maybe there are much nicer icons out there you do not even know about yet.
+Fear not, the [Icon Library](https://apps.gnome.org/app/org.gnome.design.IconLibrary/) has you covered there.
```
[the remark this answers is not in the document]
Hofer-Julian: 😢

book/src/css.md
Hofer-Julian: I will replace it with GtkMenuButton and then manipulate one if its child nodes

book/listings/todo_app/3/style.css
Hofer-Julian: Good to know.

I will first use css_name as suggested by @bilelmoussaoui to show this concept.
Then I will show that the same can be achieved by adding the `frame` style class.

book/src/css.md
```diff
@@ -0,0 +1,266 @@
...
+We can also add style classes with the interface builder.
+Just add the `<style>` element to your widget.
+The `<style>` element is documented [here](../docs/gtk4/struct.Widget.html#gtkwidget-as-gtkbuildable). 
+
+<span class="filename">Filename: listings/css/5/window/window.ui</span>
+
+```xml
+{{#rustdoc_include ../listings/css/5/window/window.ui}}
+```
+
```
Hofer-Julian, line 119: Add here a section about `#name`

book/src/css.md
```diff
@@ -0,0 +1,266 @@
+# CSS
```
Hofer-Julian, line 1: Redo all screenshots

PR #714, opened 2021-11-14
Allow chaining expressions with Closure objects

jf2048:
I had to change `ClosureExpression::with_closure` to expand out the generic parameters, it could be changed back depending on this: https://github.com/rust-lang/rust/issues/83701

Also includes some docs fixes.

### DIFF
--- a/gtk4/src/closure_expression.rs
+++ b/gtk4/src/closure_expression.rs
@@ -50,12 +50,11 @@ impl ClosureExpression {
     }
 
     #[doc(alias = "gtk_closure_expression_new")]
-    pub fn with_closure<R>(
-        params: impl IntoIterator<Item = impl AsRef<Expression>>,
-        closure: glib::Closure,
-    ) -> Self
+    pub fn with_closure<R, I, E>(params: I, closure: glib::Closure) -> Self
     where
         R: ValueType,
+        I: IntoIterator<Item = E>,
+        E: AsRef<Expression>,
     {
         assert_initialized_main_thread!();
 
--- a/gtk4/src/closure_expression.rs
+++ b/gtk4/src/closure_expression.rs
@@ -23,16 +23,13 @@ define_expression!(
 
 impl ClosureExpression {
     #[doc(alias = "gtk_closure_expression_new")]
-    pub fn new<F, R>(params: impl IntoIterator<Item = impl AsRef<Expression>>, callback: F) -> Self
+    pub fn new<R, I, E>(params: I, closure: glib::RustClosure) -> Self
     where
-        F: Fn(&[Value]) -> R + 'static,
         R: ValueType,
+        I: IntoIterator<Item = E>,
+        E: AsRef<Expression>,
     {
         assert_initialized_main_thread!();
-        let closure = glib::Closure::new_local(move |values| {
-            let ret = callback(values);
-            Some(ret.to_value())
-        });
 
         let params = params
             .into_iter()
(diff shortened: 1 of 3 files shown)


SeaDve:
I think `chain_gclosure` may be clearer

jf2048:
That may not be straightforward as the macro was changed to just `closure!`. Maybe it could be `chain_with_closure`, but that also seems confusing.

bilelmoussaoui:
Maybe rename the first `chain_closure` to `chain_callback` & keep this one as `chain_closure`?

SeaDve:
I just realized that it was named `chain_closure` before since it chains a `ClosureExpression`, but something like `chain_closure_with_closure` is a weird name

bilelmoussaoui:
> I just realized that it was named `chain_closure` before since it chains a `ClosureExpression`, but something like `chain_closure_with_closure` is a weird name

Well the ClosureExpression should only take a glib::Closure but we provide a helper constructor that takes a callback. So it's fine to not name it  `chain_closure` if it's documented properly. The double `chain_closure_with_closure` is definitely not great

SeaDve:
I see, maybe `chain_closure` and `chain_closure_with_callback` ?

bilelmoussaoui:
> I see, maybe `chain_closure` and `chain_closure_with_callback` ?

No strong opinions here, @sdroege ?:)

sdroege:
Seems fine either way. Ask your favorite random number generator and go with that :)

bilelmoussaoui:
> I see, maybe `chain_closure` and `chain_closure_with_callback` ?

Let's go with this one then

SeaDve:
semi related to pr, maybe the ClosureExpression could also ne renamed to `new` and `with_callback` for coherence

gdk4/src/event.rs
```diff
@@ -281,7 +281,7 @@ impl AsRef<Event> for Event {
 }
 
 // rustdoc-stripper-ignore-next
-/// A common trait implemented by the various [`gdk::Event`] types.
+/// A common trait implemented by the various [`gdk::Event`](crate::Event) types.
```
bilelmoussaoui, line 284: I fixed those as part https://github.com/gtk-rs/gtk4-rs/pull/719/ btw :-)

jf2048:
This is updated with the name changes, but it's a bit awkward with the recent addition of `RustClosure` because now it always has to be `closure!( ... ).into()`.

gtk4/src/closure_expression.rs
```diff
@@ -23,16 +23,13 @@ define_expression!(
 
 impl ClosureExpression {
     #[doc(alias = "gtk_closure_expression_new")]
-    pub fn new<F, R>(params: impl IntoIterator<Item = impl AsRef<Expression>>, callback: F) -> Self
+    pub fn new<R, I, E>(params: I, closure: glib::Closure) -> Self
```
bilelmoussaoui, line 26: Make it take a RustClosure ?

bilelmoussaoui:
other than the cargo fmt, lgtm

gtk4/src/closure_expression.rs
sdroege: It probably has to for type-safety reasons anyway.

jf2048: I changed it to `RustClosure`, if someone wants to add another function that takes just `Closure` it should probably be unsafe anyway.

bilelmoussaoui:
Thanks!

PR #715, opened 2021-11-15
drop instance_init per IsImplementable changes

bilelmoussaoui:


sdroege:
@GuillaumeGomez The same might also be needed for gtk3-rs.

bilelmoussaoui:
> @GuillaumeGomez The same might also be needed for gtk3-rs.

I checked already and it's not needed

PR #716, opened 2021-11-15
Add setters to RGBA & missing builder methods

### LINKED ISSUE #713: gdk rgba no longer implement DerfMut
We can no longer modify a gdk::RGBA's values & there are no setters on it neither. One or both should be provided

bilelmoussaoui:
Fixes #713

### DIFF
--- a/gdk4/src/rgba.rs
+++ b/gdk4/src/rgba.rs
@@ -45,16 +45,16 @@ impl RGBABuilder {
     pub fn build(self) -> RGBA {
         let mut rgba = RGBA::WHITE;
         if let Some(blue) = self.blue {
-            rgba.0.blue = blue;
+            rgba.set_blue(blue);
         }
         if let Some(red) = self.red {
-            rgba.0.red = red;
+            rgba.set_red(red);
         }
         if let Some(green) = self.green {
-            rgba.0.green = green;
+            rgba.set_green(green);
         }
         if let Some(alpha) = self.alpha {
-            rgba.0.alpha = alpha;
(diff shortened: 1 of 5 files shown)


gdk4/src/rgba.rs
```diff
@@ -71,22 +71,46 @@ impl RGBA {
...
+    /// This method returns an instance of [`RGBABuilder`] which can be used to create [`RGBA`] objects.
+    pub fn builder() -> RGBABuilder {
+        RGBABuilder::default()
+    }
+
     pub fn red(&self) -> f32 {
         self.0.red
     }
 
+    pub fn set_red(&self, red: f32) {
```
sdroege, line 86:
```suggestion
    pub fn set_red(&mut self, red: f32) {
```

Also for the others :scream:

gdk4/src/rgba.rs
```diff
@@ -154,3 +178,16 @@ impl FromStr for RGBA {
         RGBA::parse(s)
     }
 }
+
+impl std::ops::Deref for RGBA {
```
sdroege, line 182: Why?

bilelmoussaoui: I think setters only are fine, but I had code where I get the RGBA from the gtk::StyleContext & just change the alpha value of the color; a setter should be fine I guess :)

sdroege (approved):
You copy&paste that to gtk3-rs?

bilelmoussaoui:
> You copy&paste that to gtk3-rs?

https://github.com/gtk-rs/gtk3-rs/pull/673

PR #717, opened 2021-11-15
Update per builder pattern changes

bilelmoussaoui:


bilelmoussaoui:
cc @Hofer-Julian it might needs changes to the book, at least i found a link to ButtonBuilder =-)

Hofer-Julian:
Good catch!

Do you think you could fix the link within this PR?

bilelmoussaoui:
I will in a different PR

PR #733, opened 2021-11-21
Regen with gir#1275

bilelmoussaoui:
See https://github.com/gtk-rs/gir/pull/1275


Hofer-Julian:
@bilelmoussaoui please remember to ping for book changes :)

@sdroege, what does the new generic parameter of `emit_by_name` mean?
I should probably mention that in the book.

bilelmoussaoui:
> @bilelmoussaoui please remember to ping for book changes :)

The change is minimal that's why I thought it's not a problem

> @sdroege, what does the new generic parameter of `emit_by_name` mean? I should probably mention that in the book.

before this change, the emit_by_name returned a Option<Value> and now it returns either T if the signal handler returned a value of type T or `()` in case you expect None to be returned by the signal handler.

Hofer-Julian:
> now it returns either T if the signal handler returned a value of type T or () in case you expect None to be returned by the signal handler.

Thanks, I will open a PR to mention this in the corresponding book section.

sdroege:
> @sdroege, what does the new generic parameter of `emit_by_name` mean?

It's the expected return value (and it will panic otherwise). If the return type is not `()`, type inference will also usually figure it out.

sdroege:
FWIW I don't like that you have to provide `()` if it doesn't have a return value, but that seems to be required for now. And at least it makes sure you don't accidentally forget to do something with the return value :)

PR #735, opened 2021-11-24
examples: Fix up virtual methods example

sdroege:
Casting a `&glib::Object` to an instance struct only works if all
private types are zero-sized. Instead of that we can directly get the
subclass implementation struct with safe code from the object reference.

Also various functions were marked as `unsafe` although not actually needed.

----

CC @Cogitri

### DIFF
--- a/examples/virtual_methods/base_button/imp.rs
+++ b/examples/virtual_methods/base_button/imp.rs
@@ -15,9 +15,9 @@ pub type PinnedFuture = Pin<Box<dyn Future<Output = Result<(), Error>> + 'static
 pub struct BaseButtonClass {
     pub parent_class: gtk::ffi::GtkButtonClass,
     // If these functions are meant to be called from C, you need to make these functions
-    // `extern "C"` & use FFI-safe types (usually raw pointers).
-    pub sync_method: Option<unsafe fn(&BaseButtonInstance, extra_text: Option<String>)>,
-    pub async_method: Option<unsafe fn(&BaseButtonInstance) -> PinnedFuture>,
+    // `unsafe extern "C" fn` & use FFI-safe types (usually raw pointers).
+    pub sync_method: fn(&BaseButtonInstance, extra_text: Option<String>),
+    pub async_method: fn(&BaseButtonInstance) -> PinnedFuture,
 }
 
 unsafe impl ClassStruct for BaseButtonClass {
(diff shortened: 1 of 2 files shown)


examples/virtual_methods/base_button/mod.rs
```diff
@@ -99,20 +99,18 @@ unsafe impl<T: BaseButtonImpl> IsSubclassable<T> for BaseButton {
 }
 
 // Virtual method implementation trampolines
-unsafe fn sync_method_trampoline<T>(this: &BaseButton, extra_text: Option<String>)
+fn sync_method_trampoline<T>(this: &BaseButton, extra_text: Option<String>)
 where
     T: ObjectSubclass + BaseButtonImpl,
 {
-    let instance = &*(this as *const _ as *const T::Instance);
-    let imp = instance.impl_();
```
sdroege, line 107: This here was the problem

CC @vhdirk

examples/virtual_methods/base_button/imp.rs
```diff
@@ -85,6 +82,7 @@ impl ObjectSubclass for BaseButton {
 
 impl ObjectImpl for BaseButton {
     fn constructed(&self, obj: &Self::Type) {
+        self.parent_constructed(obj);
```
sdroege, line 85: And this here is added for good measure but it is not really necessary

examples/virtual_methods/base_button/mod.rs
```diff
@@ -65,24 +65,16 @@ impl<T: BaseButtonImpl> BaseButtonImplExt for T {
     fn parent_sync_method(&self, obj: &BaseButton, extra_text: Option<String>) {
         unsafe {
             let data = Self::type_data();
-            let parent_class = data.as_ref().parent_class() as *mut imp::BaseButtonClass;
-            if let Some(ref f) = (*parent_class).sync_method {
-                f(obj, extra_text)
-            } else {
-                unimplemented!()
-            }
+            let parent_class = &*(data.as_ref().parent_class() as *mut imp::BaseButtonClass);
```
sdroege, line 68: I think by making the `TypeData` generic over `T: ObjectSubclass`, we can have `parent_class()` return a properly typed reference and get rid of the `unsafe` code around here.

@vhdirk @Cogitri Does one of you want to give this a try?

vhdirk: You mean something like:

```rust
pub struct TypeData<T> where T: ObjectSubclass {
    #[doc(hidden)]
    pub type_: Type,
    #[doc(hidden)]
    pub parent_class: *mut T::Class,
    #[doc(hidden)]
    pub parent_ifaces: Option<HashMap<Type, ffi::gpointer>>,
    #[doc(hidden)]
    pub class_data: Option<HashMap<Type, Box<dyn Any + Send + Sync>>>,
    #[doc(hidden)]
    pub private_offset: isize,
    #[doc(hidden)]
    pub private_imp_offset: isize,
}

```

sdroege: Yes, and `parent_class() -> &T::Class` with a `NULL` check first

vhdirk: As said in chat: these generic will propagate upwards to `ObjectSubclass`, where you'd need make the `ObjectSubclassType` constraint generic over an associated type of `ObjectSubclass` itself. I don't think that's possible.

PR #740, opened 2021-12-03
Add example of a Rotation Bin widget

alatiera:
Add example of a Rotation Bin widget
    
This widget can rotate its child at 90 degrees steps.
    
Adaptation of other demo code by Alexander Mikhaylenko [1] and
Benjamin Otte [2]
    
[1] https://gitlab.gnome.org/exalm/rotation-demo
[2] https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4186


alatiera:
WIP atm cause there is a bug in the measure/request mode impl.

```
(rotation_bin:42983): Gtk-WARNING **: 09:11:17.456: RotationBin 0x55f8310d71a0 reported a horizontal baseline
**
Gtk:ERROR:../gtk/gtksizerequestcache.c:85:_gtk_size_request_cache_commit: assertion failed: (minimum_baseline == -1)
Bail out! Gtk:ERROR:../gtk/gtksizerequestcache.c:85:_gtk_size_request_cache_commit: assertion failed: (minimum_baseline == -1)
fish: Job 1, 'cargo run --release --bin rotat…' terminated by signal SIGABRT (Abort)
```

examples/rotation_bin/main.rs
```diff
@@ -0,0 +1,71 @@
...
+        .valign(gtk::Align::Center)
+        .row_spacing(6)
+        .column_spacing(6)
+        .build();
+
+    let clockwise = gtk::Button::with_label("Rotate clockwise");
+    let counter_clockwise = gtk::Button::with_label("Rotate counter clockwise");
+
+    clockwise.connect_clicked(clone!(@weak rotate_bin => move |_| rotate_bin.rotate_clockwise()));
+    counter_clockwise.connect_clicked(clone!(@weak rotate_bin => move |_| rotate_bin.rotate_counter_clockwise()));
```
alatiera, line 45: any preference on how to format these? Currently they are a bit long ~105 chars.

examples/Cargo.toml
```diff
@@ -16,6 +16,10 @@ libloading = { version = "0.7.0", optional = true }
 path = "../gtk4"
 package = "gtk4"
 
+[dependencies.gsk]
```
bilelmoussaoui, line 19: is this needed? you can use gsk from gtk directly

examples/rotation_bin/main.rs
bilelmoussaoui:
```suggestion
    counter_clockwise.connect_clicked(clone!(@weak rotate_bin => move |_| {
        rotate_bin.rotate_counter_clockwise();
    }));
```
would do something like this

examples/Cargo.toml
alatiera: Indeed, fixed

examples/rotation_bin/main.rs
alatiera: changed

sdroege:
> WIP atm cause there is a bug in the measure/request mode impl.

Do you know why and how to fix that?

examples/rotation_bin/README.me
```diff
@@ -0,0 +1,5 @@
+# RotationBin
+
+This example implements a Bin widget that can rotate its children by 90 degrees.
+
+![Screenshot](screenshot.png)
```
sdroege, line 5: You probably want to use a screenshot that is in an actually rotated state :)

examples/rotation_bin/main.rs
```diff
@@ -0,0 +1,83 @@
...
+fn build_ui(application: &gtk::Application) {
+    let window = gtk::ApplicationWindow::new(application);
+
+    let vbox = gtk::Box::new(gtk::Orientation::Vertical, 6);
+    let rotate_bin = RotationBin::new();
+    let img = gtk::Image::builder()
+        .pixel_size(128)
+        .icon_name("audio-x-generic")
+        .build();
+    rotate_bin.set_child(Some(img.upcast_ref()));
```
sdroege, line 28: Why is the explicit cast here necessary?

examples/rotation_bin/rotation_bin/imp.rs
```diff
@@ -0,0 +1,203 @@
...
+        };
+    }
+
+    fn measure(
+        &self,
+        _widget: &Self::Type,
+        orientation: gtk::Orientation,
+        for_size: i32,
+    ) -> (i32, i32, i32, i32) {
+        if self.child.borrow().is_none() {
```
sdroege, line 121: In all these functions, why don't you do

```rust
let child = match &*self.child.borrow() {
    None => return ...,
    Some(child) => child,
};
```

And then don't do the `unwrap` later?

examples/rotation_bin/rotation_bin/imp.rs
```diff
@@ -0,0 +1,203 @@
...
+            Rotation::Deg90 | Rotation::Deg270 => (height, width),
+            _ => (width, height),
+        };
+
+        let transform = gsk::Transform::new();
+        let transform = transform
+            .translate(&Point::new(width as f32 / 2.0, height as f32 / 2.0))
+            .unwrap();
+        let transform = transform.rotate(self.rotation.get().into()).unwrap();
+        let transform = transform.translate(&Point::new(-w as f32 / 2.0, -h as f32 / 2.0));
```
sdroege, line 158: Especially this virtual method could use some comments, but also the others to explain why and what they do

examples/rotation_bin/rotation_bin/mod.rs
```diff
@@ -0,0 +1,100 @@
+mod imp;
+
+use gtk::glib;
+use gtk::subclass::prelude::*;
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq, glib::Enum)]
+#[repr(i32)]
+#[enum_type(name = "RotationEnum")]
```
sdroege, line 8: Why `RotationEnum` and not just `Rotation`?

examples/rotation_bin/rotation_bin/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+            Self::Normal => Self::Deg270,
+            Self::Deg90 => Self::Normal,
+            Self::Deg180 => Self::Deg90,
+            Self::Deg270 => Self::Deg180,
+        }
+    }
+}
+
+impl From<i32> for Rotation {
+    fn from(item: i32) -> Self {
```
sdroege, line 37: Do you actually need this somewhere? If you do, as it's part of the public API of the widget this should really be a `TryFrom`.

examples/rotation_bin/rotation_bin/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+        Self::new()
+    }
+}
+
+impl RotationBin {
+    pub fn new() -> Self {
+        glib::Object::new(&[]).expect("Failed to create a RotationBin")
+    }
+
+    pub fn set_child(&self, widget: Option<&gtk::Widget>) {
```
sdroege, line 81:
```suggestion
    pub fn set_child(&self, widget: Option<&impl IsA<gtk::Widget>>) {
```

examples/rotation_bin/rotation_bin/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+        let priv_ = imp::RotationBin::from_instance(self);
+        priv_.set_child(self, widget);
+    }
+
+    pub fn set_rotation(&self, rotation: Rotation) {
```
sdroege, line 86: Also `rotation()` and `child()` helper functions for the property getters maybe?

examples/rotation_bin/rotation_bin/mod.rs
alatiera: I assume that's the GType name so having it be just Rotation seems a bit weird. Would be fine if the types where prefixed I guess.

examples/rotation_bin/rotation_bin/imp.rs
alatiera: Hmm, doesn't look as pretty. The borrow makes the compiler complain about dropping it.

alatiera:
> > WIP atm cause there is a bug in the measure/request mode impl.
> 
> Do you know why and how to fix that?

Yea, I have a rough idea how to go about debugging this

alatiera:
Oops, wrong button, sorry for the spam.

examples/rotation_bin/rotation_bin/mod.rs
sdroege: Putting the kind into the type name seems very strange though :)

examples/rotation_bin/rotation_bin/imp.rs
sdroege: How does it fail exactly and with which code?

examples/rotation_bin/rotation_bin/imp.rs
```diff
@@ -0,0 +1,232 @@
...
+                if orientation == gtk::Orientation::Horizontal {
+                    gtk::Orientation::Vertical
+                } else {
+                    gtk::Orientation::Horizontal
+                }
+            }
+            _ => orientation,
+        };
+
+        match (orientation, self.rotation(widget)) {
```
sdroege, line 150: The things around here also need some serious explanation :)

alatiera:
Fixed the bugs, and even found one in the GTK MR! Only thing left is adding some comments.

examples/rotation_bin/rotation_bin/mod.rs
alatiera: Indeed, and I don't see any other prefix in the rest of the examples, so let's go without I guess.

sdroege:
IMHO it would be more interesting to also rotate some interactive UI element like a button to show that it still works correctly :)

examples/rotation_bin/rotation_bin/mod.rs
alatiera: So this isn't needed somewhere, but its useful if there are apis that give you ints and you want to parse them to something usable. Indeed a TryFrom would have been more appropriate with arbitrary input. I removed it for now since its not used.

examples/rotation_bin/rotation_bin/imp.rs
alatiera: So say on the following code:

```
        let child = match &*self.child.borrow() {
            Some(child) => child,
            None => return (0, 0, -1, -1),
        };
 ```
 
 You get this  error
 
 ```
 error[E0716]: temporary value dropped while borrowed
   --> examples/rotation_bin/rotation_bin/imp.rs:133:29
    |
133 |         let child = match &*self.child.borrow() {
    |                             ^^^^^^^^^^^^^^^^^^^ creates a temporary which is freed while still in use
...
136 |         };
    |          - temporary value is freed at the end of this statement
...
154 |                 let (min, nat, _, _) = child.measure(orientation, for_size);
    |                                        ----- borrow later used here
    |
    = note: consider using a `let` binding to create a longer lived value
```

Which gets easily fixed with this, but then its a bit of clutter and I tend to miss what's going on inside the match or why it happens.
```
        let child = self.child.borrow();
        let child = match child.as_ref() {}
```

Anyhow the code is like this so you can take a look, should add comments on it regardless.

sdroege: Yeah read again the code I wrote above :)

```rust
let child_borrow = self.child.borrow();
let child = match &*child_borrow {
    None => return ...,
    Some(child) => child,
};
```

sdroege: You need to keep the `Ref` alive

sdroege: Your version is OK too of course, same thing

alatiera: oh indeed, tunnel vision. Anyhow, its fine either way

examples/rotation_bin/rotation_bin/imp.rs
```diff
@@ -0,0 +1,232 @@
...
+        };
+
+        let (w, h) = match self.rotation(widget) {
+            Rotation::Deg90 | Rotation::Deg270 => (height, width),
+            _ => (width, height),
+        };
+
+        let transform = gsk::Transform::new();
+        let transform = transform
+            .translate(&Point::new(width as f32 / 2.0, height as f32 / 2.0))
```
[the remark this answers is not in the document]
sdroege: This way is more accurate though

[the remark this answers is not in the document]
alatiera: Indeed, else things would get rounded.

examples/rotation_bin/README.me
```diff
@@ -0,0 +1,5 @@
+# RotationBin
```
bilelmoussaoui, line 1: The extension of this file is wrong :) also the example should be added to the main README in the examples directory.

sdroege:
What's the status here?

alatiera:
Added some comments to the code, though it does still require some familiarity with the methods we override and what its doing to fully grasp, so not sure the comments by themselves answer the "why" is it doing this. Happy to take more suggestions though.

examples/rotation_bin/main.rs
```diff
@@ -0,0 +1,83 @@
...
+fn build_ui(application: &gtk::Application) {
+    let window = gtk::ApplicationWindow::new(application);
+
+    let vbox = gtk::Box::new(gtk::Orientation::Vertical, 6);
+    let rotation_bin = RotationBin::new();
+    let img = gtk::Image::builder()
+        .pixel_size(128)
+        .icon_name("audio-x-generic")
+        .build();
+    rotation_bin.set_child(Some(&img));
```
sdroege, line 28: Don't you think it would be more interesting to also rotate an interactive widget like a button? :)

examples/rotation_bin/rotation_bin/imp.rs
```diff
@@ -0,0 +1,246 @@
...
+                    gtk::Orientation::Horizontal
+                }
+            }
+            _ => orientation,
+        };
+
+        // If the orientation is horizontal and the rotation 0 or 180 degrees,
+        // then pass through the child's measure directly
+        //
+        // Otherwise return child's measure but discard minimum_baseline and natural_baseline
```
sdroege, line 157: Why discard them?

examples/rotation_bin/rotation_bin/imp.rs
```diff
@@ -0,0 +1,246 @@
...
+        };
+
+        // Swap height and width for 90 or 270 degrees
+        let (w, h) = match self.rotation(widget) {
+            Rotation::Deg90 | Rotation::Deg270 => (height, width),
+            _ => (width, height),
+        };
+
+        let transform = gsk::Transform::new();
+        // Transformation are done with respect to their origin point, which is usually a corner.
```
sdroege, line 183: Always top/left of the widget or not?

examples/rotation_bin/rotation_bin/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+impl RotationBin {
+    pub fn new() -> Self {
+        glib::Object::new(&[]).expect("Failed to create a RotationBin")
+    }
+
+    pub fn child(&self) -> Option<impl IsA<gtk::Widget>> {
+        let priv_ = imp::RotationBin::from_instance(self);
```
bilelmoussaoui, line 71:
```suggestion
        let priv_ = self.impl_();
```
and elsewhere

bilelmoussaoui:
Otherwise looks good to me

examples/rotation_bin/rotation_bin/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    pub fn rotate_clockwise(&self) {
+        let priv_ = self.imp();
+        priv_.rotate_clockwise(self);
+    }
+
+    pub fn rotate_counter_clockwise(&self) {
+        let priv_ = self.imp();
+        priv_.rotate_counter_clockwise(self);
```
bilelmoussaoui, line 97:
```suggestion
        self.imp().rotate_counter_clockwise(self);
```
and elsewhere, also why are you passing the public type to the private one in param?

examples/rotation_bin/rotation_bin/imp.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    pub(super) fn set_rotation(&self, obj: &super::RotationBin, rotation: Rotation) {
+        if self.rotation.get() != rotation {
+            self.rotation.replace(rotation);
+            obj.queue_resize();
+            obj.notify("rotation");
+        }
+    }
+
+    pub(super) fn rotate_clockwise(&self, obj: &super::RotationBin) {
```
bilelmoussaoui, line 238: You don't need to pass super::RotationBin here, you can get it from the private struct with `self.instance()` https://gtk-rs.org/gtk-rs-core/git/docs/glib/subclass/types/trait.ObjectSubclassExt.html#tymethod.instance

alatiera: Yes, but I kinda like it having an obj param like the rest of the gtk/gst methods do

bilelmoussaoui: They do so because they are virtual methods and are there for convenience. I do wonder though why we are not using self.instance there, cc @sdroege

examples/rotation_bin/main.rs
sdroege: Nothing but being clickable I guess. Just to show that this doesn't only work for plain pictures but also interactive widgets

examples/rotation_bin/rotation_bin/imp.rs
sdroege: Because you often need it and it's annoying to have to call another function every time :)

bilelmoussaoui:
what is the status of this @alatiera @sdroege ?

alatiera:
> what is the status of this

I think only issue left is making the image an interactive button, let me do that real quick

alatiera:
Added a commit to revert the trust nullability change for gstk Transform

gsk4/Gir.toml
```diff
@@ -172,6 +172,7 @@ status = "generate"
 [[object]]
 name = "Gsk.Transform"
 status = "generate"
+trust_return_value_nullability = false
```
sdroege, line 175: Can you fix the affected functions instead? :) I don't think all of them can return `NULL`, or even most of them not.

alatiera: I can look into it, but not before tomorrow's release :)

None of them are annotated, not even `gsk_transform_translate` or `gsk_transform_invert` so need to do more research and ask people that know how Transforms are implemented.

bilelmoussaoui:
Thanks @alatiera for the example

gsk4/Gir.toml
sdroege: The only one that can return `NULL` is `gsk_transform_invert` and that is annotated accordingly already.

PR #751, opened 2021-12-10
gtk: manually implement callback getters

bilelmoussaoui:
Not sure if they are worth it but those are part of the very few remaining "not generated" functions

### DIFF
--- a/gtk4/Gir.toml
+++ b/gtk4/Gir.toml
@@ -855,6 +855,10 @@ manual_traits = ["ComboBoxExtManual"]
     manual = true # use Option<u32>
     doc_trait_name = "ComboBoxExtManual"
     [[object.function]]
+    name = "get_row_separator_func"
+    # getting the callbacks that could have been set from C is impossible and the api itself is not that useful
+    ignore = true
+    [[object.function]]
     name = "set_row_separator_func"
         [[object.function.parameter]]
         name = "func"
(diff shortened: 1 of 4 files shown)


gtk4/src/combo_box.rs
```diff
@@ -47,4 +53,13 @@ impl<O: IsA<ComboBox>> ComboBoxExtManual for O {
             x => Some(x as _),
         }
     }
+
+    fn row_separator_func<F: Fn(&TreeModel, &TreeIter) -> bool + 'static>(
+        &self,
+    ) -> Option<Box_<F>> {
+        unsafe {
+            let func = ffi::gtk_combo_box_get_row_separator_func(self.as_ref().to_glib_none().0);
```
sdroege, line 61: That won't do. The function could've also been set by C so it's not necessarily a `Box`, and also otherwise you would need to get it with the very exact closure type here (impossible) or use at the setter and getter a `dyn Fn` as type.

Also: the pointer returned here is the C function / trampoline, the boxed closure is the `user_data` (which you can't get).

If we could also get the `user_data` then we could make a wrapper around that in theory... but as we can't guarantee its lifetime that would also be unsafe.

I'd say, the way how the C API is right now you should just forget binding these functions :)

bilelmoussaoui: That's unfortunate I guess :) but also from a rust app use case, i don't think those functions are very useful and they are only used in widgets that will hopefully be deprecated in future releases and replaced with the new ones (all of them are from the CellRenderer's era).

I will go ahead and mark them as ignored then with a comment explaining why

sdroege (approved):

PR #752, opened 2021-12-10
gdk: Cursor is Send

elmarco:
It only has read-only methods.

### DIFF
--- a/gdk4/Gir.toml
+++ b/gdk4/Gir.toml
@@ -392,6 +392,7 @@ status = "generate"
 name = "Gdk.Cursor"
 status = "generate"
 generate_builder = true
+concurrency = "send"
 
 [[object]]
 name = "Gdk.Device"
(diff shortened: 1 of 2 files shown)


sdroege:
I think that's wrong. The cursor can have a reference to a texture, which might be backed by OS resources (e.g. GL, Windows HWND, etc) and that's clearly not `Send`.

elmarco:
Hmm.

https://docs.gtk.org/gdk4/class.Cursor.html#description

Cursors are immutable objects, so once you created them, there is no way to modify them later. You should create a new cursor when you want to change something about it.

Same for GdkTexture actually.

sdroege:
Yes but you can't access textures from any other thread than the main thread (with the exception of memory textures).

elmarco:
hmm ok

sdroege:
That's generally the rule for anything in GDK that works with system resources.

PR #753, opened 2021-12-12
mark dependencies as optional

### LINKED ISSUE #676: Make the x11/wayland specific crates optional
Currently gdk-x11 & gdk-wayland depends on rust crates for specific features and those dependencies are enabled by default even if you don't need them.

Instead of this situation, we can still keep using them but put them behind a feature flag & make it enabled by default. This way people not needing them can still save few crates on their lib/gui app

cc @elmarco

bilelmoussaoui:
The first commit is unrelated to the PR, it drops some functions that can't be generated nor implemented

Fixes #676


bilelmoussaoui:
cc @GuillaumeGomez, would be nice to have something similar in gtk3-rs as well someday

PR #757, opened 2021-12-14
Fix code snippets and some formatting in gobject_memory_management

hbina:
Signed-off-by: Hanif Ariffin <hanif.ariffin.4326@gmail.com>


Hofer-Julian:
Thanks, seems like these anchors got lost during a refactor.

PR #761, opened 2021-12-16
bump version number

bilelmoussaoui:


sdroege:
Why? Don't we want to release everything in sync soonish? :)

bilelmoussaoui:
> Why? Don't we want to release everything in sync soonish? :)

Yes, but I'm not doing a release heh. I didn't do a bump post the latest release

bilelmoussaoui:
> Why? Don't we want to release everything in sync soonish? :)

Also, the reason I set the versions of gtk-rs-core crates to 0.15 even if we use git for now; is to simplify the needed changes once we have to do a release by just dropping the git part of each dependency.

sdroege:
Sure, the version should always be bumped directly after release also for `cargo` un-confusion reasons :)

PR #775, opened 2021-12-26
gtk: Rework AccessibleExtManual

jf2048:
Fixes a few issues:

- Dynamic typing with `ToValue` isn't necessary as the properties/relations/states all have fixed types.
- Using `ToValue` is not ideal because there's no way to see type errors besides checking for gtk criticals.
- Makes it possible to safely set relations of type ref list.
- Handles the weirdness where some states expect a gboolean that can be -1, or where enums need a value of `G_TYPE_INT`.

### DIFF
--- a/gtk4/src/accessible.rs
+++ b/gtk4/src/accessible.rs
@@ -1,77 +1,341 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use crate::{Accessible, AccessibleProperty, AccessibleRelation, AccessibleState};
-use glib::translate::*;
+use crate::{
+    Accessible, AccessibleAutocomplete, AccessibleInvalidState, AccessibleProperty,
+    AccessibleRelation, AccessibleSort, AccessibleState, AccessibleTristate, Orientation,
+};
+use glib::{translate::*, Value};
 use glib::{IsA, ToValue};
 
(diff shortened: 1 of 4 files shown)


gtk4/src/accessible.rs
```diff
@@ -75,3 +81,232 @@ impl<O: IsA<Accessible>> AccessibleExtManual for O {
         }
     }
 }
+
+pub enum Property<'p> {
```
bilelmoussaoui, line 85: Ideally those enums could use some documentation & be exposed in lib.rs;

bilelmoussaoui: Also needs #[non_exhaustive] & probably implement stuff like Debug

bilelmoussaoui:
Looks good thanks for your work on this! The api is much more usable & less cryptic this way :)

gtk4/src/accessible.rs
jf2048: Is there an easy way to get the docs generator to copy the docs from `GtkAccessibleProperty`, etc? That would probably make the most sense.

bilelmoussaoui: No, that's not possible yet. We didn't have such use case so far.

bilelmoussaoui:
lgtm, thanks @jf2048.

PR #776, opened 2021-12-27
Use imp() everywhere instead of from_instance(), impl_()

jf2048:


bilelmoussaoui:
Would be nice if you can also update the examples

Hofer-Julian:
@jf2048 thanks for working this!

@bilelmoussaoui @sdroege now that all of this places are changed anyways it would be good time to decide on one consistent variable name. We have now self_ and imp. Both are a bit weird when the method is called impl_, but unfortunately @sdroege vetoed imp there.

sdroege:
`impl_` or `imp`, not `self_` (that's confusing with other actual `self`)

bilelmoussaoui:
> `impl_` or `imp`, not `self_` (that's confusing with other actual `self`)

imp sounds good to me

sdroege:
Only has the problem that it's an actual word: :imp:

Hofer-Julian:
+1

@sdroege you still believe that the method shouldn't be called imp as well?

Considering how much confusion this concept creates, I feel a bit of consistency wouldn't hurt.

sdroege:
> @sdroege you still believe that the method shouldn't be called imp as well?

It should be consistent, yes. And we also have `imp.rs`.

Hofer-Julian:
> > @sdroege you still believe that the method shouldn't be called imp as well?
> 
> It should be consistent, yes. And we also have `imp.rs`.

👍

@jf2048 could you please rename the few instances of self_ to imp?

@bilelmoussaoui @sdroege fine by you if I open a PR soon to rename the method impl_ to imp and fix the examples and book here?

Hofer-Julian (approved):
Looks good to me.
Thanks a lot @jf2048!

jf2048:
Let's wait for https://github.com/gtk-rs/gtk-rs-core/pull/437 and then I'll rename everything here again.

Hofer-Julian:
> Let's wait for [gtk-rs/gtk-rs-core#437](https://github.com/gtk-rs/gtk-rs-core/pull/437) and then I'll rename everything here again.

That would be nice!

pbor:
I guess I am late to the party here... in our code we are using `priv()` as that is consistent with how the same thing was done in C

bilelmoussaoui:
> I guess I am late to the party here... in our code we are using `priv()` as that is consistent with how the same thing was done in C

but in rust, there's no reason to name your private struct as `MyTypePrivate` as you can just put that behind a `mod imp` or `mod private`

pbor:
> 
> but in rust, there's no reason to name your private struct as `MyTypePrivate` as you can just put that behind a `mod imp` or `mod private`

Sure, I was not suggesting to use `MyTypePrivate`, I was suggesting to use `mod priv` and `priv()` for the function, since "priv" does not clash with "impl", it is not a word like "imp", and preserves some familiarity for those who come from C or look the code of glib and gtk.

bilelmoussaoui:
Sure, I'm pretty bad at naming things in general so better avoid my opinion on naming things

sdroege:
priv is also a reserved keyword IIRC

sdroege:
Otherwise I would prefer that naming

sdroege:
See https://doc.rust-lang.org/reference/keywords.html#reserved-keywords

Hofer-Julian:
> See https://doc.rust-lang.org/reference/keywords.html#reserved-keywords

I wouldn't mind priv (if it wouldn't be reserved) and not even private. But considering that we use imp already for years and private is longer, I'd rather stick with imp.

PR #777, opened 2021-12-28
texture: Add async wrappers

A6GibKm:


sdroege:
While this is "async", it still does all the work on the main thread. That's not very useful :)

gdk4/src/texture.rs
```diff
@@ -18,3 +21,47 @@ impl<O: IsA<Texture>> TextureExtManual for O {
...
+            glib::clone!(@weak pixbuf => move |_, _, send| {
+                let res = Texture::for_pixbuf(pixbuf);
+                send.resolve(res);
+            }),
+        ))
+    }
+
+    pub fn from_bytes_future(
+        bytes: &glib::Bytes,
+    ) -> Pin<Box<dyn future::Future<Output = Result<Texture, glib::Error>> + 'static>> {
```
bilelmoussaoui, line 40: As those are not in a trait, why make use of Pin<Box<dyn future... instead of just marking the function as `pub async fn`?

gdk4/src/texture.rs
```diff
@@ -18,3 +21,47 @@ impl<O: IsA<Texture>> TextureExtManual for O {
         }
     }
 }
+
+impl Texture {
```
bilelmoussaoui, line 25: Are all those variants thread-safe?

A6GibKm: According to the docs these functions are thread safe, ill double check that all of them are tho.

gdk4/src/texture.rs
```diff
@@ -18,3 +21,51 @@ impl<O: IsA<Texture>> TextureExtManual for O {
...
+                gio::Cancellable::NONE,
+                move |result, _| {
+                    let texture = result
+                        .downcast_ref::<gio::Task>()
+                        .unwrap()
+                        .propagate_value()
+                        .unwrap()
+                        .get::<Texture>()
+                        .unwrap();
+                    if let Some(m) = sender.replace(None) {
```
sdroege, line 45: The closure is a `FnOnce`, you don't need the `Cell` and `Option` here

gdk4/src/texture.rs
```diff
@@ -18,3 +21,51 @@ impl<O: IsA<Texture>> TextureExtManual for O {
...
+    pub fn for_pixbuf_future(
+        pixbuf: &'static gdk_pixbuf::Pixbuf,
+    ) -> Pin<Box<dyn Future<Output = Option<Texture>> + 'static>> {
+        Box::pin(async move {
+            let (sender, receiver) = futures_channel::oneshot::channel();
+
+            let sender = std::cell::Cell::new(Some(sender));
+
+            let task = gio::Task::new(
+                Some(&pixbuf.upcast_ref()),
```
sdroege, line 35: This should really take an `IsA<Object>` instead of `Object` (and then also use the same type for the callback)

gdk4/src/texture.rs
```diff
@@ -18,3 +21,51 @@ impl<O: IsA<Texture>> TextureExtManual for O {
...
+                gio::Cancellable::NONE,
+                move |result, _| {
+                    let texture = result
+                        .downcast_ref::<gio::Task>()
```
sdroege, line 39: @pbor This is always a `gio::Task` or not? Can we update the closure signature maybe?

gdk4/src/texture.rs
```diff
@@ -18,3 +21,51 @@ impl<O: IsA<Texture>> TextureExtManual for O {
...
+
+            let task_func = move |task: &gio::Task,
+                                  source_object: Option<&glib::Object>,
+                                  _: Option<&gio::Cancellable>| {
+                let pixbuf = source_object
+                    .unwrap()
+                    .downcast_ref::<gdk_pixbuf::Pixbuf>()
+                    .unwrap();
+                let texture = Texture::for_pixbuf(&pixbuf);
+                task.return_value(&texture.to_value());
```
sdroege, line 59: @pbor The API is unsound, oh no. You can pass a non-`Send` value from the `run_in_thread()` callback back to the main thread.

gdk4/src/texture.rs
```diff
@@ -18,3 +21,51 @@ impl<O: IsA<Texture>> TextureExtManual for O {
...
+            };
+
+            task.run_in_thread(task_func);
+
+            if let Ok(texture) = receiver.await {
```
sdroege, line 64: This should probably be a `Result` that is returned here

A6GibKm: I am not sure whats better, the original api does not return a Result, (but a result is used in the other functions that I plan to implement here).

gdk4/src/texture.rs
A6GibKm: I don't follow

sdroege: The `gio::Task::new()` API should be improved so you don't need to cast that much

gdk4/src/texture.rs
sdroege: The original API should also return a result (and you can configure it in `Gir.toml` like that)

gdk4/src/texture.rs
A6GibKm: Yes they all are.

gdk4/src/texture.rs
pbor: I guess... I seem to recall the reason why the C code uses `GAsyncResult` was mostly historical?

gdk4/src/texture.rs
pbor: :(

gdk4/src/texture.rs
```diff
@@ -18,3 +18,175 @@ impl<O: IsA<Texture>> TextureExtManual for O {
...
+
+        texture_result
+    }
+
+    pub async fn from_file_future(file: &impl IsA<gio::File>) -> Result<Texture, glib::Error> {
+        let (sender, receiver) = futures_channel::oneshot::channel();
+
+        let task = gio::Task::new(
+            Some(file.as_ref().upcast_ref()),
+            gio::Cancellable::NONE,
```
melix99, line 95: Is it possible to make them cancellable? I might need that.

bilelmoussaoui: Probably, by exposing the GTask's cancellable as param. But then, not sure how you would cancel the ongoing operation..

melix99: Yeah, I don't think the operation can be cancelled, but I guess it could at least return an error when the operation is done, which is still somewhat useful imho.

gdk4/src/texture.rs
A6GibKm: The original api cannot fail, if for some reason the operation fails it will instead do a g_error. I am not sure how to handle that given that it will happen in another thread.

sdroege: `g_error()` immediately kills the whole process, not just a thread :) There's no way you can handle that.

bilelmoussaoui:
I have mentioned before, ideally those wrappers would be provided by gdk itself.

sdroege:
This needs to be updated once https://github.com/gtk-rs/gtk-rs-core/pull/446 is merged

gdk4/src/texture.rs
```diff
@@ -31,3 +31,137 @@ impl<O: IsA<Texture>> TextureExtManual for O {
...
     }
 }
+
+impl Texture {
+    pub async fn for_pixbuf_future(pixbuf: &gdk_pixbuf::Pixbuf) -> Result<Texture, glib::Error> {
+        assert_initialized_main_thread!();
+        let (sender, receiver) = futures_channel::oneshot::channel();
+
+        let task = gio::Task::new(
+            Some(pixbuf.upcast_ref()),
```
bilelmoussaoui, line 41: you don't need to upcast_ref here, just `pixbuf` should be enough?

gdk4/src/texture.rs
```diff
@@ -31,3 +31,137 @@ impl<O: IsA<Texture>> TextureExtManual for O {
...
+            receiver.await.unwrap()
+        }
+    }
+
+    pub async fn from_file_future(file: &impl IsA<gio::File>) -> Result<Texture, glib::Error> {
+        assert_initialized_main_thread!();
+        let (sender, receiver) = futures_channel::oneshot::channel();
+
+        let task = gio::Task::new(
+            Some(file.as_ref().upcast_ref()),
```
bilelmoussaoui, line 94: Same here, passing file.as_ref should be enough no?

gdk4/src/texture.rs
A6GibKm: I don't see a clear way forward for pixbufs, so ill remove that one.

gdk4/src/texture.rs
```diff
@@ -31,3 +31,107 @@ impl<O: IsA<Texture>> TextureExtManual for O {
         }
     }
 }
+
+// TODO Figure out how to implement an async variant for for_pixbuf.
+impl Texture {
+    #[cfg(any(feature = "v4_6", feature = "dox"))]
```
bilelmoussaoui, line 37: All of those new functions could have a:
1 - Very simple code example or a test case if possible
2 - a bit of docs
3 - point to the upstream issue and mention that we should drop those manual implementations once those are fixed upstream

bilelmoussaoui:
What is the status quo here?

A6GibKm:
That I am not sure if this is desired to have in the bindings.

bilelmoussaoui:
I still believe this should be done upstream, so let us close this PR.

PR #780, opened 2021-12-29
Generate fundamental types code using gir

bilelmoussaoui:
Depends on https://github.com/gtk-rs/gir/pull/1294

Still needs a bit of cleanup & figure out how to handle the builder pattern, but in general it drops a bunch of manual code and that makes me happy :)


bilelmoussaoui:
Before merging this, we need to check the functions signatures one by one to ensure everything is properly generated & fix whatever needed.

bilelmoussaoui:
So I did a 1:1 check (manual implementations we had vs auto-generated ones): 

Here are a couple of things i noticed
- [x] `gtk_property_expression_get_expression` -> nullable : fixed upstream, see https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4304
- [x] `gsk_container_node_get_child` -> assert n_children < i # fixed in the latest commit
- [x] `gsk_gl_shader_node_get_child` -> assert n_children < i # fixed in the latest commit

Those that were not nullable in the manual bindings but are in the auto-generated ones. There are a ton of others that should really not be nullable because the constructor doesn't take a nullable variant and are nullable because of #68. So for those I would rather leave them as is for now and tackle the issue on #68  

- [ ] `gsk_border_node_get_outline` -> not nullable
- [ ] `gsk_conic_gradient_node_get_center` -> not nullable
- [ ] `gdk_event_get_event_sequence` -> not nullable
- [ ] `gsk_linear_gradient_node_get_end` -> not nullable
- [ ] `gsk_linear_gradient_node_get_start` -> not nullable

gdk4/Gir.toml
```diff
@@ -498,8 +501,8 @@ status = "generate"
 
 [[object]]
 name = "Gdk.Event"
-status = "manual" # fundemental type
-final_type = true # avoids the usage of EventExt trait
+status = "generate" # fundemental type
```
sdroege, line 504:
```suggestion
status = "generate" # fundamental type
```

sdroege: And in many other places, including commit messages :)

gdk4/src/auto/event.rs
```diff
@@ -0,0 +1,204 @@
...
+    #[doc(alias = "get_axis")]
+    pub fn axis(&self, axis_use: AxisUse) -> Option<f64> {
+        unsafe {
+            let mut value = mem::MaybeUninit::uninit();
+            let ret = from_glib(ffi::gdk_event_get_axis(
+                self.as_ref().to_glib_none().0,
+                axis_use.into_glib(),
+                value.as_mut_ptr(),
+            ));
+            let value = value.assume_init();
```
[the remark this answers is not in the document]
sdroege: Yes

gdk4/src/auto/event.rs
```diff
@@ -0,0 +1,205 @@
...
+                Some((x, y))
+            } else {
+                None
+            }
+        }
+    }
+
+    #[doc(alias = "gdk_event_get_seat")]
+    #[doc(alias = "get_seat")]
+    pub fn seat(&self) -> Option<Seat> {
```
sdroege, line 181: This is not `nullable`, is it? Probably similar problem for other functions

gdk4/src/auto/key_event.rs
```diff
@@ -0,0 +1,59 @@
...
+        unsafe {
+            from_glib(ffi::gdk_key_event_get_consumed_modifiers(
+                self.to_glib_none().0,
+            ))
+        }
+    }
+
+    #[doc(alias = "gdk_key_event_get_keycode")]
+    #[doc(alias = "get_keycode")]
+    pub fn keycode(&self) -> u32 {
```
sdroege, line 39: Shouldn't this return a `gdk::Key`?

gdk4/src/auto/motion_event.rs
```diff
@@ -0,0 +1,30 @@
...
+    #[doc(alias = "GdkMotionEvent")]
+    pub struct MotionEvent(Shared<ffi::GdkMotionEvent>);
+
+    match fn {
+        ref => |ptr| ffi::gdk_event_ref(ptr as *mut ffi::GdkEvent),
+        unref => |ptr| ffi::gdk_event_unref(ptr as *mut ffi::GdkEvent),
+    }
+}
+
+impl glib::StaticType for MotionEvent {
```
sdroege, line 20: Why is this not moved into `glib::wrapper!`? It can do `type_ => ...` for `Shared` types

gsk4/Gir.toml
```diff
@@ -142,6 +197,16 @@ status = "generate"
     name = "free_to_args"
     ignore = true # C convenience functions
 
+[[object]]
+name = "Gsk.ShadowNode"
+status = "generate"
+    [[object.function]]
+    name = "new"
+    manual = true # can't work due to &[&CopyType] issue
```
sdroege, line 205: Wasn't this solved?

gsk4/Gir.toml
```diff
@@ -115,25 +136,59 @@ generate_builder = true
...
+    name = "new"
     manual = true
-    doc_trait_name = "GskRendererExtManual"
+
+[[object]]
+name = "Gsk.RadialGradientNode"
+status = "generate"
+    [[object.function]]
+    name = "new"
+    manual = true
```
sdroege, line 160: Why are all these manual?

sdroege: i.e. add comments here

gsk4/src/auto/cairo_node.rs
```diff
@@ -0,0 +1,50 @@
...
+impl CairoNode {
+    #[doc(alias = "gsk_cairo_node_new")]
+    pub fn new(bounds: &graphene::Rect) -> CairoNode {
+        assert_initialized_main_thread!();
+        unsafe { from_glib_full(ffi::gsk_cairo_node_new(bounds.to_glib_none().0)) }
+    }
+
+    #[doc(alias = "gsk_cairo_node_get_draw_context")]
+    #[doc(alias = "get_draw_context")]
+    pub fn draw_context(&self) -> Option<cairo::Context> {
```
sdroege, line 35: This is also not `nullable` AFAICS

gsk4/src/auto/cairo_node.rs
```diff
@@ -0,0 +1,50 @@
...
+        unsafe { from_glib_full(ffi::gsk_cairo_node_get_draw_context(self.to_glib_none().0)) }
+    }
+
+    #[doc(alias = "gsk_cairo_node_get_surface")]
+    #[doc(alias = "get_surface")]
+    pub fn surface(&self) -> Option<cairo::Surface> {
```
sdroege, line 41: Also not `nullable`

gsk4/src/auto/color_node.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl glib::StaticType for ColorNode {
+    fn static_type() -> glib::Type {
+        unsafe { from_glib(ffi::gsk_color_node_get_type()) }
+    }
+}
+
+impl ColorNode {
+    #[doc(alias = "gsk_color_node_new")]
+    pub fn new(rgba: &gdk::RGBA, bounds: &graphene::Rect) -> ColorNode {
```
sdroege, line 28: I think clippy would complain here that we pass `Copy` types by reference?

sdroege: (I think passing by reference here is the right thing to do)

gsk4/src/auto/debug_node.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    #[doc(alias = "gsk_debug_node_get_child")]
+    #[doc(alias = "get_child")]
+    pub fn child(&self) -> Option<RenderNode> {
+        unsafe { from_glib_none(ffi::gsk_debug_node_get_child(self.to_glib_none().0)) }
+    }
+
+    #[doc(alias = "gsk_debug_node_get_message")]
+    #[doc(alias = "get_message")]
+    pub fn message(&self) -> Option<glib::GString> {
```
sdroege, line 46: Also not `nullable`, best if you could check everything returning an `Option`

sdroege:
Generally looks good to me otherwise

sdroege:
> There are a ton of others that should really not be nullable because the constructor doesn't take a nullable variant and are nullable because of #68. So for those I would rather leave them as is for now and tackle the issue on #68

You can configure the return type to be not `nullable` in the `Gir.toml` for the time being.

gsk4/Gir.toml
bilelmoussaoui: Yes, but in a different merge request, whatever lands first on Gir's side

gdk4/src/auto/event.rs
bilelmoussaoui: Yes, see my latest comment. They were nullable in the manual bindings as well

gdk4/src/auto/key_event.rs
bilelmoussaoui: For keyval yes, not sure about keycode

gdk4/src/auto/motion_event.rs
bilelmoussaoui: See the gir PR, it auto implement Value traits which is not a desired behaviour

gsk4/Gir.toml
bilelmoussaoui: They will all be auto generated once the other gir PR is merged. Eg, see the other gtk4-rs PR

bilelmoussaoui:
> > There are a ton of others that should really not be nullable because the constructor doesn't take a nullable variant and are nullable because of #68. So for those I would rather leave them as is for now and tackle the issue on #68
> 
> You can configure the return type to be not `nullable` in the `Gir.toml` for the time being.

Sure, but most of them were like that. I plan to tackle the nullability issue just after getting those two PR merged so it's fine I guess

gsk4/src/auto/color_node.rs
bilelmoussaoui: > I think clippy would complain here that we pass `Copy` types by reference?

Somehow it doesn't, no idea why :)

gdk4/src/auto/key_event.rs
sdroege: Ah right, there's key(val) and keycode. I got confused :)

gdk4/src/auto/motion_event.rs
sdroege: Right, it would need the value functions somehow then.

gsk4/src/auto/color_node.rs
sdroege: Good then!

sdroege:
> Sure, but most of them were like that. I plan to tackle the nullability issue just after getting those two PR merged so it's fine I guess

Ok, let's make sure to fix that before the release though.

bilelmoussaoui:
> > Sure, but most of them were like that. I plan to tackle the nullability issue just after getting those two PR merged so it's fine I guess
> 
> Ok, let's make sure to fix that before the release though.

Sure thing

bilelmoussaoui:
Will merge this & rebase the other one once the gir fix lands

PR #781, opened 2021-12-29
 auto-generate slice of copy types functions 

bilelmoussaoui:
Depends on #780  & https://github.com/gtk-rs/gir/pull/1296


gsk4/src/auto/gl_shader_node.rs
```diff
@@ -24,6 +24,26 @@ impl glib::StaticType for GLShaderNode {
...
+        children: &[RenderNode],
+    ) -> GLShaderNode {
+        skip_assert_initialized!();
+        let n_children = children.len() as u32;
+        unsafe {
+            from_glib_full(ffi::gsk_gl_shader_node_new(
+                shader.to_glib_none().0,
+                bounds.to_glib_none().0,
+                args.to_glib_none().0,
+                children.to_glib_none().0,
```
sdroege, line 41: This and all the others are needlessly creating a temporary `Vec`, which we should fix at the same time here. `&[RenderNode]` is equivalent to `&[*mut gsk_sys::RenderNode]` i.e. `*const *mut gsk_sys::RenderNode`.

sdroege:
Seems good to me otherwise

gsk4/src/auto/gl_shader_node.rs
bilelmoussaoui: That would need to be fixed on glib-rs side right ?

sdroege: Yes, that should be an easy fix there. The `ToGlibContainerFromSlice` trait impl

PR #792, opened 2022-01-04
gtk: use ToVariant for Actionable::set_action_target

### LINKED ISSUE #767: [BUG] Error when using ActionableExtManual::set_action_target
I tried to use `ActionableExtManual::set_action_target` like this:

```rust
button.set_action_target(&string));
```

Every time I get these errors in the logs:

```shell
GLib-CRITICAL **: 16:31:11.017: '%s' is not a valid GVariant format string
GLib-CRITICAL **: 16:31:11.017: g_variant_new_va: assertion 'valid_format_string (format_string, !endptr, NULL)' failed
```
It works however with `ActionableExt::set_action_target_value`, like this:

```rust
button.set_action_target_value(Some(&string.to_variant()));
```

bilelmoussaoui:
Fixes #767

### DIFF
--- a/gtk4/src/actionable.rs
+++ b/gtk4/src/actionable.rs
@@ -1,25 +1,23 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 use crate::Actionable;
-use glib::object::IsA;
 use glib::translate::*;
-use libc::c_char;
+use glib::{IsA, ToVariant};
 
 // rustdoc-stripper-ignore-next
 /// Trait containing manually implemented methods of [`Actionable`](crate::Actionable).
 pub trait ActionableExtManual: 'static {
     #[doc(alias = "gtk_actionable_set_action_target")]
-    fn set_action_target(&self, string: &str);
+    #[doc(alias = "gtk_actionable_set_action_target_value")]
+    fn set_action_target(&self, target: Option<&impl ToVariant>);
 }
 
 impl<O: IsA<Actionable>> ActionableExtManual for O {
-    fn set_action_target(&self, string: &str) {
-        let string: Stash<*const c_char, _> = string.to_glib_none();
+    fn set_action_target(&self, target: Option<&impl ToVariant>) {
         unsafe {
-            ffi::gtk_actionable_set_action_target(
+            ffi::gtk_actionable_set_action_target_value(
                 self.as_ref().to_glib_none().0,
-                b"%s\0".as_ptr() as *const c_char,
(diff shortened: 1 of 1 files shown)


gtk4/src/actionable.rs
```diff
@@ -1,25 +1,22 @@
...
 use glib::translate::*;
-use libc::c_char;
+use glib::{IsA, ToVariant};
 
 // rustdoc-stripper-ignore-next
 /// Trait containing manually implemented methods of [`Actionable`](crate::Actionable).
 pub trait ActionableExtManual: 'static {
     #[doc(alias = "gtk_actionable_set_action_target")]
-    fn set_action_target(&self, string: &str);
+    fn set_action_target<V: ToVariant>(&self, string: &V);
```
sdroege, line 11:
```suggestion
    fn set_action_target<V: ToVariant>(&self, target: &V);
```

Not really a string, is it? :)

sdroege: But also ...

gtk4/src/actionable.rs
```diff
@@ -1,25 +1,22 @@
...
 }
 
 impl<O: IsA<Actionable>> ActionableExtManual for O {
-    fn set_action_target(&self, string: &str) {
-        let string: Stash<*const c_char, _> = string.to_glib_none();
+    fn set_action_target<V: ToVariant>(&self, string: &V) {
```
sdroege, line 15:
```suggestion
    fn set_action_target<V: ToVariant>(&self) {
```

You're not really using the value itself...

gtk4/src/actionable.rs
```diff
@@ -1,25 +1,22 @@
...
         unsafe {
             ffi::gtk_actionable_set_action_target(
                 self.as_ref().to_glib_none().0,
-                b"%s\0".as_ptr() as *const c_char,
-                string.0,
+                string.to_variant().type_().as_str().to_glib_none().0,
```
sdroege, line 19: ... and can do just

```suggestion
                V::static_variant_type.to_glib_none().0,
```

sdroege: However I think this is wrong. This function wants a variant format string and then the values of the variant.

I think you want to call `gtk_actionable_set_action_target_value(v.to_variant().to_glib_none().0)`. Note that this function's parameter is `nullable`.

sdroege (changes requested):

sdroege (approved):

PR #794, opened 2022-01-05
gtk: provide gdk backend information to dependencies

### LINKED ISSUE #789: Provide GDK backend information to dependencies
See https://github.com/gtk-rs/gtk3-rs/blob/master/gdk/sys/build.rs#L39

bilelmoussaoui:
Copied from gdk3 with a bit of changes as GTK 4 no longer provides a pkg-config file for gdk
Fixes #789


gtk4/sys/build.rs
```diff
@@ -11,7 +15,37 @@ fn main() {} // prevent linking libraries to avoid documentation failure
...
         process::exit(1);
     }
+
+    // It's safe to assume we can call this because we found the library OK
+    // in find()
+    check_features();
+}
+
+#[cfg(not(feature = "dox"))]
+fn check_features() {
```
sdroege, line 28: It would be best to do this in the `gdk4-sys` crate. Why it's the same pkg-config file, the backend is relevant to gdk and not to gtk.

sdroege (approved):

PR #795, opened 2022-01-06
gdk: Implement Send+Sync for texture

A6GibKm:
From the docs they are immutable and safe
https://docs.gtk.org/gdk4/class.Texture.html

See https://gitlab.gnome.org/GNOME/gtk/-/commit/6785461c2651ad502f46521536619ea77f04dd63

### DIFF
--- a/gdk4/Gir.toml
+++ b/gdk4/Gir.toml
@@ -44,7 +44,6 @@ generate = [
     "Gdk.Gravity",
     "Gdk.InputSource",
     "Gdk.KeyMatch",
-    "Gdk.MemoryTexture",
     "Gdk.ModifierType",
     "Gdk.MotionEvent",
     "Gdk.NotifyType",
@@ -451,6 +450,7 @@ final_type = false
 [[object]]
 name = "Gdk.GLTexture"
 status = "generate"
+concurrency = "send+sync"
     [[object.function]]
     name = "new"
     manual = true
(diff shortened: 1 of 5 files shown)


gdk4/src/texture.rs
```diff
@@ -5,6 +5,9 @@ use crate::Texture;
 use glib::translate::*;
 use glib::IsA;
 
+unsafe impl Send for Texture {}
+unsafe impl Sync for Texture {}
```
sdroege, line 9: Do something like this instead:

https://github.com/gtk-rs/gtk-rs-core/blob/c53f692d84f7f7a3a75362cd424427842f792e66/gio/Gir.toml#L331

Also the same for all `GdkTexture` subclasses (memory, gl)

A6GibKm: Done, thanks.

sdroege (approved):

sdroege:
Needs https://github.com/gtk-rs/gtk4-rs

bilelmoussaoui:
Please rebase

sdroege:
Put an `#![allow(clippy::non_send_fields_in_send_ty)]` into `lib.rs` at the top

A6GibKm:
> Put an `#![allow(clippy::non_send_fields_in_send_ty)]` into `lib.rs` at the top

We also have a 
```
#[allow(unknown_lints)]
```

bilelmoussaoui:
> > Put an `#![allow(clippy::non_send_fields_in_send_ty)]` into `lib.rs` at the top
> 
> We also have a
> 
> ```
> #[allow(unknown_lints)]
> ```

could be done there as well no?

bilelmoussaoui:
that easier to remove later once that lint is dropped from clippy

PR #797, opened 2022-01-06
gtk: Automatically generate ParamSpecExpression

A6GibKm:


gtk4/src/param_spec_expression.rs
```diff
@@ -1,26 +1,11 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
+use crate::ParamSpecExpression;
+
 use glib::gobject_ffi;
 use glib::translate::*;
 use glib::ParamSpec;
 
-glib::wrapper! {
-    #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
bilelmoussaoui, line 8: you forgot to remove it from lib.rs

bilelmoussaoui: Also copy the value trait implementations from https://github.com/gtk-rs/gtk-rs-core/blob/master/glib/src/param_spec.rs#L37-L84  those were wrongly implemented in the past

gtk4/Gir.toml
```diff
@@ -1606,6 +1605,12 @@ status = "generate"
     pattern = "*-set" # those properties are read only
     generate = ["get", "notify"]
 
+[[object]]
+name = "Gtk.ParamSpecExpression"
+status = "generate"
+manual = true
```
bilelmoussaoui, line 1611:
```suggestion
```
not sure where you got this one from but it is not needed

gtk4/src/param_spec_expression.rs
bilelmoussaoui, line 8: As those derives are not auto-generated for shared types, i would at least implement Debug manually

A6GibKm: How do I add those traits?

bilelmoussaoui: just add the debug one, copy paste the automatically generate Display and replace Display with Debug :)

gtk4/src/param_spec_expression.rs
```diff
@@ -62,6 +53,55 @@ impl ParamSpecExpression {
     }
 }
 
+#[doc(hidden)]
+impl glib::value::ValueType for ParamSpecExpression {
+    type Type = ParamSpec;
```
bilelmoussaoui, line 58:
```suggestion
    type Type = ParamSpecExpression;
```

gtk4/src/param_spec_expression.rs
```diff
@@ -62,6 +53,55 @@ impl ParamSpecExpression {
...
+        assert!(!ptr.is_null());
+        from_glib_full(ptr as *mut gobject_ffi::GParamSpec)
+    }
+}
+
+#[doc(hidden)]
+impl glib::value::ToValue for ParamSpecExpression {
+    fn to_value(&self) -> Value {
+        unsafe {
+            let mut value = Value::from_type(ParamSpecExpression::static_type());
```
bilelmoussaoui, line 79:
```suggestion
            let mut value = Value::from_type(Self::static_type());
```

bilelmoussaoui (changes requested):

bilelmoussaoui:
Thank you!

PR #798, opened 2022-01-07
CI: add Windows MSCV

bilelmoussaoui:
Replaces #697


bilelmoussaoui:
It should work fine now but it is super slow to build gtk as it pulls the universe. We definitely need to cache the build for sure before getting this one merged

bilelmoussaoui:
Let us land this and see if the cache works properly

PR #800, opened 2022-01-08
CI: Stop passing manifest path to windows build

Hofer-Julian:
I want to check which crates are built in this case.


Hofer-Julian:
Mmmh, we'll have to wait a bit there until the problem with freedesktop gitlab is resolved: https://gitlab.freedesktop.org/freedesktop/freedesktop/-/issues/407

bilelmoussaoui:
Seems to work but needs to be adapted somehow to find the headers for the abi tests

bilelmoussaoui:
It should normally work fine now :)

PR #803, opened 2022-01-09
gtk/gdk: mark Snapshot as not final and implement necessary traits

### LINKED ISSUE #799: [BUG] `gdk::Paintable.snapshot()` doesn't accept `gtk::Snapshot`
**Bug description**

The [`gdk::Paintable.snapshot()` method](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gdk4/prelude/trait.PaintableExt.html#tymethod.snapshot) doesn't accept a reference to a `gtk::Snapshot`. The compiler complains with this:

```
error[E0308]: mismatched types
  --> src/main.rs:74:40
   |
74 |                     paintable.snapshot(&snapshot, width.into(), height.into());
   |                                        ^^^^^^^^^ expected struct `gtk4::gdk4::Snapshot`, found struct `gtk4::Snapshot`
   |
   = note: expected reference `&gtk4::gdk4::Snapshot`
              found reference `&gtk4::Snapshot`

For more information about this error, try `rustc --explain E0308`.
```

It expects a `gdk::Snapshot`, which `gtk::Snapshot` is according to the docs.

I call it with this code:

```rust
let snapshot = gtk::Snapshot::new();
paintable.snapshot(&snapshot, width, height);
```

Replacing `&snapshot` by `snapshot.upcast_ref()` gets rid of the error.

bilelmoussaoui:
Fixes #799


gdk4/src/auto/snapshot.rs
```diff
@@ -13,6 +13,10 @@ glib::wrapper! {
     }
 }
 
+impl Snapshot {
```
bilelmoussaoui, line 16: This is not great, we ideally want a config option to disable generating the const

gdk4/src/subclass/paintable.rs
```diff
@@ -34,7 +34,13 @@ pub trait PaintableImpl: ObjectImpl {
         self.parent_intrinsic_aspect_ratio(paintable)
     }
 
-    fn snapshot(&self, paintable: &Self::Type, snapshot: &Snapshot, width: f64, height: f64);
+    fn snapshot(
+        &self,
+        paintable: &Self::Type,
+        snapshot: &impl IsA<Snapshot>,
```
sdroege, line 40: Does this make any difference actually? It only makes things harder to use or not?

gdk4/src/auto/snapshot.rs
sdroege: Why do you want to disable it? It would be as necessary now as for other non-final types

gtk4/Gir.toml
```diff
@@ -1834,6 +1834,8 @@ generate_builder = false
 [[object]]
 name = "Gtk.Snapshot"
 status = "generate"
+final_type = false
```
sdroege, line 1837: This one is actually a final type or not?

gtk4/src/auto/cell_area.rs
```diff
@@ -230,7 +230,7 @@ pub trait CellAreaExt: 'static {
         &self,
         context: &impl IsA<CellAreaContext>,
         widget: &impl IsA<Widget>,
-        snapshot: &Snapshot,
+        snapshot: &impl IsA<Snapshot>,
```
sdroege, line 233: And this here can only take a `GtkSnapshot`. With any other `GdkSnapshot` it would be wrong

gtk4/src/snapshot.rs
```diff
@@ -15,16 +29,25 @@ impl Snapshot {
...
+    fn push_debug(&self, message: &str) {
+        unsafe {
+            ffi::gtk_snapshot_push_debug(self.as_ref().to_glib_none().0, message.to_glib_none().0)
+        }
     }
 }
+
+impl AsRef<Snapshot> for gdk::Snapshot {
+    fn as_ref(&self) -> &Snapshot {
+        self.downcast_ref().unwrap()
```
sdroege, line 49: This is technically wrong. `GdkSnapshot` is actually a different type than `GtkSnapshot`

sdroege:
I think this is all going in the wrong direction. What I missed is that all the GTK API actually wants a `GtkSnapshot*`, and because both are actually different types I don't see how we can really improve anything here other than allowing `IsA<gdk::Snapshot>` for `gdk::Paintable::snapshot()`

gtk4/Gir.toml
bilelmoussaoui: It is https://docs.gtk.org/gtk4/class.Snapshot.html, but it is a workaround to get the functions of a GtkSnapshot generated in a SnapshotExt trait. Will document that

sdroege: Why do you need them in a `SnapshotExt` trait though? You can only call the `gtk_snapshot*` functions on `GtkSnapshot`, not on `GdkSnapshot`.

bilelmoussaoui:
> I think this is all going in the wrong direction. What I missed is that all the GTK API actually wants a `GtkSnapshot*`, and because both are actually different types I don't see how we can really improve anything here other than allowing `IsA<gdk::Snapshot>` for `gdk::Paintable::snapshot()`

Well GtkSnapshot exists just because gsk can't be linked against gdk. The methods were then moved to a sub-class in gtk which makes sense to be able to call GtkSnapshot methods  on a GdkSnapshot because internally Gtk only cares about GtkSnapshot. That is why everytime you make use of GdkSnapshot (like in Paintable) you downcast it to a GtkSnapshot and the current changes makes it seamless as you just call `snapshot.as_ref().some_render_node_method()`

gtk4/src/snapshot.rs
bilelmoussaoui: it is the parent type of GtkSnapshot https://docs.gtk.org/gtk4/class.Snapshot.html#hierarchy

gtk4/src/auto/cell_area.rs
bilelmoussaoui: Well, there can't be any other kind of GdkSnapshot other than GtkSnapshot. This is basically caused by moving GtkSnapshot methods to a trait, can't do much about it unless we manually implement everything which is alot :)

gdk4/src/subclass/paintable.rs
bilelmoussaoui: See the paintable demo changes, you don't have to downcast and you can go through the AsRef trait

sdroege:
Your call but this breaks the type hierarchy and will probably cause new problems later that we're not aware of yet.

The only problem that I see right now is if the GTK people decide to add a second `GdkSnapshot` subclass. Then this would all fall apart with runtime assertions or worse.

gtk4/src/snapshot.rs
sdroege: Yes, i.e. not every `GdkSnapshot` is a `GtkSnapshot` so the above can fail. Just that **for now** there's only a single subclass of `GdkSnapshot`.

gtk4/src/auto/cell_area.rs
sdroege: > Well, there can't be any other kind of GdkSnapshot other than GtkSnapshot.

Until the GTK people decide to add one and all this falls apart, yes :)

gdk4/src/subclass/paintable.rs
sdroege: But if you made this an `&gdk::Snapshot` then you wouldn't even have to call `as_ref()` and could do exactly the same things with your other changes.

bilelmoussaoui:
I don't see that happening any time soon but sure. I don't really see a way to make the API nicer to use other than what I suggested above.

What do you think would be a more correct thing to implement here?

sdroege:
I don't know. I don't see a correct way to implement this based on what happened in C, but yours will probably work as long as nobody gets the idea to implement a new `GdkSnapshot` subclass.

bilelmoussaoui:
I guess let us land this and figure out something if that ever happens in the future.

sdroege:
Works for me, but very ugly :)

gtk4/src/snapshot.rs
```diff
@@ -15,16 +29,25 @@ impl Snapshot {
...
+    }
+}
+
+unsafe impl IsA<Snapshot> for gdk::Snapshot {}
```
sdroege, line 53: Will also need `Into` and `Borrow` soon

gtk4/src/snapshot.rs
sdroege, line 49: This is technically wrong but let's go with that for now... on the result, `res.type_().is_a(gtk::Snapshot)` would fail which seems unexpected and probably can cause assertions in some places.

sdroege (commented):
Ok with me, I guess

gtk4/src/snapshot.rs
bilelmoussaoui: yes, if the assumption we are doing here changes than we have a lot to fix :) so let us hope it won't happen until gtk5.

sdroege (approved):

PR #807, opened 2022-01-11
regenerate with latest gir/gir-files

bilelmoussaoui:


gdk4/sys/src/lib.rs
```diff
@@ -4508,7 +4508,7 @@ extern "C" {
         height: c_int,
         destroy: glib::GDestroyNotify,
         data: gpointer,
-    ) -> *mut GdkTexture;
+    ) -> *mut GdkGLTexture;
```
sdroege, line 4511: You can remove the `unsafe_cast()` and `Texture::` here now: https://github.com/gtk-rs/gtk4-rs/blob/9efa10698cc5ffc9ae0115238c798735652d4bf8/gdk4/src/gl_texture.rs#L18

sdroege: Also https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/4365 might be of interest

sdroege (approved):

sdroege (approved):

PR #808, opened 2022-01-11
 misc: drop unneeded Deref impls 

### LINKED ISSUE #806: Drop unwanted Deref impl for Events/RenderNodes/Expressions
see 
- https://github.com/gtk-rs/gtk4-rs/blob/master/gdk4/src/event.rs#L135-L141
- https://github.com/gtk-rs/gtk4-rs/blob/master/gtk4/src/expression.rs#L236-L242
- https://github.com/gtk-rs/gtk4-rs/blob/master/gsk4/src/render_node.rs#L114-L120

bilelmoussaoui:
fixes #806

### DIFF
--- a/gdk4/src/event.rs
+++ b/gdk4/src/event.rs
@@ -132,14 +132,6 @@ macro_rules! define_event {
             }
         }
 
-        impl std::ops::Deref for $rust_type {
-            type Target = crate::Event;
-
-            fn deref(&self) -> &Self::Target {
-                unsafe { &*(self as *const $rust_type as *const crate::Event) }
-            }
-        }
-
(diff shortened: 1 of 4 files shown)


sdroege:
These all seem like useful `Deref` impls: you can directly call into the (safe!) base class API.

bilelmoussaoui:
Right, does glib::wrapper generates those for child -> parent cases?

sdroege:
No because `GObject`s can have multiple parents (base class and interfaces), and that's why we have the `IsA` trait there.

bilelmoussaoui:
Makes sense; will go ahead and close this one then

PR #810, opened 2022-01-13
CI: don't build gtk if it is cached already

bilelmoussaoui:


.github/workflows/windows-msvc.yml
```diff
@@ -15,6 +15,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: actions/cache@v1
+      id: cache
```
sdroege, line 18: How do we update the cache if a newer gtk is needed?

bilelmoussaoui: by changing the cache key to e.g gtk4-8

PR #813, opened 2022-01-13
book: Fix error message for `activate_action`

Hofer-Julian:
@bilelmoussaoui  can [this](https://gtk-rs.org/gtk4-rs/git/docs/gtk4/prelude/trait.WidgetExt.html#returns-1) part or the whole docs be overriden?

### DIFF
--- a/book/listings/actions/3/main.rs
+++ b/book/listings/actions/3/main.rs
@@ -69,7 +69,7 @@ fn build_ui(app: &Application) {
         let parameter = 1;
         button
             .activate_action("win.count", Some(&parameter.to_variant()))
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
> @bilelmoussaoui can [this](https://gtk-rs.org/gtk4-rs/git/docs/gtk4/prelude/trait.WidgetExt.html#returns-1) part or the whole docs be overriden?

Could be, but nothing that is doable right now

bilelmoussaoui:
we have the same issue in a lot of other places, could use some gir config, open an issue there maybe?

PR #816, opened 2022-01-13
CI: use rust from the image

bilelmoussaoui:
Should save us a couple of seconds i guess


bilelmoussaoui:
> But that means that we won't get latest beta/nightly anymore. For gstreamer-rs I'm doing the same but I have a scheduled run every 3 days that updates the nightly version.

We could do the same here as well

bilelmoussaoui:
I gave up on this, too much work on the image side. Will try if it is worth it again in the future

PR #824, opened 2022-01-16
gtk4/{Text,Tree}Iter: derive Debug 

fengalin:
So that a `struct` using these types can also auto derive `Debug`.

Another solution could be to add a `Debug` `impl` in manual code,
but since these are iterators, there's not much we can show.

Other `boxed_inline` candidates in `gtk4-rs` use manual `Debug` `impl` AFAICT.


bilelmoussaoui:
Lgtm, thanks !

PR #830, opened 2022-01-17
Replace `Foo::from_instance(foo)` with `foo.imp()`

sdroege:


gtk4-macros/src/lib.rs
```diff
@@ -221,7 +221,7 @@ pub fn composite_template_derive(input: TokenStream) -> TokenStream {
 ///     }
 ///     #[template_callback]
 ///     pub fn print_both_labels(&self) {
-///         let self_ = imp::MyWidget::from_instance(self);
+///         let self_ = self.imp();
 ///         println!("{} {}", self_.label.label(), self_.button.label().unwrap().as_str());
```
bilelmoussaoui, line 225: Same as the gtk 3 pull request

sdroege: Oops

PR #831, opened 2022-01-18
book: Use closure to connect signal

Hofer-Julian:


book/listings/gobject_signals/1/main.rs
```diff
@@ -1,3 +1,5 @@
+use glib::closure_local;
+use gtk::glib;
```
bilelmoussaoui, line 2:
```suggestion
use gtk::glib::{self, closure_local};
```
maybe?

Hofer-Julian: I like your style equally.
But at the moment I use
```
use gtk::{glib, gio}
use glib::stuff;
use gio::stuff;
```
consistently.

If we change that I'd rather change it in one go.

PR #833, opened 2022-01-18
gtk: Set gtk-rs as initialized after the original `GtkApplication::st…

sdroege:
…artup` has returned

Fixes https://github.com/gtk-rs/gtk4-rs/issues/236


GuillaumeGomez (approved):

YaLTeR:
I guess would be nice to also add a test that tries to do something in startup and either panics or doesn't panic.

sdroege:
If someone wants to write a test for that, go ahead. Testing GTK things is always annoying because of the single-thread requirement and having some kind of display server.

I'd prefer to get a bugfix release out with this instead of blocking on a test.

PR #837, opened 2022-01-19
gdk: implement conversions for Rectangle

### LINKED ISSUE #836: Allow conversion between various Rectangle types
There is a gdk::Rectangle, pango::Rectangle, cairo::Rectangle. We should implement From traits for those in the gtk crate

bilelmoussaoui:
From cairo::Rectangle and pango::Rectangle. gdk doesn't depend on graphene otherwise would have added that one as well

Fixes #836

### DIFF
--- a/gdk4/src/rectangle.rs
+++ b/gdk4/src/rectangle.rs
@@ -54,8 +54,51 @@ impl AsRef<RectangleInt> for Rectangle {
 }
 
 impl From<RectangleInt> for Rectangle {
-    fn from(r: RectangleInt) -> Rectangle {
+    fn from(r: RectangleInt) -> Self {
         skip_assert_initialized!();
         unsafe { *(&r as *const _ as *const _) }
+    }
+}
+
+impl From<Rectangle> for RectangleInt {
+    fn from(r: Rectangle) -> Self {
+        skip_assert_initialized!();
+        unsafe { *(&r as *const _ as *const _) }
+    }
+}
+
+impl From<cairo::Rectangle> for Rectangle {
+    // rustdoc-stripper-ignore-next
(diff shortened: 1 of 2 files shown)


gdk4/src/rectangle.rs
```diff
@@ -59,3 +59,21 @@ impl From<RectangleInt> for Rectangle {
         unsafe { *(&r as *const _ as *const _) }
     }
 }
+
+
+impl From<cairo::Rectangle> for Rectangle {
```
bilelmoussaoui, line 64: This might be a bad idea actually... because converting an f64 to i32 might fail, we should probably at least use try_into and panic with a proper error message

bilelmoussaoui: Or even implement TryFrom

bilelmoussaoui: Actually, never mind

gdk4/src/rectangle.rs
```diff
@@ -59,3 +75,20 @@ impl From<RectangleInt> for Rectangle {
...
+    // rustdoc-stripper-ignore-next
+    /// Note that converting between a [`cairo::Rectangle`] and [`Rectangle`]
+    /// will probably lead to precisison errors. Use cautiously.
+    fn from(r: cairo::Rectangle) -> Self {
+        skip_assert_initialized!();
+        Self::new(r.x as i32, r.y as i32, r.width as i32, r.height as i32)
+    }
+}
+
+impl From<pango::Rectangle> for Rectangle {
```
sdroege, line 89: Add both also in the other direction `impl From<Rectangle> for pango::Rectangle { ... }` etc

sdroege:
Seems good to me otherwise, should backport that

PR #839, opened 2022-01-19
make use of impl T wherever possible

### LINKED ISSUE #838: ClosureExpression::new asks for all three generic types
A closure expression has to be declared like that:

```rust
gtk::ClosureExpression::new::<bool, _, _>(
    &[a_expr],
    closure!(|_: Option<Object>, a: bool| { !a }),
)
```

One with no parameters must be declared like that:

```rust
gtk::ClosureExpression::new::<String, &[gtk::Expression], _>(
    &[],
    closure!(|member: Option<Member>| { member.map(search_string).unwrap_or_default() }),
)
```

It would be great if at most we only had to annotate the `ValueType`, like in `chain_closure`. But if the `closure!` return type could be used for the `ValueType`, that would be even better.

bilelmoussaoui:
Except in ClosureExpression because we can't mix both impl & generic types due to
https://github.com/rust-lang/rust/issues/83701

Tried to fix #838 but ended up cleaning up few things, there is more to be done but that is for another time.


sdroege (approved):
Note that both changes are breaking changes, the de-manual change more likely causes problems (people sometimes don't include the prelude but only the traits they use).

bilelmoussaoui:
> Note that both changes are breaking changes, the de-manual change more likely causes problems (people sometimes don't include the prelude but only the traits they use).

Yeah, well I don't know whether to merge those right now or wait till the 0.5 release

sdroege:
master is what becomes 0.5, we're not backporting everything to the 0.4 branch :) So no need to wait either way.

bilelmoussaoui:
I am still not familiar at all with what could be considered an API break or not, your help is very appreciated on that regard, thanks!

sdroege:
FWIW, the other change is a breaking change because previously you could call the functions like `foo::<A, B, C>()` and now you can't anymore because `impl trait` does not give you a type parameter to the function. I would ignore that for our purposes here though because nobody calls it like that, there's no point in providing the type parameters like this.

The thing with the traits I wouldn't ignore because I know of people very selectively including specific traits because they don't like preludes.

PR #841, opened 2022-01-19
Release 0.4.4

bilelmoussaoui:


sdroege:
Some API changes here but IMHO these are fine.

sdroege (approved):

sdroege:
API changes:
 - Removal of the useless gsk rectangle functions (someone might call them for whatever reason...)
 -  type parameter to `impl trait ` removes the ability to call via turbofish syntax, but nobody should do that for these methods
 - `T` to `Option<T>`, but those were just wrong before and any code that stops compiling now had a timebomb there

bilelmoussaoui:
> API changes:
> 
>     * Removal of the useless gsk rectangle functions (someone might call them for whatever reason...)
> 
>     * type parameter to `impl trait ` removes the ability to call via turbofish syntax, but nobody should do that for these methods
> 
>     * `T` to `Option<T>`, but those were just wrong before and any code that stops compiling now had a timebomb there

Thanks, i will add those to the release notes

PR #850, opened 2022-01-24
Don't require Send closures for GIO-style async functions

bilelmoussaoui:


sdroege (approved):

gdk4/src/clipboard.rs
```diff
@@ -3,7 +3,7 @@
 use crate::{Clipboard, Texture};
```
sdroege, line 3:
```suggestion
use crate::Clipboard;
```

PR #852, opened 2022-01-25
book: Cleanup

Hofer-Julian:


book/src/gobject_subclassing.md
```diff
@@ -3,44 +3,50 @@
...
 {{#rustdoc_include ../listings/gobject_subclassing/1/custom_button/mod.rs:mod}}
 ```
 
-[`glib::wrapper!`](http://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/macro.wrapper.html) does the most of the work of subclassing for us.
-Coming from most other languages you would probably expect that you only have to mention the base class you want to inherit from.
-However, as of today, subclassing of GObjects in Rust requires to mention all ancestors and interfaces apart from `GObject` and `GInitiallyUnowned`.
-For `gtk::Button`, we can look them up in the corresponding [doc page](https://docs.gtk.org/gtk4/class.Button.html#hierarchy) of GTK4.
+[`glib::wrapper!`](http://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/macro.wrapper.html) implements the same traits that our `ParentType` implements.
+Theoretically that would mean that we `ParentType` is also the only thing we have to specify here.
+Unfortunately, nobody found yet a good way to do that.
```
bilelmoussaoui, line 31: maybe it is worth mentioning that the list of types you provide for `@extends` / `@implements` is used to implement various traits for you like the IsA. That is why not mentioning everything is fine, until you need some specific methods from the `IsA<T>` implementation or you have to pass the new type to a method that takes `IsA<T>`.

Hofer-Julian: I did mention that you don't have to specify everything, so I assume you want me to be more precise here.
Correct?

PR #855, opened 2022-01-28
Trust return value nullability again for gsk

sdroege:


gsk4/src/auto/transform.rs
```diff
@@ -60,7 +60,7 @@ impl Transform {
 
     #[doc(alias = "gsk_transform_perspective")]
     #[must_use]
-    pub fn perspective(&self, depth: f32) -> Option<Transform> {
+    pub fn perspective(&self, depth: f32) -> Transform {
```
sdroege, line 63: This is correct and can't return NULL

gsk4/src/auto/transform.rs
```diff
@@ -49,7 +49,7 @@ impl Transform {
 
     #[doc(alias = "gsk_transform_matrix")]
     #[must_use]
-    pub fn matrix(&self, matrix: &graphene::Matrix) -> Option<Transform> {
+    pub fn matrix(&self, matrix: &graphene::Matrix) -> Transform {
```
sdroege, line 52: This is correct and can't return NULL

sdroege:
@bilelmoussaoui Would you want to backport this or not?

bilelmoussaoui:
> @bilelmoussaoui Would you want to backport this or not?

sure that sounds good to me

PR #864, opened 2022-02-04
Gir files update

sdroege:


sdroege:
@bilelmoussaoui Needs a newer GTK version on the CI

sdroege:
How do you update GTK? Does it just need some kind of image rebuild?

bilelmoussaoui:
> How do you update GTK? Does it just need some kind of image rebuild?

Oh sorry I completely missed this one, I will update the image. Pushing changes to the Dockerfile updates the image

gdk4/sys/src/lib.rs
```diff
@@ -96,7 +96,8 @@ pub const GDK_PAD_BUTTON_RELEASE: GdkEventType = 24;
 pub const GDK_PAD_RING: GdkEventType = 25;
 pub const GDK_PAD_STRIP: GdkEventType = 26;
 pub const GDK_PAD_GROUP_MODE: GdkEventType = 27;
-pub const GDK_EVENT_LAST: GdkEventType = 28;
+pub const GDK_TOUCHPAD_HOLD: GdkEventType = 28;
+pub const GDK_EVENT_LAST: GdkEventType = 29;
```
bilelmoussaoui, line 100: This should probably be ignored no?

gdk4/src/auto/enums.rs
```diff
@@ -621,6 +621,8 @@ pub enum EventType {
     PadStrip,
     #[doc(alias = "GDK_PAD_GROUP_MODE")]
     PadGroupMode,
+    #[doc(alias = "GDK_TOUCHPAD_HOLD")]
+    TouchpadHold,
```
bilelmoussaoui, line 625: It probably needs a version guard

gdk4/sys/src/lib.rs
sdroege: Probably, but you don't ignore any of these elsewhere either in gtk4-rs from what I can see.

sdroege: Well, should be hidden in the non-sys bindings. In the -sys bindings it can be useful in the same way as in C.

gdk4/src/auto/enums.rs
sdroege: Which version though? 4.6.1? 4.8?

gdk4/sys/src/lib.rs
bilelmoussaoui: We do in GdkMemoryFormat

gdk4/src/auto/enums.rs
bilelmoussaoui: 4.8 i guess

bilelmoussaoui: It also needs to be added to https://github.com/gtk-rs/gtk4-rs/blob/master/gdk4/src/touchpad_event.rs#L9

sdroege: Conditionally though. I'll try :)

bilelmoussaoui:
The image should be updated now, I will do similar updates in the future for gtk-rs-core as part of cleaning up the base image situation

sdroege:
Ok, it builds. I'll fix the other things later :)

sdroege:
Should be all good now

bilelmoussaoui:
Thanks!

PR #866, opened 2022-02-04
Fixed path to example image

ThomasFrans:
Previous path didn't work if the working directory was the examples folder.

### DIFF
--- a/examples/clipboard/main.rs
+++ b/examples/clipboard/main.rs
@@ -87,7 +87,7 @@ fn build_ui(application: &gtk::Application) {
         .spacing(24)
         .build();
 
-    let file = gio::File::for_path("./examples/clipboard/asset.png");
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
It fixes it in this case but not when you open it from the main repository which is a bit unfortunate. Maybe we can use another approach here

Hofer-Julian:
I think only allowing to run it from the examples directory is fine.
(unless of course there is a simple way to make it work from other directories as well)

jf2048:
Other options would be to use `include_bytes` or get the path relative to `std::env::current_exe`

bilelmoussaoui:
Any updates on this one ?

ThomasFrans:
I think it's fine to close this without merging. It would be nice if the working directory didn't matter, but most users probably run examples with cargo run --example from the project root anyways, so I'll just close this one.

PR #869, opened 2022-02-06
Fix library name for epoxy on Windows.

### LINKED ISSUE #854: [BUG] Wrong Filename for Windows in glium_gl_area Example
**Bug description**

This line:

https://github.com/gtk-rs/gtk4-rs/blob/200d25e577727a1d61536533fcdaae8da3ee7d0f/examples/glium_gl_area/main.rs#L17

tries to load a DLL called "libepoxy-0.dll". However, at least on my installation this file is called "epoxy-0.dll".

Without the fix:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: GetModuleHandleExW { source: Os { code: 126, kind: Uncategorized, message: "The specified module could not be found." } }', examples\glium_gl_area/main.rs:17:85
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

with the change (removing the `lib` prefix) it works fine on my machine.

anlumo:
Fixes #854

### DIFF
--- a/examples/glium_gl_area/main.rs
+++ b/examples/glium_gl_area/main.rs
@@ -13,8 +13,7 @@ fn main() {
         #[cfg(all(unix, not(target_os = "macos")))]
         let library = unsafe { libloading::os::unix::Library::new("libepoxy.so.0") }.unwrap();
         #[cfg(windows)]
-        let library =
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
Thanks!

PR #871, opened 2022-02-07
Handle empty slices correctly

sdroege:
Passing `NULL` to `slice::from_raw_parts` is invalid.

### DIFF
--- a/gtk4/src/editable.rs
+++ b/gtk4/src/editable.rs
@@ -45,7 +45,9 @@ unsafe extern "C" fn insert_text_trampoline<T, F: Fn(&T, &str, &mut i32) + 'stat
 ) where
     T: IsA<Editable>,
 {
-    let buf = if new_text_length != -1 {
+    let buf = if new_text_length == 0 {
+        &[]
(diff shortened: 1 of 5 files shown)


sdroege:
@bilelmoussaoui Something broke with the GTK version in the images

bilelmoussaoui:
> @bilelmoussaoui Something broke with the GTK version in the images

It is caused by GDK_EVENT_LAST, see #864

bilelmoussaoui:
If you rebase, the CI should be happy

gtk4/src/editable.rs
```diff
@@ -45,7 +45,9 @@ unsafe extern "C" fn insert_text_trampoline<T, F: Fn(&T, &str, &mut i32) + 'stat
 ) where
     T: IsA<Editable>,
 {
-    let buf = if new_text_length != -1 {
+    let buf = if next_text_length == 0 {
```
bilelmoussaoui, line 48:
```suggestion
    let buf = if new_text_length == 0 {
```

PR #900, opened 2022-02-11
Draft: subclass: Add missing BuildableImpl vfuncs

bilelmoussaoui:


bilelmoussaoui:
I haven't got the time to clean things up, will get back to it pretty soon i hope

gtk4/src/subclass/buildable.rs
```diff
@@ -51,35 +51,36 @@ pub trait BuildableImpl: ObjectImpl {
...
         buildable: &Self::Type,
         builder: &Builder,
         child: Option<&Object>,
-        tagname: &str,
-        parser: BuildableParser,
-        data: ptr,
-    );
-    fn custom_tag_end(
+        tag_name: &str,
+        parser_data: *mut *mut libc::c_void,
```
sdroege, line 60: How is this pointer used exactly?

gtk4/src/subclass/buildable_parser.rs
```diff
@@ -0,0 +1,185 @@
...
+                std::ptr::null_mut(),
+                std::ptr::null_mut(),
+                std::ptr::null_mut(),
+                std::ptr::null_mut(),
+            ],
+        })
+    }
+}
+
+pub unsafe trait BuildableParserImpl: Sized {
```
sdroege, line 77: Why does it have to be unsafe?

gtk4/src/subclass/buildable.rs
bilelmoussaoui: It is used to pass the state to the BuildableParser, my idea was to create a struct that implements a certain trait and move the struct around instead of passing a random ptr around

gtk4/src/subclass/buildable_parser.rs
bilelmoussaoui: It shouldn't, I made it unsafe at first before figuring out what the user_data pointers are used for. Ideally once this PR is ready, we shouldn't need that

gtk4/src/subclass/buildable.rs
sdroege: Sounds like a good plan :) Where would the struct be freed, is the `finish()` function guaranteed to be called exactly once with the same parser data, for example?

Issue #930, opened 2022-02-17
[BUG] Could not compile `gdk4` due to 8 errors


kavanmevada:
**Bug description**

Cargo.toml

```
[dependencies]
adw = { git = "https://gitlab.gnome.org/World/Rust/libadwaita-rs.git", package = "libadwaita" }
gtk = { git = "https://github.com/gtk-rs/gtk4-rs.git", package = "gtk4" }
gdk = { git = "https://github.com/gtk-rs/gtk4-rs.git", package = "gdk4" }
glib = { git = "https://github.com/gtk-rs/gtk-rs-core.git", package = "glib" }
gio = { git = "https://github.com/gtk-rs/gtk-rs-core.git", package = "gio" }
once_cell = "1.9.0"
```

**Backtrace**

```
error[E0615]: attempted to take value of method `x` on type `cairo::Rectangle`
  --> /var/home/kavan/.cargo/git/checkouts/gtk4-rs-e74ad56283dfeb5e/c382c05/gdk4/src/rectangle.rs:92:21
   |
92 |         Self::new(r.x as i32, r.y as i32, r.width as i32, r.height as i32)
   |                     ^ method, not a field
   |
help: use parentheses to call the method
   |
92 |         Self::new(r.x() as i32, r.y as i32, r.width as i32, r.height as i32)
   |                      ++

error[E0615]: attempted to take value of method `y` on type `cairo::Rectangle`
  --> /var/home/kavan/.cargo/git/checkouts/gtk4-rs-e74ad56283dfeb5e/c382c05/gdk4/src/rectangle.rs:92:33
   |
92 |         Self::new(r.x as i32, r.y as i32, r.width as i32, r.height as i32)
   |                                 ^ method, not a field
   |
help: use parentheses to call the method
   |
92 |         Self::new(r.x as i32, r.y() as i32, r.width as i32, r.height as i32)
   |                                  ++

error[E0615]: attempted to take value of method `width` on type `cairo::Rectangle`
  --> /var/home/kavan/.cargo/git/checkouts/gtk4-rs-e74ad56283dfeb5e/c382c05/gdk4/src/rectangle.rs:92:45
   |
92 |         Self::new(r.x as i32, r.y as i32, r.width as i32, r.height as i32)
   |                                             ^^^^^ method, not a field
   |
help: use parentheses to call the method
   |
92 |         Self::new(r.x as i32, r.y as i32, r.width() as i32, r.height as i32)
   |                                                  ++

[...]

PR #931, opened 2022-02-17
gdk: Fix compilation after `cairo::Rectangle` API changes

sdroege:
Fixes https://github.com/gtk-rs/gtk4-rs/issues/930


sdroege:
merging because it's a trivial change and the build is broken otherwise

PR #937, opened 2022-02-18
0.4 backports

sdroege:


sdroege:
@Hofer-Julian And I leave backporting book things to you

bilelmoussaoui:
> Maybe something to look into in the future :)

Do you even need to do that? you can just trigger a book job manually and it will publish the output to stable instead of git :)

Any reason why the nullablity changes were not backported as well?

sdroege:
> Any reason why the nullablity changes were not backported as well?

What did I miss exactly?

sdroege:
Ah, because you had actual manual changes inside the regenerate commit. That's not ideal because different gir versions are used for 0.15 and master. I skipped all the regenerate commits and did new ones. Please don't put manual changes into regenerate commits in the future :)

I'll manually extract those changes in a few minutes.

bilelmoussaoui:
> Ah, because you had actual manual changes inside the regenerate commit. That's not ideal because different gir versions are used for 0.15 and master. I skipped all the regenerate commits and did new ones. Please don't put manual changes into regenerate commits in the future :)
> 
> I'll manually extract those changes in a few minutes.

I don't see any manual changes there & I didn't update gir neither

Hofer-Julian:
> > Maybe something to look into in the future :)
> 
> Do you even need to do that? you can just trigger a book job manually and it will publish the output to stable instead of git :)
> 
> Any reason why the nullablity changes were not backported as well?

Sure I do that all the time.
But if you want to be super fancy you could maintain a stable branch for the book or even a branch per release. (not that I plan to work on that though)

sdroege:
> I don't see any manual changes there & I didn't update gir neither

@bilelmoussaoui You updated `gir-files` in the same commit, that's the "manual" change (in contrast to "running generator.py"). I've added that as a separate commit here now and then regenerated.

The gir versions in 0.15 and master are different, so a full regenerate commit can't be cherry-picked as the commit hash of gir is going to be different so there will be a merge conflict.

The changes from `generator.py` should always be a separate commit, at least if they involve a `gir-files` or `gir` update.

bilelmoussaoui:
Oh I see, I always updated the submodules & ran the generator.py script on the same commit. I will avoid that in the future

sdroege:
So what do we do here? :)

bilelmoussaoui (approved):

Issue #951, opened 2022-02-24
[FEATURE REQUEST] Add basic testing infrastructure and documentation


AaronErhardt:
There should be some testing infrastructure for gtk4-rs apps and internal code that's reusable and documented.

At least something as a convenient entry point for running tests that rely on GTK (something similar to https://github.com/gtk-rs/gtk4-rs/blob/master/gtk4/src/lib.rs#L62).

I've seen https://github.com/gtk-rs/gtk-test/issues/47 and to be honest, I don't have a lot of resources currently, too, but I hope we can at least work together to enable some basic functionality for the upcoming release.

bilelmoussaoui:
cc @jf2048 as you wanted the test_synced to be public as well

bilelmoussaoui:
I am not sure what would be needed other than making that function available + some examples. I wonder if the `gtk_test_init` would be any useful for this case?

jf2048:
Yes, it seems like that needs to be public (or it needs to be copied) if someone wants unit tests for any gtk widgets, or for gtk4-macros, libadwaita-rs, libshumate-rs, libpanel-rs, sourceview5-rs, etc. Seeing a gtk4-test would also be good and probably not too hard

jf2048:
This issue is still a problem though: #811

Although maybe it's not that bad in practice since the tests all run synchronously anyway?

PR #952, opened 2022-02-26
Add a #[gtk::test] attribute for unit tests

jf2048:
Related to #951. Some notes:

- This allows us to remove `test_synced` from the internal unit tests
  - Downside is all the internal tests (inside `#[cfg(test)] mod tests { }` blocks) must have `use crate as gtk4` at the top to avoid issues with proc_macro_crate. Not an issue externally or for tests in the tests folder
  - Makes it possible to have a unit test run correctly in the gtk4-macros crate
- Was trivial to add support for async functions as unit tests
- The macro shadows the prelude `test`, so internally we have to use `#[std::prelude::v1::test]` for the few tests that don't need to run on the main thread. Not an issue externally unless someone uses `macro_use`
- Maybe would have been nicer to put this in another crate, but couldn't figure out how to do that without creating a circular dependency on gtk4

### DIFF
--- a/gtk4-macros/src/lib.rs
+++ b/gtk4-macros/src/lib.rs
@@ -261,3 +261,67 @@ pub fn template_callbacks(attr: TokenStream, item: TokenStream) -> TokenStream {
         Err(_) => abort_call_site!(template_callbacks_attribute::WRONG_PLACE_MSG),
     }
 }
+
+/// Attribute macro for declaring GTK tests.
+///
+/// Wraps the standard Rust [`test`] attribute with setup logic for GTK. All tests that call
+/// into GTK must use this attribute. This attribute can also be used on asynchronous functions;
+/// the asynchronous test will be run on the main thread context.
+///
+/// # Technical Details
+///
+/// GTK is a single-threaded library, so Rust's normal multi-threaded test behavior cannot be used.
+/// The `#[gtk::test]` attribute creates a main thread for GTK and runs all tests on that thread.
+/// This has the side effect of making all tests run serially, not in parallel.
+///
+/// [`test`]: <https://doc.rust-lang.org/std/prelude/v1/macro.test.html>
+///
+/// # Example
+///
+/// ```no_run
+/// use gtk::prelude::*;
+///
+/// #[gtk::test]
+/// fn test_button() {
+///     let button = gtk::Button::new();
+///     button.activate();
+/// }
+/// ```
+#[proc_macro_attribute]
+#[proc_macro_error]
+pub fn test(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    use proc_macro_error::abort_call_site;
+    use quote::quote;
+
+    match syn::parse::<syn::ItemFn>(item) {
+        Ok(mut input) => {
+            let crate_ident = util::crate_ident_new();
+            let block = &input.block;
+            let block = if input.sig.asyncness.is_some() {
+                quote! {
+                    #crate_ident::glib::MainContext::default().block_on(async move {
+                        #block
+                    })
+                }
+            } else {
+                quote! { #block }
+            };
+            input.sig.asyncness.take();
+
+            let attrs = &input.attrs;
+            let vis = &input.vis;
+            let sig = &input.sig;
+            quote! {
+                #(#attrs)*
+                #[::std::prelude::v1::test]
+                #vis #sig {
+                    #crate_ident::test_synced(move || {
+                        #block
+                    })
+                }
+            }
+            .into()
+        }
+        Err(_) => abort_call_site!("This macro should be used on a function definition"),
+    }
+}
--- a/gtk4/Cargo.toml
+++ b/gtk4/Cargo.toml
@@ -42,5 +42,4 @@ once_cell = "1.0"
 pango = {git = "https://github.com/gtk-rs/gtk-rs-core", version = "0.16", features = ["v1_46"]}
 
 [dev-dependencies]
-futures-executor = "0.3"
 gir-format-check = "^0.1"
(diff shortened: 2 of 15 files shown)


gtk4-macros/src/lib.rs
```diff
@@ -261,3 +261,69 @@ pub fn template_callbacks(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// into GTK must use this attribute. This attribute can also be used on asynchronous functions;
+/// the asynchronous test will be run on the main thread context.
+///
+/// # Technical Details
+///
+/// GTK is a single-threaded library, so Rust's normal multi-threaded test behavior cannot be used.
+/// The `#[gtk::test]` attribute creates a main thread for GTK and runs all tests on that thread.
+/// This has the side effect of making all tests run serially, not in parallel.
+///
+/// [`#[test]`]: (https://doc.rust-lang.org/std/prelude/v1/macro.test.html)
```
bilelmoussaoui, line 277:
```suggestion
/// [`#[test]`]: <https://doc.rust-lang.org/std/prelude/v1/macro.test.html>
```

gtk4/Cargo.toml
```diff
@@ -30,6 +30,7 @@ cairo-rs = {git = "https://github.com/gtk-rs/gtk-rs-core", version = "0.16"}
 ffi = {package = "gtk4-sys", path = "./sys", version = "0.5.0"}
 field-offset = "0.3"
 futures-channel = "0.3"
+futures-executor = "0.3"
```
sdroege, line 33: Where is that actually used?

gtk4-macros/src/lib.rs
```diff
@@ -261,3 +261,69 @@ pub fn template_callbacks(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+///     button.activate();
+/// }
+/// ```
+#[proc_macro_attribute]
+#[proc_macro_error]
+pub fn test(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    use proc_macro_error::abort_call_site;
+    use quote::quote;
+
+    match syn::parse::<syn::ItemFn>(item) {
```
sdroege, line 296: Might want to add some custom error if the function takes parameters or returns something

sdroege:
Looks useful to me!

gtk4-macros/src/lib.rs
jf2048: That is all handled already by `#[test]`, we can just fall back on that logic i.e. the handling of return values that implement `std::process::Termination`.

gtk4/Cargo.toml
jf2048: It's used to block on a channel:
https://github.com/gtk-rs/gtk4-rs/blob/fa3999913742fd38f2284033a445223a13edb0f1/gtk4/src/lib.rs#L79
I guess this could be rewritten to emulate it using a `glib::MainLoop` and `idle_add`?

gtk4-macros/src/lib.rs
```diff
@@ -261,3 +261,69 @@ pub fn template_callbacks(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// This has the side effect of making all tests run serially, not in parallel.
+///
+/// [`#[test]`]: (https://doc.rust-lang.org/std/prelude/v1/macro.test.html)
+///
+/// # Example
+///
+/// ```no_run
+/// use gtk::prelude::*;
+///
+/// #[gtk::test]
```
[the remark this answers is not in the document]
jf2048: It's done lazily inside the test worker thread:
https://github.com/gtk-rs/gtk4-rs/blob/fa3999913742fd38f2284033a445223a13edb0f1/gtk4/src/lib.rs#L88

gtk4-macros/src/lib.rs
```diff
@@ -261,3 +261,69 @@ pub fn template_callbacks(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+    use proc_macro_error::abort_call_site;
+    use quote::quote;
+
+    match syn::parse::<syn::ItemFn>(item) {
+        Ok(mut input) => {
+            let crate_ident = util::crate_ident_new();
+            let block = &input.block;
+            let block = if input.sig.asyncness.is_some() {
+                quote! {
+                    #crate_ident::glib::MainContext::ref_thread_default().block_on(async move {
```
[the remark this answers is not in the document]
jf2048: It's not recursive because this is blocking inside the worker thread whereas the other block happens on the thread spawned by the test runner, I suppose we should be creating a new main context here to make it work though.

gtk4/Cargo.toml
sdroege: You could create a new main context and `block_on()` with it, that would be equivalent. Or you use a `std::sync::mpsc` channel and just block on the receiver directly without all the async complications.

gtk4-macros/src/lib.rs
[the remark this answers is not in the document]
sdroege: Yes those would have to be different main contexts. Why does this not panic though, that seems problematic :)

gtk4/Cargo.toml
jf2048: Changed it to `std::sync::mpsc`

gtk4-macros/src/lib.rs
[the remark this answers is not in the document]
jf2048: Fixed it, and added an example of an async test for `Dialog::run_future` to check that it works.

jf2048:
@AaronErhardt Is this looking good enough for unit tests?

bilelmoussaoui:
Should this PR close #951 ?

AaronErhardt:
I think we can close #951 because "basic" testing should be fulfilled IMO. Of course porting gtk-test would still be nice but that's an issue of its own.

bilelmoussaoui:
For sure, I don't think it makes sense to port gtk-test as is. The current design of gtk-test is add more stuff until everyone is happy which is something that I don't want to do for gtk4-rs. Ideally, there should be some list of very common things that people do test and contains a bunch of boilerplate code and from there figure out a good way to tackle the issue rather than starting with code.

sdroege (approved):

PR #953, opened 2022-03-01
Let book depend on latest gtk4-rs release

Hofer-Julian:
This was done since we can't depend on:
-  the local gtk4-rs, as well as
- libadwaita depending on a different gtk4-rs

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -57,16 +57,6 @@ jobs:
           keep_files: false
           destination_dir: stable/latest/book
 
-      - name: Deploy to git
-        uses: peaceiris/actions-gh-pages@v3
-        if: ${{ (github.event_name == 'push' && github.ref == 'refs/heads/master') || github.event_name == 'release' }}
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_dir: ./book/public
-          keep_files: false
-          destination_dir: git/book
-
-
   codespell:
     runs-on: ubuntu-20.04
     steps:
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -48,7 +48,7 @@ jobs:
 
       - run: mdbook build -d public
         working-directory: book
-      - run: echo "DEST=$(if [ "$GITHUB_EVENT_NAME" == "release" ] || [ "$GITHUB_EVENT_NAME" == "workflow_dispatch" ]; then echo 'stable/latest'; else echo 'git'; fi)" >> ${GITHUB_ENV}
+      - run: echo "DEST=$(if [ "$GITHUB_EVENT_NAME" == "workflow_dispatch" ]; then echo 'git'; else echo 'stable/latest'; fi)" >> ${GITHUB_ENV}
 
       - name: Deploy
         uses: peaceiris/actions-gh-pages@v3
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -7,7 +7,6 @@ on:
       - "master"
   release:
     types: [published]
-  workflow_dispatch:
 
 jobs:
   test:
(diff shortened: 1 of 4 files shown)


Hofer-Julian:
@bilelmoussaoui the book now depends on the latest release of gtk4-rs.
Depending how much time you have on your hands, we could merge this PR with libadwaita included in the container.
Alternatively, I can remove the libadwaita dependency for now and we can handle this later.

(sorry for the big diff on .github/workflows/windows-msvc.yml, my formatter went overboard there. I just added the book to it)

Hofer-Julian:
> Depending how much time you have on your hands, we could merge this PR with libadwaita included in the container.

@bilelmoussaoui I've added libadwaita to dockerfile and windows CI now.
I think we can push the new container from within this PR, but I thought you probably want to have a look first.

Dockerfile
```diff
@@ -11,7 +11,13 @@ RUN dnf builddep gtk4 -y
 RUN git clone https://gitlab.gnome.org/gnOME/gtk.git --depth=1
 WORKDIR gtk
 RUN meson _build --prefix=/usr -Dgtk_doc=false -Dintrospection=disabled -Dbuild-examples=false -Dbuild-tests=false -Ddemos=false
-RUN ninja -C _build
-RUN ninja -C _build install
+RUN meson install -C _build
 WORKDIR /
 RUN rm -rf gtk
+
+RUN git clone https://gitlab.gnome.org/GNOME/libadwaita.git --depth=1
```
bilelmoussaoui, line 18: I would use latest stable release here

.github/workflows/windows-msvc.yml
```diff
@@ -14,81 +14,106 @@ jobs:
...
+        with:
+          command: clippy
+          args: --features v4_6
+      - name: Tests (workspace)
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --features v4_6
+
+      - name: "Build (book)"
```
bilelmoussaoui, line 113: Does it matter to build the book under windows?

.github/workflows/book.yml
```diff
@@ -48,11 +47,11 @@ jobs:
 
       - run: mdbook build -d public
         working-directory: book
-      - run: echo "DEST=$(if [ "$GITHUB_EVENT_NAME" == "release" ] || [ "$GITHUB_EVENT_NAME" == "workflow_dispatch" ]; then echo 'stable/latest'; else echo 'git'; fi)" >> ${GITHUB_ENV}
+      - run: echo "DEST='stable/latest'" >> ${GITHUB_ENV}
```
bilelmoussaoui, line 50: I suppose this change is not intended? or do you want to drop the git variant of the book?

Hofer-Julian: Yes. I see no reason to have a git variant of the book that depends on stable gtk4-rs

.github/workflows/windows-msvc.yml
Hofer-Julian: Mmmh fair point, we probably only verify that way that libadwaita continues to build on Windows

.github/workflows/book.yml
bilelmoussaoui: Well the git variant could be wip chapters and things like that. If that is the decision, the commit should also drop the links to the git version of the book and probably do some re-direction between old (git) links to stable/latest

.github/workflows/windows-msvc.yml
bilelmoussaoui: That is something up to libadwaita/libadwaita-rs to fix

.github/workflows/book.yml
Hofer-Julian: > Well the git variant could be wip chapters 

I'd like the option to give reviewers a rendered version of a PR. I'll experiment if that can be done with a workflow dispatch.

.github/workflows/windows-msvc.yml
Hofer-Julian: I've removed it

.github/workflows/book.yml
Hofer-Julian: The current behaviour is:
- stable: used for every release or push to master
- git: used when using workflow dispatch

.github/workflows/book.yml
```diff
@@ -48,7 +48,7 @@ jobs:
 
       - run: mdbook build -d public
         working-directory: book
-      - run: echo "DEST=$(if [ "$GITHUB_EVENT_NAME" == "release" ] || [ "$GITHUB_EVENT_NAME" == "workflow_dispatch" ]; then echo 'stable/latest'; else echo 'git'; fi)" >> ${GITHUB_ENV}
+      - run: echo "DEST=$(if [ "$GITHUB_EVENT_NAME" == "workflow_dispatch" ]; then echo 'git'; else echo 'stable/latest'; fi)" >> ${GITHUB_ENV}
```
bilelmoussaoui, line 51: I think you meant the opposite here, now every commit will be published to latest/stable and only the workflow_dispatch (manually triggered) gets published to git which is weird....

Hofer-Julian: I do want that every commit gets pushed to stable from now on. I do depend on stable after all.

It is true though, that we can get rid of workflow_dispatch for now

PR #956, opened 2022-03-04
gtk4-macros: Improve error reporting, add failure tests

jf2048:
The error messages here are quite spammy and fail the whole code generation step for non-fatal errors like duplicate attributes. So let's reduce the error reporting to only emit and then continue on, e.g. just generate an empty clause in the trait so as not to trigger any additional errors. This only works on nightly for now but will "just work" if and when `Diagnostic` gets stabilized because of proc_macro_error.

@GuillaumeGomez Could this be maybe improved in any way with trybuild2?

### DIFF
--- a/gtk4-macros/tests/template-string.rs
+++ b/gtk4-macros/tests/template-string.rs
@@ -35,7 +35,13 @@ mod imp {
         }
     }
 
-    impl ObjectImpl for MyWidget {}
+    impl ObjectImpl for MyWidget {
+        fn dispose(&self, obj: &Self::Type) {
+            if let Some(child) = obj.first_child() {
+                child.unparent();
+            }
+        }
+    }
     impl WidgetImpl for MyWidget {}
 }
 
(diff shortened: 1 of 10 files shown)


GuillaumeGomez:
I'll take a look tomorrow.

jf2048:
Sadly the improved error reporting only works on nightly, but at least we can get some more unit tests here

GuillaumeGomez:
For the errors, I can extend `trybuild2` to allow substring matching on "normal" tests as well.

jf2048:
May not be necessary, I think `compile_fail_check_sub` is enough here?

GuillaumeGomez:
Should be at least.

gtk4-macros/tests/template-string.rs
```diff
@@ -35,7 +35,11 @@ mod imp {
         }
     }
 
-    impl ObjectImpl for MyWidget {}
+    impl ObjectImpl for MyWidget {
+        fn dispose(&self, obj: &Self::Type) {
+            obj.first_child().unwrap().unparent();
```
bilelmoussaoui, line 40: dispose could potentially be called multiple times, so I would rather have something like
```rust
if let Some(child) = obj.first_child() {
  child.unparent();
}
```

bilelmoussaoui (approved):
The compile tests are really appreciated! thanks

GuillaumeGomez (approved):
Tests and macros look good to me, thanks! Just remains the comment from @bilelmoussaoui and it seems good to go!

PR #959, opened 2022-03-06
gtk4-macros: Allow async template callbacks

jf2048:
Also means we now support using by-value `self` on template callbacks. This cannot work for the private `imp` struct because it's not reference counted, it only works for the wrapper.


sdroege:
The documentation should probably make it very clear that even if there still is an async task of the previous signal emission pending that the signal can be emitted again and a new async task would be scheduled in addition.

> This cannot work for the private `imp` struct because it's not reference counted, it only works for the wrapper.

You can go from that to the wrapper, clone that into the async block, and then get the imp struct out of that again.

jf2048:
Changed it so it moves all the arguments as a `Vec<Value>` into the future. Then we can get references again. Also updated the docs.

sdroege (approved):

bilelmoussaoui (approved):

Issue #179, opened 2021-01-27
Forgetting to add #[template_child] to TemplateChild members leads to runtime crashes when accessing the member


Cogitri:
When I have a struct like:

```rs
    #[derive(Debug, CompositeTemplate)]
    #[template(resource = "/dev/Cogitri/test/ui/my_window.ui")]
    pub struct MyWindow {
        #[template_child]
        pub add_data_button: TemplateChild<gtk::Button>,
    }
```

And init it like so:

```rs
        fn new() -> Self {
            Self {
                add_data_button: TemplateChild::default(),
            }
       }
```

I don't get any error message during compilation or when creating the object. However, once I access `add_data_button`, the app crashes:

```rs
thread 'main' panicked at 'assertion failed: !self.ptr.is_null()', /var/home/rasmus/Projects/Jinsei/build/target/cargo-home/git/checkouts/gtk4-rs-e74ad56283dfeb5e/64d0e1c/gtk4/src/subclass/widget.rs:1000:13
stack backtrace:
   0: std::panicking::begin_panic
             at /var/home/rasmus/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panicking.rs:505
   1: <gtk4::subclass::widget::TemplateChild<T> as core::ops::deref::Deref>::deref
             at ./build/target/cargo-home/git/checkouts/gtk4-rs-e74ad56283dfeb5e/64d0e1c/gtk4/src/subclass/widget.rs:1000
   2: health::windows::window::imp::HealthWindow::connect_handlers
             at ./src/windows/window.rs:158
   3: <health::windows::window::imp::HealthWindow as glib::subclass::object::ObjectImpl>::constructed
             at ./src/windows/window.rs:129
   4: glib::subclass::object::constructed
             at ./build/target/cargo-home/git/checkouts/gtk-rs-48ef14c1f17c79fb/c58fab4/glib/src/subclass/object.rs:109
   5: <unknown>
   6: g_object_newv
   7: glib::object::Object::new_internal
             at ./build/target/cargo-home/git/checkouts/gtk-rs-48ef14c1f17c79fb/c58fab4/glib/src/object.rs:1165
   8: glib::object::Object::with_type
             at ./build/target/cargo-home/git/checkouts/gtk-rs-48ef14c1f17c79fb/c58fab4/glib/src/object.rs:1105
   9: glib::object::Object::new
[...]

PR #961, opened 2022-03-06
More checks for composite template children

jf2048:
To finish out the last part of https://github.com/gtk-rs/gtk4-rs/issues/179#issuecomment-1046256441:

- Try to parse the XML inside the macro and check for the existence of the ids. For now this only works for `#[template(string = "...")]` but could support `file = "..."` eventually if `proc_macro_span` is stabilized. Not sure if there is ever going to be a way to support `resource = "..."`
- Check for fields with type `TemplateChild<...>` and emit a warning if the `#[template_child]` attribute is missing. I don't want to make this an error because it's only really a heuristic for the common case. It only works on nightly. This is probably the best we'll get with #179 for now.

### DIFF
--- a/gtk4-macros/src/attribute_parser.rs
+++ b/gtk4-macros/src/attribute_parser.rs
@@ -78,7 +78,7 @@ fn parse_attribute(meta: &NestedMeta) -> Result<(String, String)> {
 }
 
 pub enum FieldAttributeArg {
-    Id(String),
+    Id(String, Span),
     Internal(bool),
 }
 
@@ -104,11 +104,19 @@ impl AttributedField {
     pub fn id(&self) -> String {
         let mut name = None;
         for arg in &self.attr.args {
-            if let FieldAttributeArg::Id(value) = arg {
+            if let FieldAttributeArg::Id(value, _) = arg {
                 name = Some(value)
             }
         }
         name.cloned().unwrap_or_else(|| self.ident.to_string())
+    }
+    pub fn id_span(&self) -> Span {
+        for arg in &self.attr.args {
+            if let FieldAttributeArg::Id(_, span) = arg {
+                return *span;
+            }
+        }
+        self.ident.span()
     }
 }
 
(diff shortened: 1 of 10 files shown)


gtk4-macros/src/attribute_parser.rs
```diff
@@ -273,9 +273,28 @@ pub fn parse_fields(fields: &Fields) -> Result<Vec<AttributedField>, Error> {
...
+            if let syn::Type::Path(syn::TypePath { path, .. }) = &field.ty {
+                if path.leading_colon.is_none()
+                    && path.segments.len() == 1
+                    && matches!(
+                        &path.segments[0].arguments,
+                        syn::PathArguments::AngleBracketed(_)
+                    )
+                    && path.segments[0].ident == "TemplateChild"
+                {
+                    proc_macro_error::emit_warning!(
```
sdroege, line 293: If you don't want to make this an error you probably want to add some kind of `#[allow(gtk4::template_child_without_attribute)]` or similar to silence it

gtk4-macros/src/attribute_parser.rs
```diff
@@ -273,9 +273,28 @@ pub fn parse_fields(fields: &Fields) -> Result<Vec<AttributedField>, Error> {
...
+                        field,
+                        "field with type `TemplateChild` possibly missing #[template_child] attribute",
+                    );
```
sdroege, line 296: Maybe this should also get a compile test?

sdroege:
Looks good to me otherwise

gtk4-macros/src/attribute_parser.rs
jf2048: We would need to have a way to only run the test on nightly, I don't believe there is an easy way to do that

jf2048: Just changed it to an error and added a compile test

gtk4-macros/src/attribute_parser.rs
jf2048: Works now with `#[template(file = "...", allow_template_child_without_attribute)]`

gtk4-macros/src/attribute_parser.rs
sdroege: Makes sense

sdroege (approved):

PR #964, opened 2022-03-09
Add is method to fundamental types

### LINKED ISSUE #860: Add is methods to fundamental types
See https://github.com/gtk-rs/gtk-rs-core/issues/502

bilelmoussaoui:
Fixes #860

### DIFF
--- a/gtk4/src/expression.rs
+++ b/gtk4/src/expression.rs
@@ -50,7 +50,6 @@ impl Expression {
     pub fn type_(&self) -> Type {
         unsafe {
             let ptr = self.to_glib_none().0;
-
             from_glib((*(*(ptr as *mut glib::gobject_ffi::GTypeInstance)).g_class).g_type)
         }
     }
--- a/gtk4/src/expression.rs
+++ b/gtk4/src/expression.rs
@@ -23,9 +23,13 @@ pub unsafe trait IsExpression:
 }
 
 impl Expression {
+    pub fn is<E: IsExpression>(&self) -> bool {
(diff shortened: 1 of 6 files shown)


gtk4/src/expression.rs
```diff
@@ -23,9 +23,13 @@ pub unsafe trait IsExpression:
 }
 
 impl Expression {
+    pub fn is<E: IsExpression>(&self) -> bool {
+        self.type_() == E::static_type()
```
bilelmoussaoui, line 27: I am wondering whether it makes sense to keep the `type_()` method as public

bilelmoussaoui: I did that in a separate commit so we can only cherry-pick the first one for the next release

sdroege: Seems useful to get the type?

sdroege (approved):

sdroege (changes requested):
I would keep `type_()` public.

gtk4/src/expression.rs
bilelmoussaoui: I can't see where you would need that in general

sdroege: It's also not hurting to keep it :) It's potentially useful information and you can't get that any other way right now.

People are also using the `type_()` function on objects.

bilelmoussaoui: I dropped that commit, I will add similar functions to Gsk.RenderNode/Gdk.Event then

sdroege (approved):

PR #968, opened 2022-03-10
book: Add nix to linux installation guide

khrj:
Adds instructions to start a new shell with gtk4 and gcc using [Nix](https://nixos.org/guides/how-nix-works.html) with [nix-command](https://nixos.wiki/wiki/Nix_command) enabled

### DIFF
--- a/book/src/installation_linux.md
+++ b/book/src/installation_linux.md
@@ -24,8 +24,9 @@ Arch and derivatives:
 sudo pacman -S gtk4 base-devel
 ```
 
-Nix (with nix-command)
+Nix:  
+Add the following inputs to your derivation:
 
 ```
-nix shell nixpkgs#{gtk4, gcc}
+gtk4, gcc
 ```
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
Not sure if we should include experimental commands. Maybe use the current way of installing nix packages?

khrj:
nix-command (and flakes) isn't really in beta, it's been around for quite some time now. At this point, there's just two different ways of using nix, via the old CLI, and via the new one. If you wish, I can add instructions for the old CLI too.

Hofer-Julian:
As long as you have to activate a experimental flag for the new one I will only accept instructions for the old one.

khrj:
@Hofer-Julian Hey, this slipped out of my mind - I just realised that libraries won't be available when installed using a shell, regardless of whether it uses the old or new syntax. The dependencies need to be added to a derivation - I've updated the instructions accordingly

Hofer-Julian (approved):
Thanks!

PR #971, opened 2022-03-17
Rebase to 0.4

bilelmoussaoui:
cc @sdroege @jf2048 

Will do a new release once this lands


bilelmoussaoui:
I wonder if the test attribute macro requires a recent rustc

jf2048:
That's an odd error, let me do some digging

gtk4-macros/src/lib.rs
```diff
@@ -256,3 +272,67 @@ pub fn template_callbacks(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+            quote! {
+                #(#attrs)*
+                #[::std::prelude::v1::test]
+                #vis #sig {
+                    #crate_ident::test_synced(move || {
+                        #block
+                    })
+                }
+            }
+            .into()
```
jf2048, line 334:
```suggestion
            let test = quote! {
                #(#attrs)*
                #[::std::prelude::v1::test]
                #vis #sig {
                    #crate_ident::test_synced(move || {
                        #block
                    })
                }
            };
            test.into()
```

I'm not sure why, but this seems to fix it

jf2048 (changes requested):

gtk4-macros/src/lib.rs
jf2048: #977

bilelmoussaoui: Awesome, thanks

bilelmoussaoui: >  It doesn't affect that because we've bumped the version to rustc 1.57.0 

There is no need for that I would say

PR #973, opened 2022-03-18
gdk: Implement Value traits for Key

### LINKED ISSUE #972: [FEATURE REQUEST] Using `gtk::gdk::Key` with `closure_local!`
Thank you for the wonderful crate.
`gtk::gdk::Key` does not implement `FromValue` and cannot be used with `closure_local!`

```rust
let text_view = gtk::TextView::new();
let controller = gtk::EventControllerKey::new();
let closure = closure_local!(
  @watch text_view =>
  move |_: &gtk::EventControllerKey,
    key: gtk::gdk::Key,
    _: u32,
    _: gtk::gdk::ModifierType| { gtk::Inhibit(false) }
);
controller.connect_closure("key-pressed", false, closure);
text_view.add_controller(&controller);
```

```sh
the trait bound `Key: FromValue<'_>` is not satisfied
the trait `FromValue<'_>` is not implemented for `Key`
```

bilelmoussaoui:
There is no way to have proper value checking for Key, so we just assumes it is used
in a correct use case.

Fixes #972

### DIFF
--- a/gdk4/src/keys.rs
+++ b/gdk4/src/keys.rs
@@ -1,7 +1,7 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 use glib::translate::*;
-use glib::GString;
+use glib::{value::*, GString, Type, Value};
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 // rustdoc-stripper-ignore-next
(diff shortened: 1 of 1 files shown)


gdk4/src/keys.rs
```diff
@@ -4662,3 +4662,46 @@ impl IntoGlib for Key {
         self.0
     }
 }
+
+impl ValueType for Key {
+    type Type = u32;
+}
+
+unsafe impl<'a> FromValue<'a> for Key {
+    type Checker = glib::value::NopChecker;
```
sdroege, line 4671: You need to check if it has the correct `GType` at least. Right now you could get a `gdk::Key` out of a value containing e.g. a widget :P

gdk4/src/keys.rs
```diff
@@ -4662,3 +4662,46 @@ impl IntoGlib for Key {
...
+    type Type = u32;
+}
+
+unsafe impl<'a> FromValue<'a> for Key {
+    type Checker = glib::value::ValueTypeMismatchError;
+
+    unsafe fn from_value(value: &'a Value) -> Self {
+        let res: u32 = glib::gobject_ffi::g_value_get_uint(value.to_glib_none().0);
+        // safe because the check is done by `Self::Checker`
+        Key::from_glib(res)
```
sdroege, line 4676: What happens if you use a random integer as a `gdk::Key`? What happens in the worst case?

This needs a bit more commenting in the code.

sdroege: And is `0` maybe a special invalid value that is used in various placed to denote the absence of a key?

bilelmoussaoui: Nothing, all the functions keeps working as is. the name returns a stringified integer, and upper/lower return the same inner value

sdroege: And is one key value used as the equivalent of `None`? :)

bilelmoussaoui: And is 0 (or -1?) maybe a special invalid value that is used in various placed to denote the absence of a key?

There is a key used for that, `VoidSymbol` but it is only used for the `from_name` implementation per my testing

sdroege: Ok, seems fine then

sdroege:
No there's only the regen check failing :)

sdroege (approved):

PR #976, opened 2022-03-18
gdk4-wayland: Use wayland_client::Display instead of a display proxy

jf2048:
I can't figure out how this is supposed to work without this, the display proxy seems useless since you can't bind any globals on it?


bilelmoussaoui:
Are the other wayland-crate related functions correct as well?

jf2048:
Yes, this is the only one that needs a special handling in wayland-client.

jf2048:
On a side note, I don't see how we can ever upgrade to wayland-client-0.30, the C API compatibility is gone from there

bilelmoussaoui:
Thanks!

PR #979, opened 2022-03-23
book: Bring todo closer to libadwaita example

Hofer-Julian:


book/listings/actions/5/resources/resources.gresource.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <gresources>
```
bilelmoussaoui, line 2: Would be nice to have an .editorconfig here to ensure all those and future files are following the same conventions.

Hofer-Julian: Can that be enforced with the CI?

bilelmoussaoui: It can't but VSCode/Builder uses those to enforce things for newly written files/changes to existing ones which is good enough

PR #984, opened 2022-03-28
Regen with latest gir/gir-files

bilelmoussaoui:


gdk4-wayland/src/auto/wayland_gl_context.rs
```diff
@@ -13,6 +13,8 @@ glib::wrapper! {
     }
 }
 
+impl WaylandGLContext {}
```
bilelmoussaoui, line 16: I don't know what is causing this regression...

sdroege: It's since a while, I also don't know

gdk4/sys/Cargo.toml
```diff
@@ -26,6 +26,9 @@ version = "4.3.2"
 [package.metadata.system-deps.gtk4.v4_6]
 version = "4.5"
```
sdroege, line 27: Should maybe be 4.6 at this point?

sdroege (commented):
Seems ok to me otherwise

gdk4-wayland/src/auto/wayland_gl_context.rs
bilelmoussaoui: I will investigate in the future. Opened an issue regarding that for now https://github.com/gtk-rs/gir/issues/1334

PR #998, opened 2022-04-10
gtk: Add NativeDialog::run_async

### LINKED ISSUE #993: Missing `NativeDialogExtManual::run_async`
The `DialogExtManual` trait has both a `run_future` and a `run_async` method. Its `NativeDialogExtManual` counterpart, however, only has `run_future`. This seems to be an oversight. Please add the missing method for consistency.

bilelmoussaoui:
Fixes #993

cc @piegamesde


piegamesde:
Thank you.

It appears that the closure does not hold a direct strong reference to the dialog. Therefore, one would still have to take care of that otherwise the dialog won't show at all.

bilelmoussaoui:
I would like to handle that separately by making the connect_response always moves a strong ref of self if possible. Let me try that in a separate commit

sdroege:
> I would like to handle that separately by making the connect_response always moves a strong ref of self if possible. Let me try that in a separate commit

That doesn't seem like a good idea. That's very surprising behaviour for signal connection and even has the potential to break existing code that assumes that the dialog can be just dropped (which is not going to happen anymore now because the response signal creates a circular reference).

sdroege (approved):

gtk4/src/native_dialog.rs
```diff
@@ -29,6 +30,26 @@ pub trait NativeDialogExtManual {
...
+    ///
+    /// ```no_run
+    /// # use gtk4 as gtk;
+    /// use gtk::prelude::*;
+    ///
+    /// let dialog = gtk::FileChooserNative::builder()
+    ///    .title("Select a File")
+    ///    .build();
+    ///
+    /// dialog.run_async(|obj, answer| {
```
sdroege, line 47:
```suggestion
    /// dialog.run_async(|dialog, answer| {
```

bilelmoussaoui:
> That doesn't seem like a good idea. That's very surprising behaviour for signal connection and even has the potential to break existing code that assumes that the dialog can be just dropped (which is not going to happen anymore now because the response signal creates a circular reference).

Indeed, that would be a breaking change of the behaviour. Although, I suppose it is fine if that happens in a new 0.x release?

sdroege:
It's also about it being rather unintuitive behaviour going forward. It behaves very different to any other signal handler, and like I said it creates a reference cycle :)

sdroege:
What's the plan/status here?

bilelmoussaoui:
> What's the plan/status here?

I will get back to this in the upcoming week

sdroege (approved):

PR #1003, opened 2022-04-23
book: Link snippets to code examples on github

Hofer-Julian:
Also adds css so the links inherit the color

CC: @ebassi


ebassi:
Thanks, this is going to be very helpful!

PR #1007, opened 2022-04-24
book: Replace git version of the book with link to stable

Hofer-Julian:


git/book
```diff
@@ -0,0 +1 @@
+/var/home/julian/Projekte/gnome/libraries/gtk4-rs/stable/latest/book
```
bilelmoussaoui, line 1: That is probably not gonna work :p

bilelmoussaoui: That is how I do it for the docs for each release
```
    - cd $repo
    - unlink ./stable/latest/docs
    - cd ./stable/latest
    - ln -sf ../$version/docs ./docs
 ```

Hofer-Julian: Mmmh weird, that's how I created the symbolic link as well.

Will have a look later.

Hofer-Julian: For some reason, I had to pass the `-r` flag for it to work.

Should be fine now.

PR #1012, opened 2022-04-28
book: Refactor `current_tasks`

Hofer-Julian:


bilelmoussaoui:
OnceCell is clearly the correct thing to use here btw

Hofer-Julian:
> OnceCell is clearly the correct thing to use here btw

Yes, but it won't anymore for the libadwaita todo version and this keeps the diff to a minimum.

PR #1019, opened 2022-05-04
Improve API for some Snapshot methods

GuillaumeGomez:
Both methods make the `Snapshot` instance unuseable and so should take ownership of it to limit potential misuse from the users.


gtk4/src/snapshot.rs
```diff
@@ -27,4 +27,19 @@ impl Snapshot {
     pub fn push_debug(&self, message: &str) {
         unsafe { ffi::gtk_snapshot_push_debug(self.to_glib_none().0, message.to_glib_none().0) }
     }
+
+    #[doc(alias = "gtk_snapshot_to_node")]
+    pub fn to_node(self) -> Option<gsk::RenderNode> {
```
sdroege, line 32: Should this really be `Option`s or `Result<T, glib::BoolError>`?

GuillaumeGomez: What the `BoolError` should be? It's not necessarily an error when it returns `NULL`.

sdroege: Ah it's not? Then it should indeed be an `Option`

GuillaumeGomez:
Anything else to be done in here?

bilelmoussaoui:
> Anything else to be done in here?

I am waiting to see if the ToGlibPtrMove traits will land soon-ish or not, we can update the PR before landing it then.

sdroege:
It's there now but the gir-related changes are still pending. @GuillaumeGomez what's the status on that?

GuillaumeGomez:
Still need to do it.

GuillaumeGomez:
I finally took the time to update it. I needed to update the gir submodules too to get the change so the first

bilelmoussaoui:
Closing in favour of https://github.com/gtk-rs/gtk4-rs/pull/1097

PR #1020, opened 2022-05-04
Add confetti_snapshot_animation example

ranfdev:
https://user-images.githubusercontent.com/23294184/166829688-19e706d3-b7b2-4dcf-b68f-026c1a77cb96.mp4

I feel there aren't many examples around on how to make animations... So, here it is


examples/confetti_snapshot_animation/animated_explosion/imp.rs
```diff
@@ -0,0 +1,29 @@
...
+use gtk::glib;
+use gtk::prelude::*;
+use gtk::subclass::prelude::*;
+use std::cell::{Cell, RefCell};
+
+#[derive(Default)]
+pub struct AnimatedExplosion {
+    pub(crate) explosion: RefCell<Option<Explosion>>,
+    pub(crate) start: RefCell<f64>,
+    pub(crate) lastupdate: RefCell<f64>,
```
bilelmoussaoui, line 12:
```suggestion
    pub(crate) start: Cell<f64>,
    pub(crate) lastupdate: Cell<f64>,
```

bilelmoussaoui (commented):
I will do a proper review tomorrow

examples/confetti_snapshot_animation/animated_explosion/imp.rs
sdroege: Also `pub(super)` is enough or not?

examples/confetti_snapshot_animation/animated_explosion/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+        let imp = this.imp();
+        imp.explosion.replace(Some(Explosion::new(parameters)));
+        this
+    }
+    pub fn animate(&self, paintable: &ConfettiWidget, duration: u32) {
+        let imp = self.imp();
+
+        // A FrameClock tells the application when to update and repaint a surface.
+        // This may be synced to the vertical refresh rate of the monitor, for example.
+        let frame_clock = paintable.frame_clock().unwrap();
```
sdroege, line 25: Why is all the implementation not in the `imp` module? :)

ranfdev: Because some of the code requires the implementation, some requires the object instance... I've never fully understood where to put things. Probably I just need to split things in more methods, and put each method in the correct module...

ranfdev:
I'll refactor some things...

examples/confetti_snapshot_animation/animated_explosion/mod.rs
sdroege: You'd pass the object as reference forward usually, see the shape of the methods in the Impl traits.

examples/confetti_snapshot_animation/animated_explosion/imp.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl AnimatedExplosion {
+    pub(super) fn init_time(&self, clock: &gdk::FrameClock, duration: f64) {
+        let time = clock.frame_time() as f64 / 1000.0;
+        self.start.replace(time);
+        self.lastupdate.replace(time);
+        self.duration.replace(duration);
+    }
+    // Returns ControlFlow::Break if the animation is finished
+    pub(super) fn update(&self, clock: &gdk::FrameClock) -> Continue {
```
bilelmoussaoui, line 32: As much as I really like the usage of ControlFlow here, I think we should wait until we do the switch in glib-rs first & the generated bindings.

examples/confetti_snapshot_animation/animated_explosion/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+impl AnimatedExplosion {
+    pub(super) fn new(parameters: ExplosionParameters) -> Self {
+        let this: Self = glib::Object::new(&[]).expect("Failed to create a AnimatedExplosion");
+        let imp = this.imp();
+        imp.explosion.replace(Some(Explosion::new(parameters)));
+        this
+    }
+    pub(super) fn init_time(&self, clock: &gdk::FrameClock, duration: f64) {
```
bilelmoussaoui, line 19: What is the point of having a public function that calls a private one? I would just have the implementation here.

bilelmoussaoui (commented):
Sorry for the delay, the rest looks good to me!

bilelmoussaoui:
Any updates here?

ranfdev:
I'll see if I can fix the issues tomorrow

bilelmoussaoui:
rustfmt is not happy, mind fixing that? otherwise lgtm

bilelmoussaoui:
Thank you

PR #1021, opened 2022-05-05
fix vendoring by removing include from gdk4/Cargo.toml

ranfdev:
With that line, nothing in gdk4 gets vendored. See this explanation: https://github.com/rust-lang/cargo/issues/10617#issuecomment-1114361442

I've decided to just remove that line, because other crates in gtk4 don't have it.


sdroege:
It also doesn't need this because the license file is symlinked into the crate directory and because of that included by default anyway.

sdroege (approved):

PR #1024, opened 2022-05-07
book: Migrate to `glib-build-tools`

RealKC:


book/src/composite_templates.md
```diff
@@ -45,25 +45,23 @@ Filename: <a class=file-link href="https://github.com/gtk-rs/gtk4-rs/blob/master
 ```
 
 Now we have to compile the resources and link it to our application.
-One way to do this is to execute [`gio::compile_resources`](https://gtk-rs.org/gtk-rs-core/stable/latest/docs/gio/fn.compile_resources.html) within a cargo [build script](https://doc.rust-lang.org/cargo/reference/build-scripts.html).
+One way to do this is to execute [`gio::compile_resources`](https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib-build-tools/fn.compile_resources.html) within a cargo [build script](https://doc.rust-lang.org/cargo/reference/build-scripts.html).
```
RealKC, line 48: This link is going to be broken until docs are built for glib-build-tools, is that fine?

RealKC:
I'm not sure what's causing the CI failures

bilelmoussaoui:
> I'm not sure what's causing the CI failures

https://github.com/gtk-rs/gtk4-rs/pull/1025#event-6567408059

Hofer-Julian (commented):
Thanks!
But I guess this crate will be added to crates.io?
I'd rather wait for that.

RealKC:
That's fine with me

Hofer-Julian:
Then let's wait for its release, update Cargo.toml and merge then :)

RealKC:
Alright, should be rebased on master now, and I fixed a broken link that snuck in. CI failed on my first push but that didn't seem to be related to my changes?

Hofer-Julian:
Thanks, could you please also update the relevant section in the book where we add the gtk build dependency?

RealKC:
That's done now

Hofer-Julian:
> That's done now

Thanks, I will later fix the build failures on your branch and merge it.

book/listings/actions/7/window/imp.rs
```diff
@@ -40,14 +40,14 @@ impl ObjectSubclass for Window {
...
+    fn constructed(&self) {
         // Call "constructed" on parent
-        self.parent_constructed(obj);
+        self.parent_constructed();
 
         // Setup
-        obj.setup_settings();
-        obj.setup_actions();
-        obj.bind_settings();
+        self.instance().setup_settings();
```
bilelmoussaoui, line 48: I would suggest calling self.instance() only once and store it in an obj / window variable.

book/listings/g_object_signals/2/custom_button/imp.rs
```diff
@@ -89,15 +78,16 @@ static MAX_NUMBER: i32 = 8;
...
 impl ButtonImpl for CustomButton {
-    fn clicked(&self, button: &Self::Type) {
+    fn clicked(&self) {
         let incremented_number = self.number.get() + 1;
         // If `number` reached `MAX_NUMBER`,
         // emit "max-number-reached" signal and set `number` back to 0
         if incremented_number == MAX_NUMBER {
-            button.emit_by_name::<()>("max-number-reached", &[&incremented_number]);
-            button.set_property("number", &0);
+            self.instance()
```
bilelmoussaoui, line 86: Same as elsewhere

book/listings/actions/7/window/imp.rs
bilelmoussaoui: I don't think there should be any, cc @sdroege for confirming that. But it doesn't make much sense to repeat the same call if you can store the obj once and re-use it.

PR #1028, opened 2022-05-08
Migrate state storing back to json

### LINKED ISSUE #1023: Moving task saving of ToDo app back to serde_json
It currently uses gio::Settings to save its data, let's move back to json.

Hofer-Julian:
Fixes #1023


sdroege (commented):
Much better :)

PR #1029, opened 2022-05-09
Base Dockerfile on gtk-rs-core image

jf2048:
Part of https://github.com/gtk-rs/gtk-rs-core/issues/484

Depends on an image published from https://github.com/gtk-rs/gtk-rs-core/pull/676


bilelmoussaoui:
What is the status quo here?

jf2048:
This seems to be working, this is probably a good time to do this in case the CI breaks again

PR #1042, opened 2022-05-20
Update gir

sdroege:
See https://github.com/gtk-rs/gir/pull/1357


GuillaumeGomez (approved):

sdroege:
```
 error: returning the result of a `let` binding from a block
   --> gtk4/src/auto/assistant.rs:219:13
    |
218 |             let res = (*callback)(current_page);
    |             ------------------------------------ unnecessary `let` binding
219 |             res
    |             ^^^
    |
```

So still one left in another context (CC @filnet). Let's keep the clippy warning for now until that's covered at some point too.

PR #1053, opened 2022-06-04
WIP: gdk4-wayland: Update wayland crate to 0.30

A6GibKm:
Marked as WIP until 0.30.0 is released.

Open questions: Should the functions return a result or option now? note that there are two different errors involved and option might be easier.


gdk4-wayland/Cargo.toml
```diff
@@ -28,7 +28,8 @@ gdk = {path = "../gdk4", package = "gdk4", version = "0.5.0"}
 gio = {git = "https://github.com/gtk-rs/gtk-rs-core", version = "0.16", features = ["v2_66"]}
 glib = {git = "https://github.com/gtk-rs/gtk-rs-core", version = "0.16", features = ["v2_66"]}
 libc = "0.2"
-wayland-client = {version = "0.29", features = ["use_system_lib"], optional = true}
+wayland-client = {version = "0.30.0-beta.3", optional = true}
+wayland-backend = {version = "0.1.0-beta.3", optional = true, features = ["client_system"]}
```
bilelmoussaoui, line 32: Do we need the wayland-backend crate or can it be enabled as a feature of wayland-client directly? as I don't see it used directly from wayland_backend.

gdk4-wayland/src/wayland_device.rs
```diff
@@ -26,8 +31,10 @@ impl WaylandDevice {
     #[cfg_attr(feature = "dox", doc(cfg(feature = "wayland_crate")))]
     pub fn wl_keyboard(&self) -> WlKeyboard {
         unsafe {
+            let cnx = wayland_client::Connection::connect_to_env().unwrap();
```
bilelmoussaoui, line 34: I would avoid unwraps and use expect here & everywhere else. Just in case

gdk4-wayland/Cargo.toml
A6GibKm: Compilation fails without it...

bilelmoussaoui:
cc @jf2048 as you have had some usages of gdk-wayland/wayland-crate usages recently

gdk4-wayland/Cargo.toml
i509VCB: The client_system feature can only be enabled via the wayland-backend crate FYI.

i509VCB:
An open question worth asking is do we want to return a type to wrap the guest `Display` from wayland-client so we don't need to construct a guest display in each function?

jf2048:
I agree, this PR doesn't seem right. I don't think we want to create a new connection for each proxy. This is creating a new event queue every time one of these functions is called, and I don't think it will even work as these proxies are already assigned to GTK's event queue.

Also, GTK can destroy these proxies at any time, so these functions seem inherently unsafe.

i509VCB:
> I agree, this PR doesn't seem right. I don't think we want to create a new connection for each proxy. This is creating a new event queue every time one of these functions is called, and I don't think it will even work as these proxies are already assigned to GTK's event queue.
> 
> Also, GTK can destroy these proxies at any time, so these functions seem inherently unsafe.

I can't recall if 0.30 added a way to create an event queue from a pointer. Creating protocol objects requires an event queue unless you use the more raw wayland-backend functions.

On proxy destruction let me look further into that

gdk4-wayland/src/wayland_device.rs
```diff
@@ -25,9 +30,15 @@ impl WaylandDevice {
     #[cfg(any(feature = "wayland_crate", feature = "dox"))]
     #[cfg_attr(feature = "dox", doc(cfg(feature = "wayland_crate")))]
     pub fn wl_keyboard(&self) -> WlKeyboard {
+        let display = self.display().downcast::<crate::WaylandDisplay>().unwrap();
         unsafe {
-            let ptr = ffi::gdk_wayland_device_get_wl_keyboard(self.to_glib_none().0);
-            Proxy::from_c_ptr(ptr as *mut wl_proxy).into()
+            let display_ptr = ffi::gdk_wayland_display_get_wl_display(display.to_glib_none().0);
+            let keyboard_ptr = ffi::gdk_wayland_device_get_wl_keyboard(self.to_glib_none().0);
```
i509VCB, line 36: What policy does gtk have with returning pointers. Are these always going to return non-null pointers?

jf2048: - `gdk_wayland_display_get_wl_display` will never return null
- `gdk_wayland_device_get_wl_keyboard` can return null if the seat has no keyboard, or if the seat caps change and the keyboard gets removed

gdk4-wayland/src/wayland_device.rs
```diff
@@ -3,12 +3,17 @@
...
-use glib::translate::ToGlibPtr;
+use crate::{glib::translate::ToGlibPtr, gdk::prelude::*};
 #[cfg(any(feature = "wayland_crate", feature = "dox"))]
 #[cfg_attr(feature = "dox", doc(cfg(feature = "wayland_crate")))]
 use wayland_client::{
-    protocol::{wl_keyboard::WlKeyboard, wl_pointer::WlPointer, wl_seat::WlSeat},
-    sys::client::wl_proxy,
+    backend::ObjectId,
+    protocol::{
+        __interfaces::{WL_KEYBOARD_INTERFACE, WL_POINTER_INTERFACE, WL_SEAT_INTERFACE},
```
i509VCB, line 12: Could use the implementations of `Proxy` with the `interface` function.

Since you already import `Proxy`, you'd use `WlSeat::interface()` for example

gdk4-wayland/src/wayland_device.rs
i509VCB: Okay so `wl_keyboard` and friends should be wrapped in an `Option` then

gdk4-wayland/src/wayland_device.rs
```diff
@@ -36,9 +47,15 @@ impl WaylandDevice {
     #[cfg(any(feature = "wayland_crate", feature = "dox"))]
     #[cfg_attr(feature = "dox", doc(cfg(feature = "wayland_crate")))]
     pub fn wl_pointer(&self) -> WlPointer {
```
i509VCB, line 49: Wrap in Option

gdk4-wayland/src/wayland_device.rs
```diff
@@ -47,9 +64,15 @@ impl WaylandDevice {
     #[cfg(any(feature = "wayland_crate", feature = "dox"))]
     #[cfg_attr(feature = "dox", doc(cfg(feature = "wayland_crate")))]
     pub fn wl_seat(&self) -> WlSeat {
```
i509VCB, line 66: Wrap in Option

gdk4-wayland/src/wayland_display.rs
```diff
@@ -38,19 +42,29 @@ impl WaylandDisplay {
     #[cfg_attr(feature = "dox", doc(cfg(feature = "wayland_crate")))]
     pub fn wl_compositor(&self) -> WlCompositor {
```
i509VCB, line 43: Wrap in option

gdk4-wayland/src/wayland_monitor.rs
```diff
@@ -3,20 +3,30 @@
...
+    protocol::{__interfaces::WL_OUTPUT_INTERFACE, wl_output::WlOutput},
+    Proxy,
+};
 
 impl WaylandMonitor {
     #[doc(alias = "gdk_wayland_monitor_get_wl_output")]
     #[doc(alias = "get_wl_output")]
     #[cfg(any(feature = "wayland_crate", feature = "dox"))]
     #[cfg_attr(feature = "dox", doc(cfg(feature = "wayland_crate")))]
     pub fn wl_output(&self) -> WlOutput {
```
i509VCB, line 20: Wrap in Option

gdk4-wayland/src/wayland_seat.rs
```diff
@@ -3,20 +3,30 @@
...
+    protocol::{__interfaces::WL_SEAT_INTERFACE, wl_seat::WlSeat},
+    Proxy,
+};
 
 impl WaylandSeat {
     #[doc(alias = "gdk_wayland_seat_get_wl_seat")]
     #[doc(alias = "get_wl_seat")]
     #[cfg(any(feature = "wayland_crate", feature = "dox"))]
     #[cfg_attr(feature = "dox", doc(cfg(feature = "wayland_crate")))]
     pub fn wl_seat(&self) -> WlSeat {
```
i509VCB, line 20: Wrap in option

gdk4-wayland/src/wayland_surface.rs
```diff
@@ -3,12 +3,16 @@
...
-use glib::translate::*;
+use crate::{glib::translate::*, gdk::prelude::*};
 use glib::IsA;
 
 #[cfg(any(feature = "wayland_crate", feature = "dox"))]
 #[cfg_attr(feature = "dox", doc(cfg(feature = "wayland_crate")))]
-use wayland_client::{protocol::wl_surface::WlSurface, sys::client::wl_proxy, Proxy};
+use wayland_client::{
+    backend::ObjectId,
+    protocol::{__interfaces::WL_SURFACE_INTERFACE, wl_surface::WlSurface},
```
i509VCB, line 13: Use `WlSurface::interface()`

gdk4-wayland/src/wayland_device.rs
A6GibKm: Thanks

A6GibKm:
> An open question worth asking is do we want to return a type to wrap the guest `Display` from wayland-client so we don't need to construct a guest display in each function?

I added a commit trying to store the connection in the display, I have no clue if it works.

bilelmoussaoui:
Would you mind bumping this to use the latest beta and see if everything is still "working" as expected?

gdk4-wayland/Cargo.toml
```diff
@@ -28,9 +28,11 @@ gdk = {path = "../gdk4", package = "gdk4", version = "0.5.0"}
 gio = {git = "https://github.com/gtk-rs/gtk-rs-core", version = "0.16", features = ["v2_66"]}
 glib = {git = "https://github.com/gtk-rs/gtk-rs-core", version = "0.16", features = ["v2_66"]}
 libc = "0.2"
-wayland-client = {version = "0.29", features = ["use_system_lib"], optional = true}
+wayland-client = {version = "0.30.0-beta.7", optional = true}
+wayland-backend = {version = "0.1.0-beta.7", optional = true, features = ["client_system"]}
 khronos-egl = {version = "4.1.0", optional = true}
 xkb = {version = "0.2.1", optional = true}
+once_cell = {version = "1.0", optional = true}
```
bilelmoussaoui, line 35: glib re-exports once_cell btw, maybe we can use it from there directly?

bilelmoussaoui:
Do you mind updating to the latest beta?

bilelmoussaoui:
Closing in favour of https://github.com/gtk-rs/gtk4-rs/pull/1146

PR #1056, opened 2022-06-16
regenerate with latest gir/gir-files

alatiera:
Regen from https://github.com/gtk-rs/gir-files/pull/116


alatiera:
wip as we need to merge the gir-files update and then fix the submodule sha

gtk4/src/auto/editable_label.rs
```diff
@@ -69,6 +69,10 @@ impl EditableLabel {
         }
     }
 
+    pub fn set_editing(&self, editing: bool) {
```
bilelmoussaoui, line 72: needs a since annotation

gtk4/src/auto/flags.rs
```diff
@@ -261,6 +261,8 @@ bitflags! {
         const A11Y = ffi::GTK_DEBUG_A11Y as u32;
         #[doc(alias = "GTK_DEBUG_ICONFALLBACK")]
         const ICONFALLBACK = ffi::GTK_DEBUG_ICONFALLBACK as u32;
+        #[doc(alias = "GTK_DEBUG_INVERT_TEXT_DIR")]
+        const INVERT_TEXT_DIR = ffi::GTK_DEBUG_INVERT_TEXT_DIR as u32;
```
bilelmoussaoui, line 265: needs a v4_8 guard

gtk4/src/auto/inscription.rs
```diff
@@ -0,0 +1,968 @@
...
+    #[doc(alias = "GtkInscription")]
+    pub struct Inscription(Object<ffi::GtkInscription, ffi::GtkInscriptionClass>) @extends Widget, @implements Accessible, Buildable, ConstraintTarget;
+
+    match fn {
+        type_ => || ffi::gtk_inscription_get_type(),
+    }
+}
+
+impl Inscription {
+    #[cfg(any(feature = "v4_8", feature = "dox"))]
```
bilelmoussaoui, line 47: it is weird that all the methods are guarded with v4_8 but not the struct itself instead.

gtk4/src/auto/inscription.rs
```diff
@@ -0,0 +1,968 @@
...
+    #[doc(alias = "get_text")]
+    pub fn text(&self) -> Option<glib::GString> {
+        unsafe { from_glib_none(ffi::gtk_inscription_get_text(self.to_glib_none().0)) }
+    }
+
+    //#[cfg(any(feature = "v4_8", feature = "dox"))]
+    //#[cfg_attr(feature = "dox", doc(cfg(feature = "v4_8")))]
+    //#[doc(alias = "gtk_inscription_get_text_overflow")]
+    //#[doc(alias = "get_text_overflow")]
+    //pub fn text_overflow(&self) -> /*Ignored*/InscriptionOverflow {
```
bilelmoussaoui, line 117: InscriptionOverflow needs to be added to Gir.toml

alatiera: Indeed good catch, I missed this.

gtk4/src/auto/bookmark_list.rs
```diff
@@ -61,6 +64,20 @@ impl BookmarkList {
         }
     }
 
+    #[cfg(any(feature = "v4_8", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "v4_8")))]
+    #[doc(alias = "item-type")]
+    pub fn item_type(&self) -> glib::types::Type {
```
ebassi, line 70: I'm not sure you want methods here calling into the properties; the `gio::ListModel` API requires a cast, but it's still going to be faster than involving `g_object_get_property()` and `GValue`…

gtk4/sys/src/lib.rs
```diff
@@ -399,6 +399,12 @@ pub const GTK_INPUT_PURPOSE_PASSWORD: GtkInputPurpose = 8;
 pub const GTK_INPUT_PURPOSE_PIN: GtkInputPurpose = 9;
 pub const GTK_INPUT_PURPOSE_TERMINAL: GtkInputPurpose = 10;
 
+pub type GtkInscriptionOverflow = c_int;
```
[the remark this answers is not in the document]
alatiera: Oh this is the same as `GdkScrollUnit` which also doesn't have a guard apart the method, so maybe we don't need one or is this another bug?

gtk4/src/auto/editable_label.rs
alatiera: wth, where does this even come from https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkeditablelabel.h

bilelmoussaoui: https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkeditablelabel.c#L376

gtk4/src/auto/bookmark_list.rs
sdroege: Erm why do we create property accessors for properties defined by interfaces? That seems like a bug to me :) All these are already available via the `ListModel` API and just a duplication like this.

Can you create a gir issue for this?

gtk4/src/auto/editable_label.rs
alatiera: The property existed before, but looks like it was changed to readwrite instead of readable, so I think we need to guard that on the toml on our side

alatiera: Ended up copying the implementation to manually and explicitly only generating get/notify. We can't easily add a version guard only for the `set_` wrapper of the property.

gtk4/src/auto/bookmark_list.rs
alatiera: Created https://github.com/gtk-rs/gir/issues/1369

gtk4/sys/src/lib.rs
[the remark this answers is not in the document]
alatiera: Looking at this again, I think its fine since its just copying the values of the int instead of referencing the C enum.

bilelmoussaoui:
Could you update the submodule and regenerate again?

gtk4/src/editable_label.rs
```diff
@@ -0,0 +1,29 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
```
bilelmoussaoui, line 1: needs to be added to lib.rs

bilelmoussaoui:
otherwise lgtm, thanks!

PR #1060, opened 2022-06-22
Improved grammar and readability

JoelMon:
There were a few strange grammatical quirks such as commas when a comma isn't called for.


Hofer-Julian (approved):
Thanks!

PR #1061, opened 2022-06-23
book: Stop using hyphens in app ids

Hofer-Julian:
See https://developer.gnome.org/documentation/tutorials/application-id.html


alatiera:
Looks good :+1:

PR #1086, opened 2022-07-23
Update docker image to Ubuntu 22.04

AaronErhardt:
Relm4 optionally supports libpanel-rs and therefore installs libpanel in it's docs CI job (which uses the gtk-rs container to reduce build times). However, libpanel requires meson 0.60 but Ubuntu 20.04 only has meson 0.59. 
Upgrading the docker image to Ubuntu 22.04 would probably solve the issue and sounds like a reasonable step anyway since it's been out for a few months already. 

It's not a big deal though, so feel free to reject this PR.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,7 +9,7 @@ name: CI
 jobs:
   check:
     name: Build
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     container:
       image: ghcr.io/gtk-rs/gtk4-rs/gtk4:latest
     strategy:
@@ -50,7 +50,7 @@ jobs:
 
   examples:
     name: examples
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     container:
       image: ghcr.io/gtk-rs/gtk4-rs/gtk4:latest
     steps:
@@ -68,7 +68,7 @@ jobs:
 
   fmt:
     name: rustfmt
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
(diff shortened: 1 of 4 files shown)


bilelmoussaoui:
> However, libpanel requires meson 0.60 but Ubuntu 20.04 only has meson 0.59.

I don't think that is the problem, we use a different container for CI, see the Dockerfile.

AaronErhardt:
I don't know what else would cause the problem though. I've tried upgrading the runner to use Ubuntu 22.04 with no success: https://github.com/Relm4/docs/pull/1
The Dockerfile uses Fedora which has meson 0.62 and the only thing that still has meson 0.59 is Ubuntu 20.04. So I concluded that this must be the source of the problem.

AaronErhardt:
I don't need this PR anymore because I've set up a custom image for Relm4 that pre-installs libpanel as well which fixes the issue and reduces down our build times even further.

That said, I wouldn't mind to help extending the gtk-rs container to support libpanel via an additional tag and drop the Relm4-specific container.

bilelmoussaoui:
> That said, I wouldn't mind to help extending the gtk-rs container to support libpanel via an additional tag and drop the Relm4-specific container.

we don't use that here, so it is fine to extend the base image we publish and build on top of it. 

> I don't know what else would cause the problem though. I've tried upgrading the runner to use Ubuntu 22.04 with no success: https://github.com/Relm4/docs/pull/1

I have triggered a rebuild of the gtk4-rs image this morning, so it should be updated to the latest available dependencies upstream.

bilelmoussaoui:
I still want to get this one merged, would you mind updating other occurrences of `ubuntu-20.04` to `ubuntu-22.04`?

bilelmoussaoui:
Any updates?

AaronErhardt:
> I still want to get this one merged, would you mind updating other occurrences of ubuntu-20.04 to ubuntu-22.04?

Sorry, I somehow missed your message. With the latest commit everything should use Ubuntu 22.04.

bilelmoussaoui:
No worries, thanks a lot!

PR #1089, opened 2022-07-27
ci: Only run book job if book was modified

bilelmoussaoui:
Avoid running unnecessary jobs as the book depends on stable release instead nowadays

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -2,9 +2,13 @@ name: book
 
 on:
   pull_request:
+    paths:
+      - "book/**"
   push:
     branches:
       - "master"
+    paths:
+      - "book/**"
   release:
     types: [published]
 


Hofer-Julian (approved):
Would be good to have it the other way around as well.

Only run the crates workflows if relevant parts of the repos has been touched and not e.g. the book or the README.

bilelmoussaoui:
we will need to have a rustfmt/clippy jobs on the book side as well right?

Hofer-Julian:
> we will need to have a rustfmt/clippy jobs on the book side as well right?

Those are handled in the book CI, now that the book is not a part of the workspace anymore: https://github.com/gtk-rs/gtk4-rs/blob/4e8b6ca611e9a2d23e096eb8f7048fd9afc8c777/.github/workflows/book.yml#L13

PR #1094, opened 2022-08-12
gtk: Make xml validation optional

### LINKED ISSUE #1018: Make the quick-xml dependency a feature
As it is not useful for applications that use gresources so I guess it should be an opt-in feature

bilelmoussaoui:
Fixes #1018

### DIFF
--- a/gtk4-macros/Cargo.toml
+++ b/gtk4-macros/Cargo.toml
@@ -14,9 +14,12 @@ rust-version = "1.57"
 [lib]
 proc-macro = true
 
+[features]
+xml_validation = ["quick-xml"]
+
 [dependencies]
 anyhow = "1.0"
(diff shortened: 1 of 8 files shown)


gtk4-macros/Cargo.toml
```diff
@@ -14,9 +14,12 @@ rust-version = "1.57"
 [lib]
 proc-macro = true
 
+[features]
+xml_validation = ["quick-xml"]
```
sdroege, line 18: Shouldn't this be the default?

bilelmoussaoui: It is only useful for xml files defined inside the source file (using `#[template (string = #'....')]`) which no ones uses in a real app. Most of people either uses a file or a gresource

PR #1098, opened 2022-08-15
gtk: Add a WidgetClassSubclassExt::install_action_async

bilelmoussaoui:
Yet another helper to get rid of usages of clone! there.

### DIFF
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -12,10 +12,9 @@ use crate::{
 use glib::subclass::SignalId;
 use glib::translate::*;
 use glib::{Cast, GString, IsA, Variant};
-use std::boxed::Box as Box_;
 use std::collections::HashMap;
 use std::fmt;
-
+use std::{boxed::Box as Box_, future::Future};
 #[derive(Debug, Default)]
 struct Internal {
     pub(crate) actions: HashMap<String, glib::ffi::gpointer>,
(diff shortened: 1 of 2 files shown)


gtk4/src/subclass/widget.rs
```diff
@@ -955,16 +954,37 @@ pub unsafe trait WidgetClassSubclassExt: ClassStruct {
...
-            + 'static,
-    >(
+    fn install_action_async<Fut, F>(
         &mut self,
         action_name: &str,
         parameter_type: Option<&str>,
         activate: F,
-    ) {
+    ) where
+        F: Fn(<<Self as ClassStruct>::Type as ObjectSubclass>::Type, &str, Option<&Variant>) -> Fut
```
bilelmoussaoui, line 963: I couldn't mange to this this to work with `&self` here

sdroege: You mean `&imp` vs. `wrapper`?

bilelmoussaoui: the `install_action` variant takes `&Self::Type` as a first param of the callback where the `install_action_async`, takes `Self::Type`. I wonder if I should make them both take `Self::Type`

sdroege: Ah so the question is just about the reference. `&T` seems more consistent

bilelmoussaoui: Yes but that fails with 
```
error: lifetime may not live long enough
  --> examples/video_player/video_player_window/imp.rs:49:49
   |
49 |               |win, _action_name, _action_target| async move {
   |  ______________---______________________________-_^
   | |              |                                |
   | |              |                                return type of closure `impl Future<Output = ()>` contains a lifetime `'2`
   | |              has type `&'1 video_player_window::VideoPlayerWindow`
50 | |                 let dialog = &win.imp().dialog;
51 | |                 dialog.set_transient_for(Some(win));
52 | |                 if dialog.run_future().await == gtk::ResponseType::Accept {
...  |
55 | |                 dialog.destroy();
56 | |             },
   | |_____________^ returning this value requires that `'1` must outlive `'2`

```
when trying to make use of it

sdroege: You need to move a `T` into the `async` block in either case

bilelmoussaoui: yes, which is something I don't have to care about if I use `T` directly instead of `&T`

bilelmoussaoui: otherwise you will end up using yet another clone! block and avoiding it is what motivated this PR

sdroege: Oh I see... because we don't have async closures yet. Nevermind then :)

sdroege (approved):

PR #1099, opened 2022-08-16
book: Address Alexander's review comments

Hofer-Julian:


book/listings/todo/6/window/imp.rs
```diff
@@ -33,7 +33,7 @@ impl ObjectSubclass for Window {
     // `NAME` needs to match `class` attribute of template
     const NAME: &'static str = "TodoWindow";
     type Type = super::Window;
-    type ParentType = adw::ApplicationWindow;
+    type ParentType = gtk::ApplicationWindow;
```
bilelmoussaoui, line 36: You no longer have a use of adw::prelude here, could switch to gtk.

PR #1104, opened 2022-08-19
ci: Only run book job if relevant dirs were modified

Hofer-Julian:
Follow up to #1089

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,7 +1,26 @@
 on:
   push:
-    branches: [master]
+    branches:
+      - "master"
+    paths:
+      - ".github/workflows/CI.yml"
+      - "examples/**"
+      - "gdk4/**"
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -1,7 +1,24 @@
 on:
   push:
-    branches: [master]
+    branches:
+      - "master"
+    paths:
```
bilelmoussaoui, line 5: changes to the ci config should also trigger a build

Hofer-Julian: True

bilelmoussaoui: I think it should exclude changes to the book job manifest though.

Hofer-Julian: I've limited it to the CI.yml now

bilelmoussaoui:
Thanks a lot!

PR #1109, opened 2022-08-25
Use mold as default linker in gtk-rs image

AaronErhardt:
Mold should yield better performance for linking while building the docker image and running gtk-rs CI jobs.

I already tested this with Relm4 and it seems to perform slightly better although the actual numbers are hard to estimate due to high deviations in general. Probably, GitLab runners will profit more from this because GitHub actions are limited to just two cores AFAIK. In any case, there shouldn't be any regressions caused by this change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,6 @@ RUN dnf update -y
 RUN dnf install git xorg-x11-server-Xvfb procps-ng wget libjpeg-turbo-devel expat-devel mold 'dnf-command(builddep)' -y
 RUN dnf builddep gtk4 -y
 
-# setup mold as default linker
-RUN ln -sf $(which mold) $(realpath /usr/bin/ld)
-
 # build gtk4
 RUN git clone https://gitlab.gnome.org/gnome/gtk.git --depth=1
 WORKDIR gtk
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
This should be done in the gtk-rs-core image first as we are looking into making the gtk4 image based on top of that one to avoid any potential differences that could cause issues (like we had with debian/fedora)

AaronErhardt:
That sounds reasonable but also means we have to fix building GTK with mold which fails currently with the following error:

``` 
[448/1113] Generating gtk/gtkresources_blob.o with a custom command
FAILED: gtk/gtkresources_blob.o 
/usr/bin/ld -z noexecstack -r -b binary gtk/gtk.gresource -o gtk/gtkresources_blob.o
mold: fatal: mold does not support `-b binary`. If you want to convert a binary file into an object file, use `objcopy -I binary -O default <input-file> <output-file.o>` instead.
```

Either we remove the use of `-b binary` in GTK or we fix it mold. The first one seems to be the better solution, I'll look into that if I have time.

bilelmoussaoui:
Closing as this should be changed in gtk-rs-core base image, which most of the bindings will end up using in the future.

PR #1113, opened 2022-09-04
Don't checkout submodules by default

sdroege:
This makes sure that cargo does not clone and checkout all the
submodules if pointing to this repository as a git dependency.

To checkout the submodules `git submodule update --checkout` can be
used.


.github/workflows/docs.yml
```diff
@@ -30,8 +30,6 @@ jobs:
   build-deploy:
     runs-on: ubuntu-22.04
     name: build
-    container:
-      image: ghcr.io/gtk-rs/gtk4-rs/gtk4:latest
```
sdroege, line 34: This is required for building, but because of https://github.com/actions/checkout/issues/915 it doesn't work correctly

sdroege:
```
Adding repository directory to the temporary git global config as a safe directory
/usr/bin/git config --global --add safe.directory /__w/gtk4-rs/gtk4-rs
[...]
Run git submodule update --checkout
fatal: detected dubious ownership in repository at '/__w/gtk4-rs/gtk4-rs'
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/gtk4-rs/gtk4-rs
```

:roll_eyes:

bilelmoussaoui:
thanks

PR #1120, opened 2022-09-13
Fixes #1114 by passing a null ptr when `colors` is empty

### LINKED ISSUE #1114: ColorChooserExtManual add_palette does not accept None even though it should
[According to the documentation](https://gtk-rs.org/gtk4-rs/git/docs/gtk4/prelude/trait.ColorChooserExtManual.html#tymethod.add_palette) it should be possible to clear the previously set palettes by passing `None` as the `colors` argument, but the type `&[RGBA]` does not allow it.

**Current behavior**
```rust
.add_palette(gtk::Orientation::Horizontal, 0, None);
```
Throws the following error:
```
mismatched types
expected reference `&[RGBA]`
        found enum `std::option::Option<_>`rustc[E0308](https://doc.rust-lang.org/error-index.html#E0308)
```
**Expected behavior**
```rust
//for clearing the palette
.add_palette(gtk::Orientation::Horizontal, 0, None);
//for adding new palettes
.add_palette(gtk::Orientation::Horizontal, 0, Some(&[gdk::RGBA::BLACK]));
```

FineFindus:
[According to the documentation,](https://gtk-rs.org/gtk4-rs/git/docs/gtk4/prelude/trait.ColorChooserExtManual.html#tymethod.add_palette) passing in `NULL`/`None` for the `colors`argument should clear the previously added palettes of the color chooser. This is not possible since `colors` has to be a `&[RGBA]` slice. 

This pull request fixes this by checking if the slice is empty and if so passing on a null pointer instead.

### DIFF
--- a/gtk4/src/color_chooser.rs
+++ b/gtk4/src/color_chooser.rs
@@ -21,7 +21,11 @@ impl<O: IsA<ColorChooser>> ColorChooserExtManual for O {
                 orientation.into_glib(),
                 colors_per_line,
                 colors.len() as c_int,
-                colors.as_ptr() as *mut gdk::ffi::GdkRGBA,
+                if colors.is_empty() {
+                    std::ptr::null_mut()
+                } else {
+                    colors.as_ptr() as *mut gdk::ffi::GdkRGBA
+                },
             )
         }
     }


sdroege (approved):

bilelmoussaoui:
Thanks!

PR #1127, opened 2022-09-22
Various updates

bilelmoussaoui:
Check the commits for details


gdk4/Cargo.toml
```diff
@@ -25,7 +25,7 @@ v4_8 = ["ffi/v4_8", "v4_6"]
 features = ["dox"]
 
 [dependencies]
-bitflags = "1.0"
+bitflags = "1.3"
```
sdroege, line 28: Why?

gtk4-macros/Cargo.toml
```diff
@@ -19,8 +19,8 @@ xml_validation = ["quick-xml"]
 
 [dependencies]
 anyhow = "1.0"
-quick-xml = {version = "0.22.0", optional = true}
-proc-macro-crate = "1.0"
+quick-xml = {version = "0.25", optional = true}
+proc-macro-crate = "1.2"
```
sdroege, line 23: Why?

gdk4/Cargo.toml
bilelmoussaoui: isn't using 1.0 != 1.3?

sdroege: cargo will automatically pull in 1.3 if 1.0 is set here. But unless we use features that were added in 1.0 (we probably don't) then keeping it 1.0 is more correct as it allows using an old version if someone really insists.

sdroege: That is, 1.3 is considered backwards compatible to 1.0.

gtk4-macros/Cargo.toml
```diff
@@ -19,8 +19,8 @@ xml_validation = ["quick-xml"]
 
 [dependencies]
 anyhow = "1.0"
-quick-xml = {version = "0.22.0", optional = true}
-proc-macro-crate = "1.0"
+quick-xml = {version = "0.25", optional = true}
```
sdroege, line 22: However this here is necessary: 0.22 and 0.25 are not considered compatible so cargo won't automatically update it (and as you saw you need to apply some source changes too).

PR #1131, opened 2022-09-29
examples: Add a GIF paintable

zecakeh:
This is based on #544 which was based on #152.

I believe I addressed all the concerns in the previous PRs except maybe the one about the GIF drifting, but I'm not sure if we should drop frames or not.


bilelmoussaoui:
Don't mind the CI failure, I will finish up #1127 which should fix the CI.

examples/gif_paintable/gif_paintable/imp.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    fn new() -> Self {
+        Self {
+            next_frame: RefCell::new(None),
+            frames: RefCell::new(None),
+            width: Cell::new(Some(0)),
+            height: Cell::new(Some(0)),
+            current_idx: Cell::new(0),
+        }
+    }
```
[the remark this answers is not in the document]
sdroege: Deriving `Default`, yes

examples/gif_paintable/gif_paintable/imp.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn intrinsic_height(&self, _paintable: &Self::Type) -> i32 {
+        self.height.get().unwrap_or(-1)
+    }
+
+    fn intrinsic_width(&self, _paintable: &Self::Type) -> i32 {
+        self.width.get().unwrap_or(-1)
+    }
+
+    fn snapshot(&self, _paintable: &Self::Type, snapshot: &gdk::Snapshot, width: f64, height: f64) {
+        if let (Some(texture), Some(width), Some(height)) = (
```
sdroege, line 46: Maybe either: width/height/texture could be stored in the same `Option`, or you could simply take `texture.width()` / `height()`? The latter is probably nicer, you currently have these stored redundantly.

examples/gif_paintable/gif_paintable/imp.rs
```diff
@@ -0,0 +1,62 @@
...
+            self.next_frame.borrow_mut().take(),
+            self.width.take(),
+            self.height.take(),
```
sdroege, line 49: Snapshotting would now give the correct result only on the first call and future calls would give black until the next frame is selected. Snapshotting here probably shouldn't be destructive but always just render the latest selected frame.

examples/gif_paintable/gif_paintable/frame.rs
```diff
@@ -0,0 +1,53 @@
+use std::time::Duration;
+
+use gtk::prelude::*;
+use gtk::{gdk, glib};
+
+pub struct Frame {
+    pub width: i32,
+    pub height: i32,
+    pub texture: gdk::Paintable,
```
sdroege, line 9: Store it as a `gdk::Texture` and you can get width/height from it instead of storing it separately

examples/gif_paintable/gif_paintable/frame.rs
```diff
@@ -0,0 +1,53 @@
...
+        let image = f.into_buffer();
+
+        let samples = image.into_flat_samples();
+        let (stride, width, height) = samples.extents();
+
+        // glib::Bytes must contain gtk_stride * height number of bytes.
+        // since each pixel in the gif frame contains `stride` bytes
+        // of information we must map this stride to a value that gdk
+        // understands (see https://gtk-rs.org/gtk4-rs/git/docs/gdk4/struct.MemoryTexture.html)
+        let gtk_stride = stride * width;
```
sdroege, line 31: This would be `samples.layout.width_stride` FWIW. It's not guaranteed to be pixel/sample stride * width.

examples/gif_paintable/gif_paintable/frame.rs
```diff
@@ -0,0 +1,53 @@
...
+
+        // convention is to use 100 milliseconds duration if it is defined as 0.
+        if frame_duration.is_zero() {
+            frame_duration = Duration::from_millis(100);
+        }
+
+        let image = f.into_buffer();
+
+        let samples = image.into_flat_samples();
+        let (stride, width, height) = samples.extents();
```
sdroege, line 25: You probably want to get `samples.layout.channel_stride` here but see below

examples/gif_paintable/gif_paintable/frame.rs
```diff
@@ -0,0 +1,53 @@
...
+        let mut frame_duration = Duration::from(f.delay());
+
+        // convention is to use 100 milliseconds duration if it is defined as 0.
+        if frame_duration.is_zero() {
+            frame_duration = Duration::from_millis(100);
+        }
+
+        let image = f.into_buffer();
+
+        let samples = image.into_flat_samples();
```
sdroege, line 24: I think you need to ensure somewhere that this is actually `P: RGBA` and not anything else?

examples/gif_paintable/gif_paintable/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+        let read = Cursor::new(bytes);
+        let decoder = GifDecoder::new(read)?;
+        let frames = decoder
+            .into_frames()
+            .collect_frames()?
+            .into_iter()
+            .map(Frame::new)
+            .collect::<Vec<Frame>>();
+
+        imp.frames.replace(Some(frames));
```
sdroege, line 40: Could also store the iterator directly so that you decode frames one by one as needed?

examples/gif_paintable/gif_paintable_window/imp.rs
```diff
@@ -0,0 +1,62 @@
...
+        gif_filter.add_mime_type("image/gif");
+        gif_filter.set_name(Some("GIF Image"));
+
+        let dialog = gtk::FileChooserNative::builder()
+            .title("Open File")
+            .action(gtk::FileChooserAction::Open)
+            .accept_label("Open")
+            .cancel_label("Cancel")
+            .modal(true)
+            .filter(&gif_filter)
```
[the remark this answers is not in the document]
zecakeh: In this case, filtering works when I launch the program with `cargo run`. The only difference is that using `add_filter` adds a dropdown with a single entry.

examples/gif_paintable/gif_paintable_window/imp.rs
```diff
@@ -0,0 +1,62 @@
...
+        klass.install_action_async(
+            "win.open",
+            None,
+            |win, _action_name, _action_target| async move {
+                let dialog = &win.imp().dialog;
+                dialog.set_transient_for(Some(&win));
+                if dialog.run_future().await == gtk::ResponseType::Accept {
+                    win.set_file(dialog.file().unwrap());
+                }
+                dialog.destroy();
```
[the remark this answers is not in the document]
sdroege: Ah good. So it just works again next time?

examples/gif_paintable/gif_paintable/frame.rs
zecakeh: What do you mean? `Frame::into_buffer()` returns an `RgbaImage` so we already know that `P: Rgba`.

sdroege: Ah I didn't realize it would return an `RgbaImage`. All good then, I assumed it's returning a generic thing.

examples/gif_paintable/gif_paintable/mod.rs
zecakeh: If we rely only on the iterator, it won't be possible to loop the animation once all the frames have been loaded because we will have consumed the iterator.

Should we load each frame as needed the first time and then use the loaded frames for the subsequent loops?

sdroege: Seems overly complicated, let's just keep it like this but maybe add a comment. In a real application you probably don't want to do this because you could easily end up loading a small GIF that decodes to multiple GB of frames :)

examples/gif_paintable/gif_paintable_window/imp.rs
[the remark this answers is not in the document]
bilelmoussaoui: yes, similar to the video player example

examples/gif_paintable/gif_paintable/mod.rs
zecakeh: We could recommend to use `GifDecoder::with_limits()`, that allows to limit the memory allocated by the GIF, or event use it in the example with a sane value.

sdroege: Sounds good to me

examples/gif_paintable/gif_paintable/frame.rs
zecakeh: Don't we just want to use `samples.layout.height_stride`? which would be the stride to the next row, aka rowstride?

sdroege: Erm yes. `height_stride` is what I meant, not `width_stride`.

zecakeh:
I added a README an screenshot since it seems like all other examples do that.

examples/gif_paintable/gif_paintable/frame.rs
```diff
@@ -0,0 +1,38 @@
...
+use gtk::prelude::*;
+use gtk::{gdk, glib};
+
+pub struct Frame {
+    pub texture: gdk::Texture,
+    pub frame_duration: Duration,
+}
+
+impl Frame {
+    pub fn new(f: image::Frame) -> Self {
```
bilelmoussaoui, line 12: Can't this be a `impl From<image::Frame> for Frame` instead?

bilelmoussaoui: 😅🤣

bilelmoussaoui:
glib::Object constructors needs fixes otherwise lgtm

PR #1133, opened 2022-10-03
book: Update step on getting gtk4 crate version

kianmeng:
Use placeholder value that follows the semantic versioning and find gtk4 crate version using `crate search`.

### DIFF
--- a/book/src/project_setup.md
+++ b/book/src/project_setup.md
@@ -1,34 +1,52 @@
 # Project Setup
 
-Let's begin by installing all necessary tools.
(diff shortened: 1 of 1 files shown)


Hofer-Julian:
Thanks for your contribution @kianmeng .

There is indeed room for improvement, but I think there is a better alternative to your PR.
We should instead replace all cases where dependencies are introduced with calls of [cargo add](https://doc.rust-lang.org/cargo/commands/cargo-add.html).

Are you interested in picking this up?
If yes, make sure to disable your markdown formatter to avoid big diffs.

kianmeng:
@Hofer-Julian Sounds good. Should I close this PR and create a new one?

PR #1135, opened 2022-10-06
Fix docs.rs builds

pentamassiv:
Analogous to: https://github.com/gtk-rs/gtk-rs-core/pull/778

Whenever a crate is published on crates.io, its documentation is automatically built on docs.rs regardless of the documentation field in the Cargo.toml file. On docs.rs it is possible to look at all previous versions. This is why there is value in having working builds on docs.rs. Having broken builds on docs.rs also prevents crates using that library from providing their own documentation on docs.rs. I enabled the dox feature wherever it is available to ensure the docs always build, even if the dependencies are not available.

In order to pass the build Actions, I also fixed some clippy lints.

### DIFF
--- a/examples/custom_buildable/custom_buildable/imp.rs
+++ b/examples/custom_buildable/custom_buildable/imp.rs
@@ -59,7 +59,7 @@ impl BuildableImpl for CustomBuildable {
         } else if Some("prefix") == type_ {
             // Check if the child was added using `<child type="prefix">`
             buildable.add_prefix(child.downcast_ref::<gtk::Widget>().unwrap());
-        } else if None == type_ {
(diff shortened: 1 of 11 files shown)


pentamassiv:
https://github.com/gtk-rs/gtk-rs-core/pull/778#discussion_r989132506
I guess I have to teach it to gir for gtk4 too

pentamassiv:
Bilal fixed it with his commit [b8f02dc9604080d92c302821d06f835157298f81
](https://github.com/gtk-rs/gtk4-rs/commit/b8f02dc9604080d92c302821d06f835157298f81)
Thank you very much!

PR #1142, opened 2022-10-09
Regen with latest gir-files

bilelmoussaoui:
Adds gtk4.10 APIs support


bilelmoussaoui:
The CI will fail until the image is build & updated

PR #1152, opened 2022-10-18
 gtk: Automatically generate Snapshot::to_(node|paintable) methods 

bilelmoussaoui:


bilelmoussaoui:
> Probably also for `window.close()`?

it is already done

PR #1157, opened 2022-10-23
Move from `imp.instance()` to `imp.obj()`

sdroege:
It's doing the same thing and is shorter.

### DIFF
--- a/book/listings/actions/7/window/imp.rs
+++ b/book/listings/actions/7/window/imp.rs
@@ -45,10 +45,10 @@ impl ObjectImpl for Window {
         self.parent_constructed();
 
         // Setup
-        let instance = self.instance();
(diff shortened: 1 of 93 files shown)


GuillaumeGomez (approved):

book/listings/actions/7/window/imp.rs
```diff
@@ -45,7 +45,7 @@ impl ObjectImpl for Window {
         self.parent_constructed();
 
         // Setup
-        let instance = self.instance();
+        let instance = self.obj();
```
Hofer-Julian, line 48: Could you please rename this and similar cases to obj?

sdroege: Sure, makes sense. Done!

PR #1160, opened 2022-10-25
gtk: Add gnome_43 feature

### LINKED ISSUE #1156: [FEATURE REQUEST] Add gnome_{version} feature flags
### The problem

The gtk4 crate re-exports quite a few dependencies. Yet at the same time, these dependencies have quite a few items behind a certain feature flag. This makes the re-export less useful because to activate these feature flags, you need to specify the re-exports like regular dependencies in the README.md.

One particularly common situation are version requirements. You might need to enable features from GTK 4.8, glib 2.70 and also gdk-pixbuf 2.36. This means that you need to re-define two crates as dependencies that were already re-exports.

For example, [Amberol](https://gitlab.gnome.org/World/amberol/-/blob/main/Cargo.toml) is affected by this.

### The solution

Developers usually target a certain version of the entire GNOME stack, which means a certain baseline of dependencies. For example, for GNOME 43 that should be GTK 4.8, libadwaita 1.2, glib 2.74 and so on. Anyone targeting Ubuntu 22.10, Fedora 37 or the Flatpak GNOME runtime in version 43 should have this baseline of dependencies. 

So naturally, it would make sense to also offer this as a feature flag for gtk4-rs. For example, enabling the "gnome_43" feature flag could take care enabling the correct feature flags.

AaronErhardt:
Fixes #1156

This only adds the feature for GNOME 43, but I think going with the latest SDK version is fine.
Of course, new features should be added as soon as newer GNOME versions become available.

### DIFF
--- a/gtk4/Cargo.toml
+++ b/gtk4/Cargo.toml
@@ -25,6 +25,11 @@ v4_10 = ["ffi/v4_10", "v4_8"]
 xml_validation = ["gtk4-macros/xml_validation"]
 unsafe-assume-initialized = []
 
+# Versions from https://gitlab.gnome.org/GNOME/gnome-build-meta/-/tree/gnome-43/elements/sdk
+# gtk takes care of setting the versions of gsk and gdk
+# gio takes care of setting the version of glib
+gnome_43 = ["v4_8", "cairo-rs/v1_16", "gdk-pixbuf/v2_42", "gio/v2_74", "pango/v1_50"]
+
 [package.metadata.docs.rs]
 features = ["dox"]
 
--- a/gtk4/Cargo.toml
+++ b/gtk4/Cargo.toml
@@ -29,6 +29,7 @@ unsafe-assume-initialized = []
 # gtk takes care of setting the versions of gsk and gdk
 # gio takes care of setting the version of glib
 gnome_43 = ["v4_8", "cairo-rs/v1_16", "gdk-pixbuf/v2_42", "gio/v2_74", "pango/v1_50"]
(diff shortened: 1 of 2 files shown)


gtk4/Cargo.toml
```diff
@@ -25,6 +25,11 @@ v4_10 = ["ffi/v4_10", "v4_8"]
 xml_validation = ["gtk4-macros/xml_validation"]
 unsafe-assume-initialized = []
 
+# Versions from https://gitlab.gnome.org/GNOME/gnome-build-meta/-/tree/gnome-43/elements/sdk
+# gtk takes care of setting the versions of gsk and gdk
+# gio takes care of setting the version of glib
+gnome_43 = ["v4_8", "cairo-rs/v1_16", "gdk-pixbuf/v2_42", "gio/v2_74", "pango/v1_50"]
```
bilelmoussaoui, line 31: what about glib?

bilelmoussaoui: we should also look at adding 42 tbh, as it is commonly used still.

AaronErhardt: > what about glib?

gio takes care of setting the version of glib. If I understand it correctly, gio is also part of GLib (as C library) or at least they share the same version. Since gio depends on glib, setting the version there also sets it for glib.

Alright, I'll add GNOME 42 too :)

bilelmoussaoui: > gio takes care of setting the version of glib. If I understand it correctly, gio is also part of GLib (as C library) or at least they share the same version. Since gio depends on glib, setting the version there also sets it for glib.

Ah that is true, never mind my comment then.

alatiera:
Looks good :+1:

PR #1162, opened 2022-10-26
image: Rebuild once every week

AaronErhardt:
Same as https://github.com/gtk-rs/gtk-rs-core/pull/817, but for the gtk4-rs image.

This will likely fix the Relm4 CI and the Relm4 docs build so merging both PRs fast would be highly appreciated (please merge the gtk-rs-core PR first so the gtk4-rs image can then pick up the freshly built core image).

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -1,6 +1,9 @@
 name: github packages
 
 on:
+  # Rebuild the container once every week
+  schedule:
+  - cron: '0 1 * * 1'
   push:
     branches:
       - "master"


Dockerfile
```diff
@@ -14,4 +14,4 @@ RUN git clone https://gitlab.gnome.org/gnome/gtk.git --depth=1 && \
     (cd /libadwaita && \
         meson setup builddir --prefix=/usr --buildtype release -Dintrospection=disabled -Dvapi=false -Dexamples=false -Dtests=false && \
         meson install -C builddir) && \
-    rm -rf /gtk /libadwaita
+    rm -rf /gtk /libadwaita
```
bilelmoussaoui, line 17: Please revert the change on this file

PR #1166, opened 2022-10-29
book: Use `Object::builder()` instead of `Object::new()`

Hofer-Julian:


book/listings/composite_templates/4/custom_button/mod.rs
```diff
@@ -11,7 +11,7 @@ glib::wrapper! {
 
 impl CustomButton {
     pub fn new() -> Self {
-        Object::new(&[])
+        Object::builder().build()
```
bilelmoussaoui, line 14: I wonder if we can have Default implement for Object?

sdroege: You can't do a blanket implementation of a trait outside the crate. We could add a `default()` constructor though, or `impl Default for SpecificObject` via the `wrapper!` macro.

bilelmoussaoui: I meant in gtk-rs-core itself, but i suppose that would conflict with people having their own default implementation?

sdroege: We don't define the `Default` trait :)

PR #1174, opened 2022-11-07
gdk-win32: implement Win32Display.add_filter()

elmarco:
Not really happy about the result, comments & suggestions welcome!

### DIFF
--- a/gdk4-win32/src/win32_display.rs
+++ b/gdk4-win32/src/win32_display.rs
@@ -1,6 +1,7 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use crate::Win32Display;
+use crate::{ffi, glib, Win32Display, Win32MessageFilterReturn, MSG};
+use glib::translate::*;
 
 #[cfg(any(all(feature = "v4_4", feature = "egl"), feature = "dox"))]
 #[cfg_attr(feature = "dox", doc(cfg(all(feature = "v4_4", feature = "egl"))))]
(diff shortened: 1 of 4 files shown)


gdk4-win32/src/win32_display.rs
```diff
@@ -24,4 +26,64 @@ impl Win32Display {
...
+    where
+        F: Fn(&Self, &mut MSG, &mut i32) -> Win32MessageFilterReturn + 'static,
+    {
+        let f = Box::into_raw(Box::new(filter_func)) as *mut _;
+        unsafe {
+            ffi::gdk_win32_display_add_filter(self.to_glib_none().0, Some(trampoline::<F>), f);
+        }
+
+        Win32DisplayFilterHandle {
+            display: self.clone(),
```
elmarco, line 41: If we use this approach for handler/drop, I would rather keep a weak reference to self. How to do that?

gdk4-win32/src/win32_display.rs
```diff
@@ -24,4 +26,69 @@ impl Win32Display {
             }
         }
     }
+
+    #[doc(alias = "gdk_win32_display_add_filter")]
+    pub fn add_filter<F>(&self, filter_func: F) -> Win32DisplayFilterHandle<F>
```
sdroege, line 31: This will be impossible to store in a struct because of the type parameter. Can you remove it and simply store a `Box<dyn Fn ...>` instead?

elmarco: ok, lets try that, not sure it's correct. please check! :)

gdk4-win32/src/win32_display.rs
```diff
@@ -24,4 +25,71 @@ impl Win32Display {
...
+        &mut *return_value,
+    )
+    .into_glib()
+}
+
+pub struct Win32DisplayFilterHandle {
+    display: glib::WeakRef<Win32Display>,
+    func: ffi::GdkWin32MessageFilterFunc,
+    box_: glib::ffi::gpointer,
+    drp: Box<dyn Fn()>,
```
sdroege, line 79: You're doing it wrong currently (you're not creating a `Box<dyn Fn...>` but... why don't you just store the free-function here instead? That could be a plain `fn(*mut c_void)` and doesn't need any type parameters.

elmarco: ok, check the result please :)

gdk4-win32/src/win32_display.rs
```diff
@@ -24,4 +25,74 @@ impl Win32Display {
...
+        Win32DisplayFilterHandle {
+            display,
+            func,
+            box_,
+            drop_,
+        }
+    }
+}
+
+unsafe extern "C" fn trampoline<
```
bilelmoussaoui, line 56: The trampoline is only used inside add_filter, so it should be moved into the function itself. That is how we do it everywhere

gdk4-win32/src/win32_display.rs
```diff
@@ -24,4 +25,74 @@ impl Win32Display {
...
+    let f: &F = &*(box_ as *const F);
+    f(
+        &from_glib_borrow(display as *mut ffi::GdkWin32Display),
+        &mut *(msg as *mut MSG),
+        &mut *return_value,
+    )
+    .into_glib()
+}
+
+/// An owned `Win32Display` filter.
```
bilelmoussaoui, line 73:
```suggestion
// rustdoc-stripper-ignore-next
/// An owned `Win32Display` filter.
```
otherwise the CI will fail

gdk4-win32/src/win32_display.rs
sdroege: Looks correct. Does it work? :)

elmarco: yep, works for me :)

bilelmoussaoui (approved):

PR #1179, opened 2022-11-12
macros: Allow using re-exports of gtk

AaronErhardt:
I don't think it's a great idea to simply panic if there's no gtk4 dependency in Cargo.toml because crates like libadwaita or Relm4 re-export gtk. Instead, I suggest to fallback to "gtk" for re-exports which still results in a good error message when the dependency is missing. 
This shouldn't be a semver breaking change because existing code will compile as usual.


bilelmoussaoui:
cc @jf2048

jf2048:
LGTM

PR #1188, opened 2022-11-25
Update dependencies

sdroege:


sdroege:
Hm that test failure is strange. It was passing fine here for me

sdroege:
Can you reproduce the failure?

bilelmoussaoui:
> Can you reproduce the failure?

it is probably caused by a gtk change

sdroege:
Let's get this in then and handle the failure separately. It also happens otherwise in git currently.

PR #1191, opened 2022-11-26

bilelmoussaoui:


gtk4/src/font_dialog.rs
```diff
@@ -0,0 +1,111 @@
...
+use glib::IsA;
+use std::boxed::Box as Box_;
+use std::pin::Pin;
+use std::ptr;
+
+impl FontDialog {
+    #[doc(alias = "gtk_font_dialog_choose_font_and_features")]
+    #[doc(alias = "gtk_font_dialog_choose_font_and_features_finish")]
+    pub fn choose_font_and_features<
+        P: FnOnce(Result<(pango::FontDescription, glib::GString, pango::Language), glib::Error>)
```
sdroege, line 14: Is it really guaranteed that either all 3 are NULL and the error is set, or none of them are NULL?

gtk4/src/auto/string_list.rs
```diff
@@ -25,14 +22,6 @@ impl StringList {
         unsafe { from_glib_full(ffi::gtk_string_list_new(strings.to_glib_none().0)) }
     }
 
-    // rustdoc-stripper-ignore-next
-    /// Creates a new builder-pattern struct instance to construct [`StringList`] objects.
-    ///
-    /// This method returns an instance of [`StringListBuilder`](crate::builders::StringListBuilder) which can be used to create [`StringList`] objects.
-    pub fn builder() -> StringListBuilder {
```
sdroege: Instead you might want to implement the `Extend` and `FromIterator` traits here

bilelmoussaoui: they are already implemented

sdroege: Oops

sdroege:
Looks fine to me otherwise :)

gtk4/src/font_dialog.rs
bilelmoussaoui: Right, FontDescription can be NULL, the two others should always be set if error is NULL.

PR #1193, opened 2022-11-27
Backport to 0.5

bilelmoussaoui:


gdk4/src/auto/enums.rs
```diff
@@ -264,6 +272,14 @@ impl ToValue for CrossingMode {
     }
 }
 
+impl From<CrossingMode> for glib::Value {
```
sdroege, line 275: We don't want these things backported. That's something for the next release. Adding so many trait impls (and also without the gtk-rs-core parts of it!) is dangerous

sdroege: I.e. use the latest gir from the 0.16 branch here, not the master branch

sdroege:
I'll review again once the gir noise is gone :)

gdk4/src/auto/enums.rs
bilelmoussaoui: > We don't want these things backported. That's something for the next release. Adding so many trait impls (and also without the gtk-rs-core parts of it!) is dangerous

oh i thought that was backported, my bad

sdroege: It needed quite some changes in gstreamer-rs so it continued compiling :)

bilelmoussaoui: I went the lazy way and added an extra commit for switching to 0.16 branch & regenerating

gdk4-win32/Cargo.toml
```diff
@@ -27,13 +27,13 @@ egl = ["khronos-egl"]
...
-ffi = {path = "./sys", package = "gdk4-win32-sys", version = "0.5.2"}
-gdk = {path = "../gdk4", package = "gdk4", version = "0.5.2"}
+ffi = {path = "./sys", package = "gdk4-win32-sys", version = "0.5.3"}
+gdk = {path = "../gdk4", package = "gdk4", version = "0.5.3"}
 gio = {git = "https://github.com/gtk-rs/gtk-rs-core", branch = "0.16", version = "0.16", features = ["v2_66"]}
 glib = {git = "https://github.com/gtk-rs/gtk-rs-core", branch = "0.16", version = "0.16", features = ["v2_66"]}
 libc = "0.2"
 khronos-egl = {version = "4.1.0", optional = true}
-windows = { version = "0.41.0", features = [
+windows = { version = "0.43.0", features = [
```
sdroege, line 36: Does this crate export types from the windows crate? If so this would be a breaking change.

gdk4-wayland/Cargo.toml
```diff
@@ -23,8 +23,8 @@ xkb_crate = ["xkb"]
 features = ["dox"]
 
 [dependencies]
-ffi = {path = "./sys", package = "gdk4-wayland-sys", version = "0.5.2"}
-gdk = {path = "../gdk4", package = "gdk4", version = "0.5.2"}
+ffi = {path = "./sys", package = "gdk4-wayland-sys", version = "0.5.3"}
+gdk = {path = "../gdk4", package = "gdk4", version = "0.5.3"}
```
sdroege, line 27: Is this needed?

sdroege: (same elsewhere)

sdroege:
Seems good otherwise

gdk4-wayland/Cargo.toml
bilelmoussaoui: it is not actually but too lazy to fix it :P

gdk4-win32/Cargo.toml
bilelmoussaoui: it does, only if you opt-in for the windows feature, otherwise we provide our own types. I think @elmarco is the only one using this crate anyway, would it be fine to you to backport this change?

gdk4-wayland/Cargo.toml
sdroege: Let's not do that for 0.6 then :)

gdk4-win32/Cargo.toml
sdroege: Seems ok to me if @elmarco is happy with it

bilelmoussaoui: Alright, let me drop it :)

bilelmoussaoui: Hmmm, I am sure i didn't backport that commit

bilelmoussaoui: see https://github.com/gtk-rs/gtk4-rs/blob/0.5/gdk4-win32/Cargo.toml#L36

PR #1194, opened 2022-11-27
use Into<Value> and Into<Variant> where possible

jf2048:


gtk4/src/expression.rs
```diff
@@ -313,5 +379,16 @@ macro_rules! define_expression {
                 value
             }
         }
+
+        impl From<$rust_type> for glib::Value {
+            fn from(e: $rust_type) -> Self {
+                skip_assert_initialized!();
+                let mut value = glib::Value::for_value_type::<$rust_type>();
+                unsafe {
+                    ffi::gtk_value_set_expression(value.to_glib_none_mut().0, e.into_glib_ptr())
```
sdroege, line 388: This will leak the pointer

jf2048: Are you sure? This is using `into_glib_ptr` like the rest

sdroege: Yes, but it's `set_expression()` and not `take_expression()`

jf2048: Oh right :)

sdroege (approved):

PR #1195, opened 2022-11-28
gtk: Fix missing version guards

bilelmoussaoui:
We should look at adding the missing since annotations upstream for next release


sdroege (approved):

bilelmoussaoui:
Fixed the missing since annotations upstream https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/5271

Issue #1198, opened 2022-12-03
[BUG] Building with gvsbuild GTK4 no longer works with gvsbuild==2022.5.0: (LoadLibraryExW)


mitchhentges:
**Bug description**

1. [Build GTK4 with `gvsbuild==2022.5.0`](https://github.com/wingtk/gvsbuild#build-gtk)
2. Within `gtk4-rs`, check out `0.5.2` (though I tested with current master, `5a8e72b4d13bdf774b1aa479338fd08074f61a4e`)
3. `cargo build`.

```
C:\dev\gtk4-rs [master]> cargo clippy
...
    Checking gtk4 v0.6.0 (C:\dev\gtk4-rs\gtk4)
error: LoadLibraryExW failed
  --> gtk4\src\lib.rs:29:1
   |
29 | pub extern crate gtk4_macros;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: cannot determine resolution for the import
  --> gtk4\src\auto\functions.rs:23:5
   |
23 | use glib::translate::*;
   |     ^^^^^^^^^^^^^^^^^^
   |
   = note: import resolution is stuck, try simplifying other imports

...
```

Two other things worth mentioning:
* This works with the last release of `gvsbuild` (`2022.4.1`)
* I can still build a C-based GTK4 project in Visual Studio successfully

bilelmoussaoui:
Definitely not a gtk4-rs bug.

mitchhentges:
You're very likely right, but I'm not sure how to dig into the issue further (`cargo build` in the `gtk4_macros` subdir works), and building a regular C GTK app works.
So, my thoughts behind reporting it here are:
* _Maybe_ there's an issue with `gtk4-rs`'s way that it binds with the underlying GTK library, and
* Since `gvsbuild` is recommended in the GTK docs, other people may run into the issue with `gtk4-rs` specifically.

sdroege:
I think the problem here is that loading the proc-macro library into the compiler fails, most likely because a DLL is not found at build time. I don't know how the exact file is called on Windows, but can you check `target/debug/deps/libgtk4_macros-XXXXXXXXXX.dll` or similar to see if there are missing DLLs? You can use https://github.com/lucasg/Dependencies for that.

Can you also check if this also happens after removing the `gtk` dev-dependency from `gtk4-macros/Cargo.toml` (second last line)?

[...]

PR #1203, opened 2022-12-04
Recommend (with instructions) building with `gvsbuild`

### LINKED ISSUE #1199: [FEATURE REQUEST] Update MSVC installation docs to recommend gvsbuild
I think that `gvsbuild` is an easier way for developers to build GTK4 on Windows than the current recommended steps.

* The upstream GTK docs [recommend gvsbuild for building with MSVC](https://www.gtk.org/docs/installations/windows/).
* It would allow significantly simplifying the `installation_windows` and `libadwaita#windows` (only `gvsbuild build libadwaita`, nice!) sections.
* (👎) [There are risks with icons not working](https://gitlab.gnome.org/GNOME/gtk/-/issues/5303)
* ~(👎) [there is currently a compatibility issue with the most-recent `gvsbuild` release and `gtk4-rs`](https://github.com/gtk-rs/gtk4-rs/issues/1198)~

I will happily make the PR to migrate the docs for this, if you think that the tradeoffs are worth making this change.

mitchhentges:
Updates Windows build instructions for `libadwaita` as well.

Keep manual `Path`-modification steps because
the `gvsbuild` docs only mention how to change the `Path` for the
current shell.

This workflow has a few advantages:
* It's more closely synced with upstream GTK docs (which recommends
  `gvsbuild`)
* There's less steps for users to follow, so there should be less
  risk-of-error
* We get the "stable version combination management" provided by
  `gvsbuild` for free. Before, there was risk of instability
  due to the usage of the tips of `gtk`/`libxml2`/`librsvg`. Now,
  new users get new code based on releases of `gvsbuild`, yet
  aren't too close to the cutting edge
* There's no less directories cluttering `C:\` (just `C:\gtk-build`)
  and less `Path` modifications

Fixes #1199

### DIFF
--- a/book/src/installation_windows.md
+++ b/book/src/installation_windows.md
@@ -21,29 +21,42 @@ Also set the rust toolchain back to msvc by executing:
 rustup default stable-msvc
 ```
 
-### Visual Studio
+### Build GTK 4
+
+1. Follow [the `gvsbuild` docs to build GTK4](https://github.com/wingtk/gvsbuild#development-environment)
+   * When choosing the GTK version to build, select `gtk4` instead of `gtk3`: `gvsbuild build gtk4`
+2. Update your `Path` environment variable to include the GTK4 libraries:
+   1. Go to settings -> Search and open `Advanced system settings` -> Click on `Environment variables`
+   2. Select `Path` -> Click on `Edit` -> Add `C:\gtk-build\gtk\x64\release\bin`
+
+#### Build GTK 4 manually instead
+
+If it's not possible to build with `gvsbuild` (or you want to customize your build), you
+can build GTK 4 and the minimum dependencies you need manually:
+
+##### Visual Studio
 
 Install Visual Studio Community from [visualstudio.microsoft.com](https://visualstudio.microsoft.com/de/vs/community/).
 Make sure to check the box "Desktop development with C++" during the installation process.
 
 <div style="text-align:center"><img src="img/vs-install.png" /></div>
 
-### Git
+##### Git
 
 Download git from [gitforwindows.org](https://gitforwindows.org/).
 
 
-### CMake
+##### CMake
 Download CMake from [https://cmake.org/download/](https://cmake.org/download/)
 
 
-### Python
+##### Python
 
 Download python from [python.org](https://www.python.org/downloads).
 Make sure to opt-in to adding Python to your Path during the installation process.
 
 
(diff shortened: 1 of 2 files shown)


mitchhentges:
Marking this as `draft`, as this doesn't currently work due to #1198.

Hofer-Julian:
Thanks @mitchhentges

I see that MSYS2 is required.
Does that then also mean that you have to switch to the GNU toolchain?

Hofer-Julian:
Before we can merge this, the libadwaita section would need to be adapted as well

sdroege:
> Does that then also mean that you have to switch to the GNU toolchain?

No, it's building GTK and stuff with MSVC. msys2 is only needed AFAIU for crappy autotools build systems of some dependencies so that some kind of POSIX-compatible shell is available.

book/src/installation_windows.md
```diff
@@ -21,98 +21,16 @@ Also set the rust toolchain back to msvc by executing:
...
-cd librsvg/win32
-where python
-nmake /f generate-msvc.mak generate-nmake-files PYTHON=<output from last command>
-xcopy /s C:\gnome\include\cairo C:\gnome\include
-nmake /f Makefile.vc CFG=release install PREFIX=C:\gnome
-cd /
-```
+### Build GTK4
+
+1. Follow [the `gvsbuild` docs to build GTK4](https://github.com/wingtk/gvsbuild#development-environment)
```
Hofer-Julian, line 26: Please mention that they need to install GTK 4 not GTK 3.

mitchhentges:
Currently finding the minimal `gvsbuild` usage and follow-up commands that work around that icon-related bug, will de-`[WIP]` afterwards.

mitchhentges:
> #1198 is no longer a problem with the newer version of `gvsbuild`, so I'm opening this for review. I'm seeing a lot of discussion in this PR's [associated issue](https://github.com/gtk-rs/gtk4-rs/issues/1199) - once a conclusion is reached, just request the changes in this PR and I'll update it accordingly 👍

I'm guessing [based on the status of the conversation at this point](https://github.com/gtk-rs/gtk4-rs/issues/1199#issuecomment-1348917635) that we indeed want to keep the old workflow. I'll update this PR accordingly soon.

Hofer-Julian:
Thanks, I will add afterwards a couple of things on top of these changes.

PR #1212, opened 2022-12-09
gtk: Cleanup template related functions

### LINKED ISSUE #1189: Improve bindings for gtk::Widget::dispose_template
Currently one has to call
```rust
       fn dispose(&self) {
            self.obj().dispose_template(Self::Type::static_type());
        }
```
but this could be turn into
```rust
        fn dispose(&self) {
            self.dispose_template();
        }
```

bilelmoussaoui:
Fixes #1189

### DIFF
--- a/gtk4/src/auto/widget.rs
+++ b/gtk4/src/auto/widget.rs
@@ -429,9 +429,6 @@ pub trait WidgetExt: 'static {
     #[doc(alias = "gtk_widget_in_destruction")]
     fn in_destruction(&self) -> bool;
 
-    #[doc(alias = "gtk_widget_init_template")]
-    fn init_template(&self);
-
     #[doc(alias = "gtk_widget_insert_action_group")]
     fn insert_action_group(&self, name: &str, group: Option<&impl IsA<gio::ActionGroup>>);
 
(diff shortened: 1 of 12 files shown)


sdroege:
That's of course not backwards compatible so don't backport that :)

bilelmoussaoui:
> That's of course not backwards compatible so don't backport that :)

yes, that is why i didn't add the corresponding label

gtk4/src/auto/widget.rs
```diff
@@ -372,10 +372,6 @@ pub trait WidgetExt: 'static {
     #[doc(alias = "get_style_context")]
     fn style_context(&self) -> StyleContext;
 
-    #[doc(alias = "gtk_widget_get_template_child")]
-    #[doc(alias = "get_template_child")]
-    fn template_child(&self, widget_type: glib::types::Type, name: &str) -> glib::Object;
```
sdroege, line 382: Doesn't hurt to have it, does it?

bilelmoussaoui: There is no point in having it, you can retrieve child widgets with `#[template_child]`. This is probably a helper for languages like Python

PR #1213, opened 2022-12-09
Regenerate with latest gir

bilelmoussaoui:


gtk4/src/expression.rs
```diff
@@ -23,6 +23,14 @@ pub unsafe trait IsExpression:
 }
 
 impl Expression {
+    pub fn upcast(self) -> Self {
```
bilelmoussaoui, line 26: Really not happy with this :(

sdroege: Well, theoretically you should have a trait that implements this on all expression types (including `Expression`), but until then having it as a method on `Expression` (like on all other expression types) would be symmetric at least.

bilelmoussaoui: We already have those methods on a trait actually, but it is only implemented for the sub-types. I will see what can be done here

bilelmoussaoui: The problem with going through a trait is auto-generated bindings won't include it :( so we really need to start importing `use glib::prelude::*;` & `use $crate::prelude::*;` instead from gir side.

sdroege: `use $crate::prelude::*` should already include the `glib` prelude or not?

bilelmoussaoui: it should indeed yes but a task for another time

PR #1215, opened 2022-12-11
gtk: Subclass BuilderCScope for the BuilderRustScope

### LINKED ISSUE #1207: [BUG] crash on assertion failure in GtkListStore
The main bug is at https://gitlab.gnome.org/GNOME/gtk/-/issues/5398, but in short:

- Some type information does not propagate correctly, causing assertion failures
- When using a ListStore in an IconView, `gtk_icon_view_set_pixbuf_column` will print
  ```
  WARN  [Gtk] Unknown type GdkPixbuf specified in treemodel list_store
  WARN  [Gtk] ../gtk/gtk/gtkliststore.c:516: Invalid type (null)
  ERROR [Gtk] gtk_icon_view_set_pixbuf_column: assertion 'gtk_tree_model_get_column_type (icon_view->priv->model, column) == GDK_TYPE_PIXBUF' failed
  ```
- Alternatively, when trying to append rows to the ListStore Gtk will print `gtk_icon_view_set_pixbuf_column: assertion 'gtk_tree_model_get_column_type (icon_view->priv->model, column) == GDK_TYPE_PIXBUF' failed` and then it crashes in the binding logic: `thread 'main' panicked at 'column 0 is of type <invalid> but found value of type GdkPixbuf': /gtk4-0.5.2/src/list_store.rs:150`
- The weird part is that it looks like a Gtk regression: I did not run into the issue previously, and merely by changing the Gtk version without anything else it appears. However, I've been told that this part of Gtk has not been touched in ages so it might be a bindings issue

I currently suspect that some change in Gtk indirectly caused this issue, and that the bindings might be involved.

Reproducer: https://gitlab.gnome.org/GNOME/gtk/-/issues/5398#note_1609551

bilelmoussaoui:
Apprently, it is needed for keeping the get_type_from_name working properly Should fix #1207

### DIFF
--- a/gtk4/src/builder_cscope.rs
+++ b/gtk4/src/builder_cscope.rs
@@ -1,0 +1,12 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::BuilderScope;
+
+glib::wrapper! {
+    #[doc(alias = "GtkBuilderCScope")]
+    pub struct BuilderCScope(Object<ffi::GtkBuilderCScope, ffi::GtkBuilderCScopeClass>) @implements BuilderScope;
+
+    match fn {
+        type_ => || ffi::gtk_builder_cscope_get_type(),
+    }
+}
--- a/gtk4/src/builder_rust_scope.rs
+++ b/gtk4/src/builder_rust_scope.rs
@@ -1,7 +1,6 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-use crate::subclass::prelude::*;
-use crate::BuilderScope;
+use crate::{subclass::prelude::*, BuilderCScope, BuilderScope};
 use std::rc::Rc;
 
 glib::wrapper! {
@@ -36,6 +35,7 @@ glib::wrapper! {
     /// # }
     /// ```
     pub struct BuilderRustScope(ObjectSubclass<imp::BuilderRustScope>)
+        @extends BuilderCScope,
         @implements BuilderScope;
 }
 
@@ -66,12 +66,10 @@ impl BuilderRustScope {
 }
 
 mod imp {
-    use crate::prelude::*;
-    use crate::subclass::prelude::*;
-    use crate::{Builder, BuilderClosureFlags, BuilderError, BuilderScope};
-    use glib::translate::*;
-    use glib::{Closure, RustClosure};
-    use std::{cell::RefCell, collections::HashMap, rc::Rc};
+    use super::*;
+    use crate::{prelude::*, Builder, BuilderClosureFlags, BuilderError};
+    use glib::{translate::*, Closure, RustClosure};
+    use std::{cell::RefCell, collections::HashMap};
 
     type Callback = dyn Fn(&[glib::Value]) -> Option<glib::Value>;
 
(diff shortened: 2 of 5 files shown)


bilelmoussaoui:
Before this patch, the reproducer mentioned on the issue produces the following logs
```
     Running `target/debug/gdk_pixbuf_test`

(process:374631): Gtk-WARNING **: 09:37:50.132: Locale not supported by C library.
	Using the fallback 'C' locale.

(gdk_pixbuf_test:374631): Gtk-WARNING **: 09:37:50.157: Unknown type GdkPixbuf specified in treemodel list_store

(gdk_pixbuf_test:374631): Gtk-WARNING **: 09:37:50.157: ../gtk/deprecated/gtkliststore.c:545: Invalid type (null)

(gdk_pixbuf_test:374631): Gtk-CRITICAL **: 09:37:50.157: gtk_icon_view_set_pixbuf_column: assertion 'gtk_tree_model_get_column_type (icon_view->priv->model, column) == GDK_TYPE_PIXBUF' faile
```

and no warnings or errors once the patch is applied

bilelmoussaoui:
> This seems strange. Did you check with the GTK people if this is intentional?

Well, they are the ones that mentioned that. As BuilderScope is an interface, without defaulting to the vfuncs provided by BuilderCSCope the default get_type_from_name isn't that useful and we would have  to provide our own variant I guess. See https://gitlab.gnome.org/GNOME/gtk/-/commit/db3b216ad174f611f60176ab3447148475844bd6

sdroege:
Right, that's a bit annoying. At least the `get_type_from_name` of the C implementation is about the best we could also do.

For `gtk_builder_cscope_get_type_from_function()`, maybe override that one and always return `glib::Type::INVALID` there. The C implementation will never succeed. But I'm also not sure what this is actually supposed to do. Do you know? Is that something we should/can implement?

bilelmoussaoui:
> gtk_builder_cscope_get_type_from_function

We can't override that one, the CScope doesn't provide any vfuncs so not much that can be done I guess.

bilelmoussaoui:
And I have no idea where that can be useful actually, I would wait until someone hits a problem with it and then figure out a way forward

sdroege:
> We can't override that one, the CScope doesn't provide any vfuncs so not much that can be done I guess.

It's on the interface :)

bilelmoussaoui:
> > We can't override that one, the CScope doesn't provide any vfuncs so not much that can be done I guess.
> 
> It's on the interface :)

Oh I see what you meant. Fixed

matthiasclasen:
> gtk_builder_cscope_get_type_from_function

It is there because GtkBuilder has a feature where you can name a get_type() function in an attribute,

```
  <object class='MyFantasticObject' id='fantastic' type-func='my_fantastic_get_type'>
    ...
```

Its not very important. Nobody ever does that.

sdroege:
Ah, exciting. Thanks for the explanation :)

bilelmoussaoui:
> Very cool, many thanks. Do I correctly see that I can now drop all my `type_ensure` calls, including for my own custom widgets and it will automagically find them?

I don't think you can drop the ones for your custom widgets iirc

PR #1223, opened 2022-12-18
examples: add a menubar one

mikedilger:
I've written an example for creating and showing an application menubar


melix99:
That screenshot looks strangely old. Can you upload a new one using the "Default" theme?

mikedilger:
@melix99 I don't run gnome, so I'm not sure I have a "Default" theme on my ArchLinux XFCE box.

bilelmoussaoui:
> @melix99 I don't run gnome, so I'm not sure I have a "Default" theme on my ArchLinux XFCE box.

You definitely have as it comes with gtk itself.

examples/menubar/main.rs
```diff
@@ -0,0 +1,56 @@
...
+    about.connect_activate(|_, _| {
+        println!("About was pressed");
+    });
+    app.add_action(&about);
+
+    let quit = gio::SimpleAction::new("quit", None);
+    quit.connect_activate(glib::clone!(@weak app => move |_action, _parameter| {
+        app.quit();
+    }));
+    app.add_action(&quit);
```
bilelmoussaoui, line 26:
```suggestion
let about = gio::ActionEntry::builder("about")
    .activate(|_, _,_ | {
        println!("About was pressed")
    }).build();

let quit = gio::ActionEntry::builder("quit")
    .activate(|app, _, _| app.quit())
    .build();
app.add_action_entries([about, quit]).unwrap();
```

please re-format the code after updating to the suggested one

examples/menubar/main.rs
```diff
@@ -0,0 +1,56 @@
...
+    app.set_menubar(Some(&menubar));
+}
+
+fn build_ui(application: &gtk::Application) {
+    let window = gtk::ApplicationWindow::new(application);
+
+    window.set_title(Some("Menubar Example"));
+    window.set_default_size(350, 70);
+
+    window.set_show_menubar(true);
```
bilelmoussaoui, line 53:
```suggestion
    let window = gtk::ApplicationWindow::new(application);

    window.set_title(Some("Menubar Example"));
    window.set_default_size(350, 70);

    window.set_show_menubar(true);
```
Would prefer if you use the builder pattern here

bilelmoussaoui:
That screenshot looks like anything but the default theme. If you can't provide one that is similar to other ones, please drop it and I will add it in a future commit after merging your example.

melix99:
> That screenshot looks like anything but the default theme

The problem here is that the `GtkWindow` doesn’t have a titlebar, so the compositor is drawing a server side decoration and the author is probably not using mutter, so it doesn’t look “gnomish”. This can be easily solved with client side decoration by adding a `GtkHeaderBar` via `gtk_window_set_titlebar()`.

bilelmoussaoui:
Replaced by https://github.com/gtk-rs/gtk4-rs/pull/1570, thanks!

PR #1227, opened 2022-12-21
Fix compilation after `Stash` `PhantomData` usage in glib

sdroege:
@bilelmoussaoui There are probably more places where you can apply the same changes, and the other changes from the glib PR.


bilelmoussaoui:
Thanks, i will have a look next week during the holidays

PR #1232, opened 2023-01-01
document BitsetIter functions

euclio:


bilelmoussaoui:
As much as I would like to merge this, the real solution is to fix https://github.com/GuillaumeGomez/rustdoc-stripper to handle the cases where the impl trait contains a lifetime parameter.

euclio:
I'm not familiar with this tool. Can you explain a bit more about how it's used in gtk-rs?

bilelmoussaoui:
> I'm not familiar with this tool. Can you explain a bit more about how it's used in gtk-rs?

https://github.com/gtk-rs/gir is used to parse the xml files (gir files) and find the C docs for each item (function, struct & so on), convert the docs links to rust compatible ones, dump them in a markdown file and then the tool i mentioned is used to embed the docs from that markdown file in the rust files before running cargo doc in the CI.

euclio:
Isn't this manually implemented though? The existing docs wouldn't be very helpful since the Rust functions have a different return type.

bilelmoussaoui:
> Isn't this manually implemented though? The existing docs wouldn't be very helpful since the Rust functions have a different return type.

Even when manually implemented, rustdoc-stripper is capable of injecting the docs back.  The different return types are not really a huge issue, we could improve gir by making it not document the return value and do only that part manually but re-copying the whole docs for every manually implemented types/function is going to be a lot of work.

Feel free to open an issue on gir side to discuss how we could improve the situation here, but I am not going to merge the docs here.

PR #1234, opened 2023-01-02
file_chooser: Fix add_choice

A6GibKm:


bilelmoussaoui:
Does this patch fix your segafault crash?

A6GibKm:
Yes, seems like the issue was passing empty Some(vec![]) instead of None.

PR #1254, opened 2023-01-12
no need to move in the hello world exercise

harshshredding:
As far as I understand, `move` tries to move the current environment into the closure. However, since we are only using the closure's parameter(`|button|`) in its body, we don't need to `move`. Removing `move` could make the code clearer for first-time readers (like me).

The updated code seems to compile without issues.

Thank you for this book; I am so grateful.


bilelmoussaoui (approved):

Hofer-Julian:
Thanks!

PR #1256, opened 2023-01-14
Use wrapper macros where possible

### LINKED ISSUE #1244: Make use of new transparent type marker traits
See https://github.com/gtk-rs/gtk-rs-core/commit/b1513a6e043bddb69f2c20e840682a02d628ece0

bilelmoussaoui:
Indirectly fixes #1244


gtk4/src/pad_action_entry.rs
```diff
@@ -21,32 +19,34 @@ impl PadActionEntry {
...
-            mode,
-            label: label.to_owned(),
-            action_name: action_name.to_owned(),
+        unsafe {
+            Self::unsafe_from(ffi::GtkPadActionEntry {
+                type_: type_.into_glib(),
+                index,
+                mode,
+                label: label.as_ptr() as *const _,
+                action_name: action_name.as_ptr() as *const _,
```
sdroege, line 28: This borrows the strings without constraining things by lifetime. You should use `to_glib_full()` here and then you also need a `Drop` impl that frees the strings (or rather: add the corresponding functions into the macro above).

gtk4/src/recent_data.rs
```diff
@@ -1,85 +1,73 @@
...
+        app_name: &str,
+        app_exec: &str,
+        groups: &[&str],
+        is_private: bool,
+    ) -> Self {
+        unsafe {
+            Self::unsafe_from(ffi::GtkRecentData {
+                display_name: display_name.to_glib_full(),
+                description: description.to_glib_full(),
+                mime_type: mime_type.to_glib_full(),
```
sdroege, line 25: This also needs a `Drop` impl of some kind, i.e. ideally via the macro functions

gtk4/src/pad_action_entry.rs
```diff
@@ -21,32 +27,68 @@ impl PadActionEntry {
...
     pub fn action_name(&self) -> &str {
-        &self.action_name
+        unsafe { CStr::from_ptr(self.inner.action_name).to_str().unwrap() }
     }
 }
+
+unsafe fn copy_action_entry(
+    action_entry: *const ffi::GtkPadActionEntry,
+) -> *mut ffi::GtkPadActionEntry {
+    let copy = glib::ffi::g_malloc0(std::mem::size_of::<ffi::GtkPadActionEntry>())
```
sdroege, line 65: You can skip `copy` / `free` if it's just malloc/free

gtk4/src/pad_action_entry.rs
```diff
@@ -21,32 +27,68 @@ impl PadActionEntry {
...
+        as *mut ffi::GtkPadActionEntry;
+    copy_into_action_entry(copy, action_entry);
+    copy
+}
+
+unsafe fn free_action_entry(action_entry: *mut ffi::GtkPadActionEntry) {
+    clear_action_entry(action_entry)
```
sdroege, line 72: If you keep it however, you have to call `g_free(action_entry)`

gtk4/src/pad_action_entry.rs
```diff
@@ -21,32 +27,68 @@ impl PadActionEntry {
...
+}
+
+unsafe fn init_action_entry(action_entry: *mut ffi::GtkPadActionEntry) {
+    std::ptr::write(action_entry, std::mem::zeroed());
```
sdroege, line 76: That's the default already, zero initialization

gtk4/src/recent_data.rs
```diff
@@ -1,85 +1,120 @@
...
+    (*dest).is_private = (*src).is_private;
+}
+
+unsafe fn clear_recent_data(recent_data: *mut ffi::GtkRecentData) {
+    glib::ffi::g_free((*recent_data).display_name as *mut _);
+    glib::ffi::g_free((*recent_data).description as *mut _);
+    glib::ffi::g_free((*recent_data).mime_type as *mut _);
+    glib::ffi::g_free((*recent_data).app_name as *mut _);
+    glib::ffi::g_free((*recent_data).app_exec as *mut _);
+    glib::ffi::g_free((*recent_data).groups as *mut _);
```
sdroege, line 119:
```suggestion
    glib::ffi::g_strfreev((*recent_data).groups as *mut _);
```

gtk4/src/pad_action_entry.rs
sdroege: (some in the other one of course)

gtk4/src/pad_action_entry.rs
bilelmoussaoui: if i provide a copy_into/clear an init function is required

sdroege: That's strange. This https://github.com/gtk-rs/gtk-rs-core/blob/2e2af7ceaf2ec0e4eee474588f0fd0da86fad863/glib/src/boxed_inline.rs#L73 here should allow for exactly your use case

bilelmoussaoui: the problem is more here i believe https://github.com/gtk-rs/gtk-rs-core/blob/2e2af7ceaf2ec0e4eee474588f0fd0da86fad863/glib/src/wrapper.rs#L308-L312

sdroege: That should allow passing in only those though, and not the others.

sdroege: What exactly is the problem? Your current version that does not provide copy/free compiles fine for me.

bilelmoussaoui: Yes, it is more the init function that you cannot drop

PR #1262, opened 2023-01-15
Make use of IntoGStr/IntoStrV

### LINKED ISSUE #1240: Use `IntoGStr` / `IntoStrV` and collection types more widely in manual code
See also https://github.com/gtk-rs/gtk-rs-core/issues/884

bilelmoussaoui:
Fixes #1240

### DIFF
--- a/gtk4/src/constraint_layout.rs
+++ b/gtk4/src/constraint_layout.rs
@@ -2,18 +2,17 @@
 
 use crate::{prelude::*, Constraint, ConstraintLayout, Widget};
 use glib::{translate::*, IntoStrV};
-use std::collections::HashMap;
 
 impl ConstraintLayout {
     #[doc(alias = "gtk_constraint_layout_add_constraints_from_descriptionv")]
     #[doc(alias = "gtk_constraint_layout_add_constraints_from_description")]
     #[doc(alias = "add_constraints_from_descriptionv")]
-    pub fn add_constraints_from_description<W: IsA<Widget>>(
+    pub fn add_constraints_from_description<'a, W: IsA<Widget>>(
         &self,
         lines: impl IntoStrV,
         hspacing: i32,
         vspacing: i32,
-        views: &HashMap<&str, &W>,
+        views: impl IntoIterator<Item = (&'a str, &'a W)>,
     ) -> Result<Vec<Constraint>, glib::Error> {
         unsafe {
             let mut err = std::ptr::null_mut();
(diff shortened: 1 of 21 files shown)


gtk4/src/param_spec_expression.rs
```diff
@@ -84,15 +86,26 @@ impl FromGlibPtrFull<*mut gobject_ffi::GParamSpec> for ParamSpecExpression {
 impl ParamSpecExpression {
     #[allow(clippy::new_ret_no_self)]
     #[doc(alias = "gtk_param_spec_expression")]
-    pub fn new(name: &str, nick: &str, blurb: &str, flags: glib::ParamFlags) -> ParamSpec {
+    pub fn new(
+        name: impl IntoGStr,
+        nick: impl IntoGStr,
+        blurb: impl IntoGStr,
```
sdroege, line 92: These are optional or not? `IntoOptionalGStr`

gtk4/src/constraint_layout.rs
```diff
@@ -10,7 +10,7 @@ impl ConstraintLayout {
     #[doc(alias = "add_constraints_from_descriptionv")]
     pub fn add_constraints_from_description<W: IsA<Widget>>(
         &self,
-        lines: &[&str],
+        lines: impl IntoStrV,
         hspacing: i32,
         vspacing: i32,
         views: &HashMap<&str, &W>,
```
sdroege, line 16: Could also make this a `IntoIterator<Item = (&str, &W)>` FWIW

sdroege (commented):
Looks good to me otherwise. You could also look at `&str` and `Vec<String>` / `Vec<GString>` return values

gtk4/src/param_spec_expression.rs
bilelmoussaoui: Yes they are optional

bilelmoussaoui:
> Looks good to me otherwise. You could also look at `&str` and `Vec<String>` / `Vec<GString>` return values

Checked those, there were not that many. Only one case where we were returning `String` instead of `GString`

bilelmoussaoui:
> > Looks good to me otherwise. You could also look at `&str` and `Vec<String>` / `Vec<GString>` return values
> 
> Checked those, there were not that many. Only one case where we were returning `String` instead of `GString`

actually, i just noticed that i forgot doing the same thing for subclassing code. Will handle that first

Issue #1257, opened 2023-01-14
[HELP] Windows install: "cannot open input file 'cairo-gobject.lib'"


pieterdd:
When following the [Windows installation instructions](https://gtk-rs.org/gtk4-rs/stable/latest/book/installation_windows.html), I ran into a build problem during the librsvg steps at this command:

```
nmake /f Makefile.vc CFG=release install PREFIX=C:\gnome
```

Sanity checks I did based on a checklist by @wroyca:

- `git --version` outputs 'git version 2.37.2.windows.2'
- I found `C:\pkg-config-lite-0.28-1\bin` in the output of `echo $Env:Path`, same with `C:\gnome\bin`
- `echo $Env:PKG_CONFIG_PATH` outputs `C:\gnome\lib\pkgconfig`
- I'm using the PowerShell version of Visual Studio 2022 Developer Command Prompt v17.4.4. When I run the command in the legacy command prompt, I get the same error.

The error is:

```
  = note: LINK : fatal error LNK1181: cannot open input file 'cairo-gobject.lib'


The following warnings were emitted during compilation:

warning: File path C:/msys64/ucrt64/lib/libz.dll.a found in pkg-config file for cairo-pdf, but could not extract library base name to pass to linker command line
warning: File path C:/msys64/ucrt64/lib/libz.dll.a found in pkg-config file for cairo-ps, but could not extract library base name to pass to linker command line

error: could not compile `librsvg` due to previous error

Caused by:
  process didn't exit successfully: `rustc --crate-name rsvg_convert (...)
```

I could not find `cairo-gobject.lib` in either C:\gnome or C:\gtk. I'm not sure in which folder I should expect it. Do you think the presence of MSYS64 on my system is somehow interfering here? `C:/msys64/ucrt64/lib/` is not on my path but `C:\msys64\ucrt64\bin` is.

wroyca:
Yes, it seems that the presence of MSYS64 is interfering, best to remove it temporarily from path and see if you can continue without it. It is also worth checking that you have gettext 0.21 and iconv 1.16 (static) properly installed and available in your path as well

`GETTEXTCLDRDIR C:\Program Files\gettext-iconv\lib\gettext`

pieterdd:
[...]

PR #1263, opened 2023-01-15
book: warn about precedence over pkg-config-lite

wroyca:
Tell users to make sure that nothing takes precedence over pkgconfig-lite. This can be quite sneaky as sometimes a random install will pack its own pkgconfig.

Fix https://github.com/gtk-rs/gtk4-rs/issues/1257

### DIFF
--- a/book/src/installation_windows.md
+++ b/book/src/installation_windows.md
@@ -63,7 +63,7 @@ Download Gettext 0.21 from [mlocati.github.io](https://mlocati.github.io/article
 ### Pkg-config
 
 > **Warning**
-> **To avoid any potential problems, we strongly recommend that you make sure that nothing takes precedence over pkg-config-lite. You can use `where pkg-config` (legacy cmd) or `(Get-Command pkg-config).Path` (PowerShell) in that regards. https://github.com/gtk-rs/gtk4-rs/issues/1257**
+> **To avoid any potential problems, we strongly recommend that you make sure that nothing takes precedence over pkg-config-lite. You can use `where pkg-config` (legacy cmd) in that regards. https://github.com/gtk-rs/gtk4-rs/issues/1257**
 
 Download pkg-config-lite from [sourceforge.net](https://sourceforge.net/projects/pkgconfiglite/).
 Then extract and unpack it in `C:/`, so that the executable is in `C:\pkg-config-lite-0.28-1\bin`.
--- a/book/src/installation_windows.md
+++ b/book/src/installation_windows.md
@@ -62,6 +62,9 @@ Download Gettext 0.21 from [mlocati.github.io](https://mlocati.github.io/article
 
 ### Pkg-config
 
+> **Warning**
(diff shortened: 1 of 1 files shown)


book/src/installation_windows.md
```diff
@@ -62,6 +62,9 @@ Download Gettext 0.21 from [mlocati.github.io](https://mlocati.github.io/article
 
 ### Pkg-config
 
+> **Warning**
+> **To avoid any potential problems, we strongly recommend that you make sure that nothing takes precedence over pkg-config-lite. https://github.com/gtk-rs/gtk4-rs/issues/1257**
```
pieterdd, line 66: To be more specific, maybe the instructions could recommend users to check that `where pkg-config` (legacy cmd) or `(Get-Command pkg-config).Path` (PowerShell) resolves to the correct binary.

Optionally - although this may be a bit too verbose - even verify that `echo %PKG_CONFIG_PATH%` (legacy cmd) or `echo $Env:PKG_CONFIG_PATH` (PowerShell) displays the expected output.

Hofer-Julian: Powershell alone is fine

wroyca: @Hofer-Julian Any preference on how we could phrase this while keeping it short?

pieterdd: The guide currently doesn't mention PowerShell though, only cmd:

> From the Windows start menu, search for `x64 Native Tools Command Prompt for VS 2019`.

To use PowerShell I launched the Windows Terminal and picked the 'Developer PowerShell VS 2022' profile, which judging from its invocation I assume uses the x64 architecture:

```
powershell.exe -NoExit -Command "&{Import-Module """C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\Common7\Tools\Microsoft.VisualStudio.DevShell.dll"""; Enter-VsDevShell a6486c7f -SkipAutomaticLocation -DevCmdArguments """-arch=x64 -host_arch=x64"""}"
```

wroyca: I agree, I personally use cmd so we should cover both cases and not make assumptions

wroyca: Added @pieterdd recommendation to be more specific. Thank!

Hofer-Julian: Ah yeah my bad, I forgot that we have to rely on cmd for this. Can you please remove powershell then?

wroyca: > Ah yeah my bad, I forgot that we have to rely on cmd for this. Can you please remove powershell then?

Done! I think powershell works just as well, but that's untested territory on my part, so maybe we can add it later when we're more confident.

pieterdd: Fair enough. It worked for me in PowerShell, but agree to stick with the safer route for now.

Hofer-Julian:
Thanks.

PR #1267, opened 2023-01-16
book: remove `try_property`

Jedsek:


Jedsek:
just a minor change, see [new release blog](https://gtk-rs.org/blog/2022/10/18/new-release.html#removing-result-wrapping-in-some-functions-returned-values)

Hofer-Julian:
Thanks!

Feel free to just remove the whole sentence as well as the one before it.

alatiera:
It's possible that you might have code that deals with objects that have different properties and needs to check if its present. While `try_property` is now gone you can still get an optional `some_object.property::<Option<T>>("property-name")`

Hofer-Julian:
@alatiera Good point!

Hofer-Julian:
Thanks!

PR #1268, opened 2023-01-16
Enable introspection for gtk

YuraIz:
blueprint-compiler requires introspection to work


YuraIz:
looks like introspection for gtk requires introspection for core libraries in https://github.com/gtk-rs/gtk-rs-core/blob/master/Dockerfile

PR #1272, opened 2023-01-17
book: Move listings to feature 4_8

### LINKED ISSUE #1229: Listing from the book is incompatible with the current stable release of gtk4-rs.
In this [listing](https://github.com/gtk-rs/gtk4-rs/blob/master/book/listings/list_widgets/4/main.rs) line 39, in the signal handler for the `SignalListItemFactory`,
```rs
// ANCHOR: factory_setup
    factory.connect_setup(move |_, list_item| {
        // Create label
        let label = Label::new(None);
        list_item.set_child(Some(&label));

        // Bind `list_item->item->number` to `label->label`
        list_item
            .property_expression("item")
            .chain_property::<IntegerObject>("number")
            .bind(&label, "label", Widget::NONE);
    });
    // ANCHOR_END: factory_setup
```
No longer works because of an API change for the callbacks for [SignalListItemFactory](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/struct.SignalListItemFactory.html#method.connect_setup), the item passed to the callback is a `glib::object::Object` that needs to be downcast before `ListItem` methods such as `set_child()` can be used.

```rs
 let list_item = list_item
            .downcast_ref::<ListItem>()
            .expect("Needs to be ListItem");
```

Hofer-Julian:
Fixes  #1229

### DIFF
--- a/book/listings/todo/2/window/mod.rs
+++ b/book/listings/todo/2/window/mod.rs
@@ -189,8 +189,7 @@ impl Window {
                 .downcast_ref::<ListItem>()
                 .expect("Needs to be ListItem")
                 .item()
-                .expect("The item has to exist.")
-                .downcast::<TaskObject>()
+                .and_downcast::<TaskObject>()
                 .expect("The item has to be an `TaskObject`.");
 
             // Get `TaskRow` from `ListItem`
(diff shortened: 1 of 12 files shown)


book/listings/todo/2/window/mod.rs
```diff
@@ -176,20 +176,27 @@ impl Window {
...
 
         // Tell factory how to bind `TaskRow` to a `TaskObject`
         factory.connect_bind(move |_, list_item| {
             // Get `TaskObject` from `ListItem`
             let task_object = list_item
+                .downcast_ref::<ListItem>()
+                .expect("Needs to be ListItem")
                 .item()
                 .expect("The item has to exist.")
                 .downcast::<TaskObject>()
```
bilelmoussaoui, line 193: could use and_downcast here btw :)

Hofer-Julian: Oh nice, thanks.

PR #1298, opened 2023-02-10
gtk4: Use correct length for the `StrV` when passing to C

sdroege:
C doesn't want the `NUL`-terminator to be counted.

----

CC @pbor Can you check all your commits for the same bug?

Maybe the API should be changed to not return the `NULL` pointer at the end, but only return a slice of the valid elements (while still keeping the `NULL` pointer behind the last element). Opinions? Also @jf2048?


jf2048:
> Maybe the API should be changed to not return the `NULL` pointer at the end, but only return a slice of the valid elements (while still keeping the `NULL` pointer behind the last element). Opinions?

That would make sense and be consistent with `GStr`/`GString`, is there any problem with doing that?

sdroege:
No, only that it would change the API and has to wait until the next release

PR #1305, opened 2023-02-15
examples: add gtk4 femtovg example

jf2048:
This has been requested a few times in the matrix channel, should be stable enough to upstream now


bilelmoussaoui:
The code looks good but the screenshot doesn't :) would you mind fixing that?

jf2048:
Whoops, that was https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/6050

PR #1333, opened 2023-03-17
Remove mention of "clear_button" in the book

zekefast:
I assume "clear_button" was used in some earlier iterations of "todo_2" application. At the moment menu item is used for this purpose.
So, it make sense to remove its mentions from the text.


Hofer-Julian (approved):
Good catch, thank you!

zekefast:
Wow! It was merged quite fast! Thank you for the awesome book!

PR #1336, opened 2023-03-17
Change xml code block to diff in "Adapt Todo App".

zekefast:
# Overview

In order to signify addition of styles in [Adapt Todo App](https://gtk-rs.org/gtk4-rs/stable/latest/book/css.html#adapt-todo-app) of 14th chapter it probably worth to change code block type from `xml` to `diff` like it was done in other places e.g. [chapter 13](https://github.com/gtk-rs/gtk4-rs/blame/master/book/src/todo_2.md#L26).


Hofer-Julian (approved):
Thanks!

PR #1342, opened 2023-03-24
Update per glib::ControlFlow addition

bilelmoussaoui:
Depends on https://github.com/gtk-rs/gtk-rs-core/pull/1066


bilelmoussaoui:
Already handled.

PR #1348, opened 2023-03-27
gtk4-macros: Support blueprint files in CompositeTemplate

yuraiz:
That PR allows to refer to blueprint files using path related to `CARGO_MANIFEST_DIR`
I'm not sure how effective that implementation is, but I don't think I can do much better without 
[`proc_macro::Span::source_file`](https://doc.rust-lang.org/proc_macro/struct.Span.html#method.source_file)
Maybe I can make a simple file search to allow setting files without full path

image

### DIFF
--- a/gtk4-macros/src/composite_template_derive.rs
+++ b/gtk4-macros/src/composite_template_derive.rs
@@ -17,12 +17,28 @@ use crate::{attribute_parser::*, util::*};
 
 fn gen_set_template(source: &TemplateSource, crate_ident: &proc_macro2::Ident) -> TokenStream {
     match source {
-        TemplateSource::File(file) => quote! {
-            #crate_ident::subclass::widget::WidgetClassSubclassExt::set_template_static(
-                klass,
-                include_bytes!(#file),
-            );
-        },
+        TemplateSource::File(file) => {
+            let template = if file.ends_with(".blp") {
+                if cfg!(feature = "blueprint") {
+                    quote! {
+                        #crate_ident::gtk4_macros::include_blueprint!(#file).as_bytes()
+                    }
+                } else {
+                    panic!("blueprint feature is disabled")
+                }
+            } else {
+                quote! {
+                    include_bytes!(#file)
+                }
+            };
+
+            quote! {
+                #crate_ident::subclass::widget::WidgetClassSubclassExt::set_template_static(
+                        klass,
+                        #template,
+                );
+            }
+        }
         TemplateSource::Resource(resource) => quote! {
             #crate_ident::subclass::widget::WidgetClassSubclassExt::set_template_from_resource(
                 klass,
--- a/gtk4-macros/src/lib.rs
+++ b/gtk4-macros/src/lib.rs
@@ -15,6 +15,49 @@ use proc_macro::TokenStream;
 use proc_macro_error::proc_macro_error;
 use syn::{parse_macro_input, DeriveInput};
 
+/// That macro includes and compiles blueprint file by path relative to project rood
+///
+/// It expected to run inside composite_template_derive, not by users
+#[cfg(feature = "blueprint")]
+#[proc_macro]
+pub fn include_blueprint(input: TokenStream) -> TokenStream {
+    use quote::quote;
+
+    let tokens: Vec<_> = input.into_iter().collect();
+
+    if tokens.len() != 1 {
+        panic!("File name not found");
+    }
+
+    let root = std::env::var("CARGO_MANIFEST_DIR").unwrap_or_else(|_| ".".into());
+
+    let file_name = tokens[0].to_string();
+    let file_name = file_name.trim();
+    let file_name = &file_name[1..file_name.len() - 1];
+
+    let path = std::path::Path::new(&root).join(file_name);
+
+    if !path.exists() {
+        panic!("{path:?} not found");
+    }
+
+    let path = path.to_string_lossy().to_string();
+
+    let template = blueprint::compile_blueprint(
+        std::fs::read_to_string(&path)
+            .unwrap_or_else(|err| panic!("{err}"))
+            .as_bytes(),
+    )
+    .unwrap();
+
+    quote!({
+        // Compiler reruns macro if file changed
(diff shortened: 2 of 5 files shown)


gtk4-macros/src/lib.rs
```diff
@@ -5,8 +5,6 @@
 //! The crate aims to provide useful macros to use with the GTK 4 Rust bindings.
 
 mod attribute_parser;
-#[cfg(feature = "blueprint")]
-mod blueprint;
```
bilelmoussaoui, line 9: Why this change?

yuraiz: I can't export functions from gt4-macros. I'll revert this back 'cause I actually need to use a proc macro

gtk4-macros/src/composite_template_derive.rs
```diff
@@ -17,12 +17,24 @@ use crate::{attribute_parser::*, util::*};
...
+                    );
+                };
+            }
+
+            quote! {
+                #crate_ident::subclass::widget::WidgetClassSubclassExt::set_template_static(
+                    klass,
+                    include_bytes!(#file),
+                );
+            }
```
bilelmoussaoui, line 36:
```suggestion
            if file.ends_with(".blp") {
                quote! {
                    #crate_ident::subclass::widget::WidgetClassSubclassExt::set_template_static(
                        klass,
                        #crate_ident::gtk4_macros::include_blueprint!(#file).as_bytes(),
                    );
                }
            } else {
                quote! {
                    #crate_ident::subclass::widget::WidgetClassSubclassExt::set_template_static(
                        klass,
                        include_bytes!(#file),
                    );
                }
            }
```

gtk4-macros/src/lib.rs
```diff
@@ -15,6 +15,49 @@ use proc_macro::TokenStream;
...
+/// It expected to run inside composite_template_derive, not by users
+#[cfg(feature = "blueprint")]
+#[proc_macro]
+pub fn include_blueprint(input: TokenStream) -> TokenStream {
+    use quote::quote;
+
+    let tokens: Vec<_> = input.into_iter().collect();
+
+    if tokens.len() != 1 {
+        panic!("File name not found");
```
bilelmoussaoui, line 29: Not sure if the panics here are macros friendly. cc @GuillaumeGomez @jf2048

bilelmoussaoui:
Please add an entry to https://github.com/gtk-rs/gtk4-rs/blob/master/CHANGELOG.md#unreleased with potentially an example as well as mentioning at https://github.com/gtk-rs/gtk4-rs/blob/master/gtk4-macros/src/lib.rs#L90 that blueprint feature can be used with both string & file but not with resource.

gtk4-macros/src/lib.rs
yuraiz: [`pest_derive`](https://docs.rs/pest_derive/latest/src/pest_derive/lib.rs.html#322-325) calls [`derive_parser`](https://docs.rs/pest_generator/2.5.7/src/pest_generator/lib.rs.html#43-109) from `pest_generator` that panics

GuillaumeGomez: No they're not. You're supposed to return an error so people have an idea where it crashed and why. If the crate used panics, then another crate is needed.

GuillaumeGomez: Then don't use `anyhow` either. A proc macro is supposed to return a `TokenStream`, so if you encounter an error, you generate a `TokenStream` containing the right span and `compile_error!` with the correct message.

You can find an example [here](https://github.com/gtk-rs/gtk-rs-core/blob/master/glib-macros/src/clone.rs#L67).

yuraiz: That macro is called inside [`gen_set_template`](https://github.com/gtk-rs/gtk4-rs/pull/1348/files#:~:text=%23crate_ident%3A%3Agtk4_macros%3A%3Ainclude_blueprint!(%23file).as_bytes()) and I can't get correct span inside it, so there's no difference between `abort_call_site` and `compile_error!`

image
image

gtk4-macros/src/lib.rs
```diff
@@ -15,6 +15,51 @@ use proc_macro::TokenStream;
 use proc_macro_error::proc_macro_error;
 use syn::{parse_macro_input, DeriveInput};
 
+/// That macro includes and compiles blueprint file by path relative to project rood
+///
+/// It expected to run inside composite_template_derive, not by users
+#[cfg(feature = "blueprint")]
+#[proc_macro]
+#[proc_macro_error]
+pub fn include_blueprint(input: TokenStream) -> TokenStream {
```
bilelmoussaoui, line 24:
```suggestion
#[doc(hidden)]
pub fn include_blueprint(input: TokenStream) -> TokenStream {
```
To avoid people finding about it and misusing it

bilelmoussaoui:
Would you mind rebasing / fixing the remaining issues? I would like to get this in for the next release

yuraiz:
> Would you mind rebasing / fixing the remaining issues? I would like to get this in for the next release

Ok, I'll do that tomorrow

yuraiz:
What about using search like that https://github.com/yuraiz/gtk4-rs/commit/959adb41c7a3d7f5bace56f8b23b08c6eef24c70?

bilelmoussaoui:
> What about using search like that [yuraiz/gtk4-rs@blueprint...yuraiz:gtk4-rs:file-search-test](https://github.com/yuraiz/gtk4-rs/compare/blueprint...yuraiz:gtk4-rs:file-search-test)?

Looks too magical for the end-user to figure out the behavior. I would leave that feature out, for now at least.

bilelmoussaoui:
> What about using search like that [yuraiz/gtk4-rs@blueprint...yuraiz:gtk4-rs:file-search-test](https://github.com/yuraiz/gtk4-rs/compare/blueprint...yuraiz:gtk4-rs:file-search-test)?

Looks too magical for the end-user to figure out the behavior. I would leave that feature out, for now at least.

PR #1361, opened 2023-04-08
gtk: Implement convenience traits for StringObject

### LINKED ISSUE #1360: Implement convenience traits for StringObject
Things like `impl From<StringObject> for String`  / `impl From<StringObject> for GString`  & so on.

bilelmoussaoui:
Sadly we can't have AsRef/Deref implemented as well :(

Fixes #1360

### DIFF
--- a/gtk4/src/string_object.rs
+++ b/gtk4/src/string_object.rs
@@ -1,0 +1,97 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::StringObject;
+use glib::{GStr, GString};
+
+impl From<StringObject> for String {
+    #[inline]
+    fn from(value: StringObject) -> Self {
+        skip_assert_initialized!();
+        Self::from(&value)
+    }
+}
+
+impl From<StringObject> for GString {
+    #[inline]
+    fn from(value: StringObject) -> Self {
+        skip_assert_initialized!();
+        Self::from(&value)
+    }
+}
+
+impl<'a> From<&'a StringObject> for String {
+    #[inline]
+    fn from(value: &'a StringObject) -> Self {
+        skip_assert_initialized!();
+        value.string().to_string()
+    }
+}
+
+impl<'a> From<&'a StringObject> for GString {
+    #[inline]
+    fn from(value: &'a StringObject) -> Self {
+        skip_assert_initialized!();
+        value.string()
+    }
+}
+
+impl From<String> for StringObject {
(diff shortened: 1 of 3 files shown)


gtk4/src/string_object.rs
```diff
@@ -0,0 +1,59 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::StringObject;
+use glib::{GStr, GString};
+
+impl From<StringObject> for String {
```
sdroege, line 6:
```suggestion
impl<'a> From<&'a StringObject> for String {
```

etc? Or maybe in addition? There's no reason to consume the object here

gtk4/src/string_object.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+impl From<StringObject> for GString {
+    #[inline]
+    fn from(value: StringObject) -> Self {
+        value.string()
+    }
+}
+
+impl From<String> for StringObject {
```
sdroege, line 20:
```suggestion
impl<'a> From<&String> for StringObject {
```

in addition (and also for `GString`). It's usually covered by the `&str` / `&GStr` impls but not in every situation.

gtk4/src/string_object.rs
bilelmoussaoui: I added that in addition, thanks

gtk4/src/string_object.rs
```diff
@@ -0,0 +1,97 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+use crate::StringObject;
+use glib::{GStr, GString};
+
+impl From<StringObject> for String {
+    #[inline]
+    fn from(value: StringObject) -> Self {
+        assert_initialized_main_thread!();
```
sdroege, line 9:
```suggestion
        skip_assert_initialized!();
```

You can only get a `StringObject` after initialization. Also below

gtk4/src/string_object.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn from(value: &'a StringObject) -> Self {
+        assert_initialized_main_thread!();
+        value.string()
+    }
+}
+
+impl From<String> for StringObject {
+    #[inline]
+    fn from(value: String) -> Self {
+        assert_initialized_main_thread!();
```
sdroege, line 41: Also here skip, because `Self::new()` already has the very same assertion

PR #1363, opened 2023-04-08
book: Move to property macro

Hofer-Julian:


book/listings/g_object_properties/3/custom_button/imp.rs
```diff
@@ -57,7 +57,7 @@ impl WidgetImpl for CustomButton {}
 impl ButtonImpl for CustomButton {
     fn clicked(&self) {
         let incremented_number = self.number.get() + 1;
```
bilelmoussaoui, line 61: could have been a `self.obj().number() + 1` instead

Hofer-Julian: I was contemplating if I should change that.
But I remembered, that @sdroege suggested doing it that way since he considered property getters mostly public API.

bilelmoussaoui: Sure, but at least in this specific case it does exactly the same thing as the public API and it avoids using `.get()` or `.borrow()` for non copy cases

PR #1370, opened 2023-04-13
Bind Settings to `active` instead of `state`

Schmiddiii:
When binding to `state` the `Switch` can be in a wrong position where it is colored to be active, but the slider is at the wrong spot. As per advice in the GTK-chat, one should instead bind to the active-property.

image

### DIFF
--- a/book/listings/settings/2/main.rs
+++ b/book/listings/settings/2/main.rs
@@ -31,7 +31,7 @@ fn build_ui(app: &Application) {
 
     // ANCHOR: settings_bind
     settings
-        .bind("is-switch-enabled", &switch, "state")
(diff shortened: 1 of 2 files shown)


Hofer-Julian (approved):
Had another look and the other instances are probably fine.
But the book text has to be updated, can you do that before I merge it? :)

Schmiddiii:
I ammended that change to the existing commit. I am not sure how much the rest of the page should be changed as this also heavily relies on the "state", but has nothing to do with bindings so not sure if that can stay that way. Also, do you want to explicitly mention that "active" instead of "state" should be used in the text?

Hofer-Julian:
The current state is fine.
Thanks!

PR #1371, opened 2023-04-13
gtk4: allow subclassing WindowGroup

jf2048:
This is to complete subclassing `PanelWorkbench` in libpanel: [panel-workbench.h#L39](https://gitlab.gnome.org/GNOME/libpanel/-/blob/e2f8a7281ea7034819c1e647a2c2957e9fd1483f/src/panel-workbench.h#L39)


bilelmoussaoui (approved):
I thought it was final!

PR #1385, opened 2023-04-28
examples: update for 4.10 deprecations

jf2048:
After this, entry completion demo is the only one left with deprecation warnings. That depends on https://gitlab.gnome.org/GNOME/gtk/-/issues/5689

### DIFF
--- a/examples/entry_completion/main.rs
+++ b/examples/entry_completion/main.rs
@@ -1,3 +1,8 @@
+// TODO - EntryCompletion is deprecated without replacement in GTK 4.10. This example should be
+// updated to remove the deprecated code when a replacement lands.
+// See: https://gitlab.gnome.org/GNOME/gtk/-/issues/5689
+#![allow(deprecated)]
+
 use gtk::glib;
 use gtk::prelude::*;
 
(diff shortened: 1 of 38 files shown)


examples/entry_completion/main.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(deprecated)]
```
bilelmoussaoui, line 1: do you mind adding a comment here that points to the upstream gtk issue?

examples/video_player/video_player_window/imp.rs
```diff
@@ -1,12 +1,12 @@
...
-use gtk::{glib, prelude::*, CompositeTemplate};
+use gtk::{gio, glib, CompositeTemplate};
 
 #[derive(Debug, CompositeTemplate)]
 #[template(file = "video_player_window.ui")]
 pub struct VideoPlayerWindow {
     #[template_child(id = "video")]
     pub video: TemplateChild<gtk::Video>,
-    pub dialog: gtk::FileChooserNative,
+    pub dialog: gtk::FileDialog,
```
bilelmoussaoui, line 9: We no longer have to keep a reference to the dialog here, so let us drop it :)

bilelmoussaoui (approved):

PR #1386, opened 2023-04-28
manually add typed and with_type methods

### LINKED ISSUE #1384: New and with_type methods as in gtk-rs-core
Example: `Widget::ancestor(widget, type)` (which currently returns a `Option<Widget>`) could become `Widget::ancestor::<gtk::Label>(widget)` and directly return a typed `Option<Label>`.

ranfdev:
Closes #1384 

I've also found `DropTarget::new` which could receive the same treatment, but that's a more complex case, because it can also accept the type `G_TYPE_INVALID` when the user wants to specify multiple GTypes, so I've just skipped that.

### DIFF
--- a/gdk4/src/content_provider.rs
+++ b/gdk4/src/content_provider.rs
@@ -1,17 +1,22 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
 use crate::{prelude::*, ContentProvider};
-use glib::translate::*;
+use glib::{translate::*, value::FromValue};
 
 // rustdoc-stripper-ignore-next
 /// Trait containing manually implemented methods of [`ContentProvider`](crate::ContentProvider).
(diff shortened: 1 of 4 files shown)


jf2048:
Maybe it can be split into `DropTarget::new`, `DropTarget::with_type` and `DropTarget::with_types`?

bilelmoussaoui:
There are other ones as you can tell from the link i shared in the issue

bilelmoussaoui:
Please add a note to the changelog as well

bilelmoussaoui:
> Maybe it can be split into DropTarget::new, DropTarget::with_type and DropTarget::with_types?

That is what I ended up doing. It is a much better API :-)

Replaced by https://github.com/gtk-rs/gtk4-rs/pull/1415

PR #1389, opened 2023-05-03

bilelmoussaoui:


bilelmoussaoui:
Meh, analysing vfuncs adds new types to the auto imports.

PR #1392, opened 2023-05-04
Fix required features not shown in docs

AaronErhardt:
Fix for https://github.com/gtk-rs/gtk-rs-core/issues/1090 on the gtk4-rs side

So it seems like you need both RUSTFLAGS and RUSTDOCFLAGS because `build.rs` doesn't care about the doc flags and cargo only passes RUSTDOCFLAGS to rustdoc...


bilelmoussaoui:
Thanks, let us try and see how it works :)

PR #1394, opened 2023-05-07
gtk: Add missing IMContextImpl methods

### LINKED ISSUE #1388: Add missing IMContextImpl functions
Noticed while working on generating docs for virtual methods
 - [ ] set_surrounding_with_selection
 - [ ] surrounding_with_selection

bilelmoussaoui:
Fixes #1388

### DIFF
--- a/gtk4/src/subclass/im_context.rs
+++ b/gtk4/src/subclass/im_context.rs
@@ -59,6 +59,16 @@ pub trait IMContextImpl: ObjectImpl + ObjectSubclass<Type: IsA<IMContext>> {
     fn set_surrounding(&self, text: &str, cursor_index: i32) {
         self.parent_set_surrounding(text, cursor_index)
     }
+    #[cfg(feature = "v4_2")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "v4_2")))]
+    fn set_surrounding_with_selection(&self, text: &str, cursor_index: i32, anchor_index: i32) {
+        self.parent_set_surrounding_with_selection(text, cursor_index, anchor_index)
+    }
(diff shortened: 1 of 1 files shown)


gtk4/src/subclass/im_context.rs
```diff
@@ -506,6 +576,44 @@ unsafe extern "C" fn im_context_set_use_preedit<T: IMContextImpl>(
...
+    text: *const libc::c_char,
+    len: i32,
+    cursor_index: i32,
+    anchor_index: i32,
+) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.imp();
+
+    imp.set_surrounding_with_selection(
+        &glib::GString::from_glib_none_num(text, len as usize),
```
sdroege, line 591: Why create a copy here, and not use e.g. `glib::GStr`?

sdroege: Also in other places here

bilelmoussaoui: IIRC, we don't use glib::GStr in subclassing code at all. That is why I kept it using &str. But anyways, will update once #1678 lands

sdroege: It might be useful to use if otherwise you just have to convert to a `NUL`-terminated string again in the end. I don't know if that's the case here

bilelmoussaoui: The default impl does https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkimcontext.c#L431 so the len is important, i made use of glib::GStr where it made sense.

sdroege (approved):

PR #1396, opened 2023-05-12
book: Use property docs of of gtk-rs

Hofer-Julian:


book/src/g_object_properties.md
```diff
@@ -3,7 +3,7 @@
 Properties provide a public API for accessing state of GObjects.
 
 Let's see how this is done by experimenting with the [`Switch`](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/struct.Switch.html) widget.
-One of its properties is the [state](https://docs.gtk.org/gtk4/property.Switch.state.html).
+One of its properties is the [state](../docs/gtk4/struct.Switch.html#state).
```
bilelmoussaoui, line 6: oh, this should definitely make use of active instead of state

Hofer-Julian: Ah yes, will change that

PR #1407, opened 2023-06-28
Update to latest gir/gir-files

bilelmoussaoui:
TODO:

- [x] Add missing since configs for new enum members
- ~~Figure out the GdkGLTextureBuilder API~~ See https://github.com/gtk-rs/gtk4-rs/issues/1412
- [x] Seal manual traits
- [x] Figure out why sealing DialogExtManual / BuilderScopeImplExt causes a build error


gtk4/src/auto/enums.rs
```diff
@@ -1679,8 +1684,10 @@ pub enum Align {
     End,
     #[doc(alias = "GTK_ALIGN_CENTER")]
     Center,
-    #[doc(alias = "GTK_ALIGN_BASELINE")]
-    Baseline,
```
bilelmoussaoui, line 1683: This is a bit problematic, as it maps to the same value as `BaselineFill`. I guess we will have to manually write this trait so we can conditionally either have `Baseline` or `BaselineFill`

bilelmoussaoui: Fixed

gdk4/src/auto/enums.rs
```diff
@@ -1757,6 +1789,14 @@ impl FromGlib<ffi::GdkMemoryFormat> for MemoryFormat {
             }
             #[cfg(feature = "v4_6")]
             ffi::GDK_MEMORY_R32G32B32A32_FLOAT => Self::R32g32b32a32Float,
+            ffi::GDK_MEMORY_G8A8_PREMULTIPLIED => Self::G8a8Premultiplied,
+            ffi::GDK_MEMORY_G8A8 => Self::G8a8,
+            ffi::GDK_MEMORY_G8 => Self::G8,
+            ffi::GDK_MEMORY_G16A16_PREMULTIPLIED => Self::G16a16Premultiplied,
```
sdroege, line 1831: These need a `v4_12`

gsk4/sys/src/lib.rs
```diff
@@ -1508,6 +1529,12 @@ extern "C" {
     pub fn gsk_transform_node_get_child(node: *const GskTransformNode) -> *mut GskRenderNode;
     pub fn gsk_transform_node_get_transform(node: *const GskTransformNode) -> *mut GskTransform;
 
+    //=========================================================================
+    // GskVulkanRenderer
+    //=========================================================================
+    pub fn gsk_vulkan_renderer_get_type() -> GType;
```
sdroege, line 1535: No versions? That's problematic.

Also will the Vulkan renderer exist on all platforms? Will it exist in every GTK build or only if built with Vulkan support?

gtk4/src/auto/enums.rs
```diff
@@ -942,6 +942,8 @@ pub enum AccessibleRole {
     #[cfg_attr(docsrs, doc(cfg(feature = "v4_10")))]
     #[doc(alias = "GTK_ACCESSIBLE_ROLE_TOGGLE_BUTTON")]
     ToggleButton,
+    #[doc(alias = "GTK_ACCESSIBLE_ROLE_APPLICATION")]
+    Application,
```
sdroege, line 948: Version

gsk4/sys/src/lib.rs
bilelmoussaoui: Only if built with vulkan support, it needs a feature gate

gdk4-wayland/Cargo.toml
```diff
@@ -15,6 +15,7 @@ rust-version = "1.64"
 [features]
 v4_4 = ["ffi/v4_4"]
 v4_10 = ["ffi/v4_10", "v4_4"]
+v4_12 = ["ffi/v4_12", "v4_10"]
```
sdroege, line 18: Here and elsewhere, does it maybe need `gtk/v4_12` or so too? I.e. is enabling this feature requiring newly added APIs in dependencies?

sdroege: Also why do you only add it for `gdk4-wayland/Cargo.toml` and not all the other crates?

gdk4/src/gl_texture.rs
```diff
@@ -38,4 +41,14 @@ impl GLTexture {
...
         ))
     }
+
+    #[cfg(feature = "v4_12")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "v4_12")))]
+    // rustdoc-stripper-ignore-next
+    /// Creates a new builder-pattern struct instance to construct [`GLTexture`] objects.
+    ///
+    /// This method returns an instance of [`GLTextureBuilder`](crate::builders::GLTextureBuilder) which can be used to create [`GLTexture`] objects.
+    pub fn builder() -> GLTextureBuilder {
```
sdroege, line 51: You might also want to fix the API then. It's not really a Rust-style builder right now because you can't chain the methods (they're setters)

gtk4/src/auto/enums.rs
sdroege: In a later commit. Makes sense!

gsk4/sys/src/lib.rs
sdroege: Yeah, both in sys and non-sys :)

sdroege (commented):
Seems good to me otherwise

gdk4-wayland/Cargo.toml
bilelmoussaoui: > Here and elsewhere, does it maybe need gtk/v4_12 or so too? I.e. is enabling this feature requiring newly added APIs in dependencies?

That is true. I have enabled such features for gdk, there is no gtk dependency involved here.

> Also why do you only add it for gdk4-wayland/Cargo.toml and not all the other crates?

Other crates have that already.

gtk4/src/auto/column_view_cell.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[doc(alias = "GtkColumnViewCell")]
+    pub struct ColumnViewCell(Object<ffi::GtkColumnViewCell, ffi::GtkColumnViewCellClass>) @extends ListItem;
+
+    match fn {
+        type_ => || ffi::gtk_column_view_cell_get_type(),
+    }
+}
+
+impl ColumnViewCell {
+    // rustdoc-stripper-ignore-next
```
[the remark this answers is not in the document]
bilelmoussaoui: For the builder method that is "manually" documented

gdk4/src/gl_texture.rs
bilelmoussaoui: I have an open issue to handle the GLTextureBuilder API. Will leave that to later as it is not really blocking anything

sdroege: Makes sense

gtk4/src/auto/column_view_cell.rs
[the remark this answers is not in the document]
sdroege: How do you get gir to inject it here though?

gsk4/sys/src/lib.rs
bilelmoussaoui: good you brought this up as the broadway one wasn't properly configured

sdroege (approved):

PR #1410, opened 2023-07-04
Improve grammar

gianzellweger:
"Unfortunately, nobody found yet a good way to do that" => "Unfortunately, nobody has yet found a good way to do that"


Hofer-Julian (approved):
Thanks, that's better indeed!

PR #1421, opened 2023-07-09
Regen with latest gir

bilelmoussaoui:


gdk4/src/clipboard.rs
```diff
@@ -65,25 +65,22 @@ impl Clipboard {
...
-        Box<
-            dyn future::Future<Output = Result<(gio::InputStream, GString), glib::Error>> + 'static,
-        >,
-    > {
+    ) -> Result<(gio::InputStream, GString), glib::Error> {
         let mime_types = mime_types
             .iter()
             .copied()
             .map(String::from)
             .collect::<Vec<_>>();
```
sdroege, line 77: This is an API change now btw. The previous version was only borrowing `mime_types` for the duration of the call, this one is borrowing it until the resulting future is gone.

Not necessarily a problem but something to be aware of.

bilelmoussaoui:
Closing as I don't plan to push this forward any time soon

PR #1422, opened 2023-07-12
Upgrade syn to 2.0

### LINKED ISSUE #1409: Bump syn usage in gtk-macros
glib-macros has switched to using syn 2, which makes using gtk4-rs pulls both syn & syn 2

We should also get rid of proc-macro-error as well, as it is unmaintained and pulls syn1 back

jplatte:
Since proc-macro-error seems to be unmaintained, I've started removing it. It's used a bunch, so this PR won't be quite as simple as I would like.

Opening this PR to show work has started, and so sbd. can start where I left off if I don't pull through.

Resolves #1409

### DIFF
--- a/gtk4-macros/Cargo.toml
+++ b/gtk4-macros/Cargo.toml
@@ -23,10 +23,9 @@ blueprint = []
 anyhow = "1.0"
 quick-xml = {version = "0.28", optional = true}
 proc-macro-crate = "1.0"
-proc-macro-error = "1.0"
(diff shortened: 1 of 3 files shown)


bilelmoussaoui:
Any chance you could push this forward the upcoming days?

jplatte:
I don't think so, sorry.

bilelmoussaoui:
This is now done in main

PR #1440, opened 2023-07-29
Add dependabot for github actions

Hofer-Julian:
That way dependabot will open PRs on a weekly basis to update Github Actions dependencies

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
Thanks!

bilelmoussaoui:
Would you mind doing the same for crates as well? could be useful, at least to get an idea of when a new update is available for a crate

Hofer-Julian:
> Would you mind doing the same for crates as well? could be useful, at least to get an idea of when a new update is available for a crate

So you want to have a PR simply as notification which you then close again?

bilelmoussaoui:
> > Would you mind doing the same for crates as well? could be useful, at least to get an idea of when a new update is available for a crate
> 
> So you want to have a PR simply as notification which you then close again?

yeah, or merge them if no changes are needed

PR #1441, opened 2023-07-29
gdk/sys: Re-ignore N_FORMATS

bilelmoussaoui:
It was reverted in 8e7e745913d42f698cba83ac7c1bf2f96fdff0a6 as it is useful for -sys bindings
but it is probelmatic for us now as new values breaks our sys tests until we add a way to ignore certains consts from the abi test


bilelmoussaoui:
This is so annoying as ignoring an enum member doesn't drop it from the sys test. I guess we will just push the new gir-files in as well.

PR #1452, opened 2023-08-04
Impl Write on text buffers

ranfdev:
Text is always written at the end. I could insert the text at cursor position, but I think inserting always at the end provides more consistent behavior.

### DIFF
--- a/gtk4/src/entry_buffer.rs
+++ b/gtk4/src/entry_buffer.rs
@@ -127,3 +127,11 @@ impl Default for EntryBuffer {
         glib::Object::new()
     }
 }
+
+impl std::fmt::Write for EntryBuffer {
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
> I wanted to implement it also over `IsA<Editable>` but I'm unable to, because of rust's restrictions on doing blanked implementations of foreign traits.
> 
> Nevermind, I'll settle for implementing this on text buffers.

You could at least implement it for the provided Editables like Entry/Text ? Not sure if that is a good idea anyway

ranfdev:
Well I could, but the user can also call `entry.buffer()` and then use `Write` on that, so... We aren't missing much anyway.

This also keeps things consistent with libadwaita. I'm not going to implement `Write` on the adw's `Editable`'s.

PR #1468, opened 2023-08-16
gdk: Make RGBA::new const and add with_* constructors

SeaDve:
With adding the `with_` constructors, the builder may not be necessary anymore

### DIFF
--- a/gdk4/src/rgba.rs
+++ b/gdk4/src/rgba.rs
@@ -67,15 +67,15 @@ impl RGBABuilder {
 
 impl RGBA {
     #[inline]
-    pub fn new(red: f32, green: f32, blue: f32, alpha: f32) -> Self {
+    pub const fn new(red: f32, green: f32, blue: f32, alpha: f32) -> Self {
         skip_assert_initialized!();
-        unsafe {
-            Self::unsafe_from(ffi::GdkRGBA {
+        Self {
+            inner: ffi::GdkRGBA {
                 red,
                 green,
                 blue,
                 alpha,
-            })
+            },
         }
     }
 
--- a/gdk4/src/rgba.rs
+++ b/gdk4/src/rgba.rs
@@ -80,6 +80,84 @@ impl RGBA {
     }
 
     // rustdoc-stripper-ignore-next
+    /// Creates an owned [`RGBA`] like `self` but with the given red value.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use gdk4::RGBA;
+    ///
+    /// let rgba = RGBA::new(1.0, 1.0, 1.0, 1.0);
+    /// assert_eq!(rgba.with_red(0.5), RGBA::new(0.5, 1.0, 1.0, 1.0));
+    /// ```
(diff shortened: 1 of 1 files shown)


gdk4/src/rgba.rs
```diff
@@ -67,15 +67,49 @@ impl RGBABuilder {
 
 impl RGBA {
     #[inline]
-    pub fn new(red: f32, green: f32, blue: f32, alpha: f32) -> Self {
+    pub const fn new(red: f32, green: f32, blue: f32, alpha: f32) -> Self {
         skip_assert_initialized!();
-        unsafe {
-            Self::unsafe_from(ffi::GdkRGBA {
```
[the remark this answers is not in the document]
sdroege: That needs a separate trait then, and I'm not sure this can actually be const for everything (needs a bit of research)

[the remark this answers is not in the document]
bilelmoussaoui: Actually we can't have const fn in traits so scratch that proposal.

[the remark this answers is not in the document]
bilelmoussaoui: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cae7ddc92a024a2012edde38110c1bd1

gdk4/src/rgba.rs
```diff
@@ -67,15 +67,49 @@ impl RGBABuilder {
...
-            })
+            },
+        }
+    }
+
+    #[inline]
+    pub const fn with_red(self, red: f32) -> Self {
+        Self {
+            inner: ffi::GdkRGBA { red, ..self.inner },
+        }
```
bilelmoussaoui, line 86: just call self.set_red(red); and return self?

SeaDve: That's possible, but it can't be const since `set_red` can't be const because it takes `&mut self`

bilelmoussaoui: I am fine with either

SeaDve: As in, the latter is not clear that it is actually copying the struct, and not mutating

gdk4/src/rgba.rs
```diff
@@ -79,6 +79,40 @@ impl RGBA {
         }
     }
 
+    #[inline]
+    pub const fn with_red(self, red: f32) -> Self {
```
bilelmoussaoui, line 83: The commit message mentions constructors but they are more of helpers functions. Also I would be demanding and ask for at least a doc comment about the use case with an example.

SeaDve: Do we btw actually need an example?

bilelmoussaoui: I would prefer having one yes

bilelmoussaoui: It would also test that the code keep doing what it says it does

SeaDve: Makes sense

bilelmoussaoui (approved):

PR #1482, opened 2023-08-31
gsk: Add builder for Stroke

SeaDve:


gsk4/src/stroke.rs
```diff
@@ -0,0 +1,55 @@
...
+
+use crate::{LineCap, LineJoin, Stroke};
+
+impl Stroke {
+    // rustdoc-stripper-ignore-next
+    /// Creates a new builder-pattern struct instance to construct a [`Stroke`].
+    ///
+    /// This method returns an instance of [`StrokeBuilder`](crate::builders::StrokeBuilder) which can be used to create a [`Stroke`].
+    pub fn builder(line_width: f32) -> StrokeBuilder {
+        skip_assert_initialized!();
```
SeaDve, line 11: I'm not sure if this is the correct macro to use

bilelmoussaoui: It should be `assert_initialized_main_thread`

SeaDve: I see, thanks. I'm confused on how it is determined

bilelmoussaoui: You end up calling gsk::Stroke::new which has `assert_initialized_main_thread`

SeaDve: Oh, that makes sense

PR #1484, opened 2023-09-01
book: Move from `pub` to `pub(crate)`

Hofer-Julian:
It is arguably the correct way for this

### DIFF
--- a/book/listings/actions/5/window/imp.rs
+++ b/book/listings/actions/5/window/imp.rs
@@ -5,9 +5,9 @@ use gtk::{glib, CompositeTemplate, Label};
 // Object holding the state
 #[derive(CompositeTemplate, Default)]
 #[template(resource = "/org/gtk_rs/example/window.ui")]
-pub struct Window {
+pub(crate) struct Window {
     #[template_child]
-    pub label: TemplateChild<Label>,
+    pub(crate) label: TemplateChild<Label>,
 }
 
 // The central trait for subclassing a GObject
--- a/book/listings/actions/5/window/mod.rs
+++ b/book/listings/actions/5/window/mod.rs
@@ -7,7 +7,7 @@ use gtk::subclass::prelude::*;
 use gtk::{gio, glib, Application};
 
 glib::wrapper! {
-    pub struct Window(ObjectSubclass<imp::Window>)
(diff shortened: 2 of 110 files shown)


book/listings/actions/5/window/imp.rs
```diff
@@ -5,9 +5,9 @@ use gtk::{glib, CompositeTemplate, Label};
 // Object holding the state
 #[derive(CompositeTemplate, Default)]
 #[template(resource = "/org/gtk_rs/example/window.ui")]
-pub struct Window {
```
bilelmoussaoui, line 8: the imp struct could be pub(super) though

book/listings/actions/5/window/imp.rs
```diff
@@ -5,9 +5,9 @@ use gtk::{glib, CompositeTemplate, Label};
...
+pub(crate) struct Window {
     #[template_child]
-    pub label: TemplateChild<Label>,
+    pub(crate) label: TemplateChild<Label>,
```
bilelmoussaoui, line 10: same, it should be pub(crate)

Hofer-Julian:
> Using pub(crate) for a binary doesn't really make sense.

I thought so, too.
Turns out that also for binaries, cargo is not able to detect dead code if it's fully public.

book/listings/actions/5/window/imp.rs
bilelmoussaoui: yes, edited

Hofer-Julian:
> > > Using pub(crate) for a binary doesn't really make sense.
> > 
> > 
> > I thought so, too. Turns out that also for binaries, cargo is not able to detect dead code if it's fully public.
> 
> you can't really detect dead code for gtk apps, especially if you use composite template callbacks

Yeah, just noticed that too. Not sure if it's a macro thing, but static analysis totally fails there.
This makes this PR mostly useless.
Thanks for the comments @bilelmoussaoui!

bilelmoussaoui:
> Yeah, just noticed that too. Not sure if it's a macro thing, but static analysis totally fails there.
This makes this PR mostly useless.

The macro actually uses the functions to create closures from them, but they can only be used at runtime. so nothing can be done at build time...

PR #1494, opened 2023-09-23
New release!

bilelmoussaoui:


bilelmoussaoui:
The CI failure is not "crucial" and fixing it would be an API break so let us just land this

PR #1495, opened 2023-09-23
gtk: Implement HasParamSpec for TemplateChild<T>

### LINKED ISSUE #1456: Support `TemplateChild` in properties macro
We should implement `glib::Property` on that

bilelmoussaoui:
Fixes #1456

Untested. I am not sure about the use cases though, so help with that would be appreciated

### DIFF
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -1230,6 +1230,19 @@ where
     }
 }
 
+impl<T> glib::HasParamSpec for TemplateChild<T>
+where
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
I guess a use case is to have it for a child widgets, i see

sdroege (approved):
Seems good to me. Can't really hurt to have that, can it?

ranfdev (approved):

ranfdev:
The use case is being able to expose a child widget as a read only property.

blueprint-compiler is able to bind nested values, as `mywidget.mywidget_child.name`,  but in that expression `mywidget_child` must be a property. With this PR exposing `mywidget_child` becomes trivial

PR #1508, opened 2023-10-09
book: Explain async ecosystems

Hofer-Julian:
@sdroege can you please verify if this is a correct summary of the Rust async situation?

### DIFF
--- a/book/src/main_event_loop.md
+++ b/book/src/main_event_loop.md
@@ -86,6 +86,12 @@ Filename: <a class=file-link href="https://github.com/gtk-rs/gtk4-rs/blob/master
 {{#rustdoc_include ../listings/main_event_loop/4/main.rs:callback}}
 ```
 
+> There are two async ecosystems in Rust: [`tokio`](https://docs.rs/tokio) and [`futures`](https://docs.rs/futures).
+> Crates based on [`glib`](https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/), [`smol`](https://docs.rs/smol/latest/smol/) or [`async-std`](https://docs.rs/async-std/1.12.0/async_std/) work with `futures`.
+> In order to use `tokio` based crates you can either:
+> - maintain a static variable of type `once_cell::sync::Lazy<tokio::runtime::Runtime>` and spawn tasks with it, or
+> - use the [`async-compat`](https://docs.rs/async-compat/latest/async_compat/) crate.
+
 Since we are single-threaded again, we could even get rid of the channels while achieving the same result.
 
 Filename: <a class=file-link href="https://github.com/gtk-rs/gtk4-rs/blob/master/book/listings/main_event_loop/5/main.rs">listings/main_event_loop/5/main.rs</a>


book/src/main_event_loop.md
```diff
@@ -86,6 +86,12 @@ Filename: <a class=file-link href="https://github.com/gtk-rs/gtk4-rs/blob/master
 {{#rustdoc_include ../listings/main_event_loop/4/main.rs:callback}}
 ```
 
+> There are two async ecosystems in Rust: [`tokio`](https://docs.rs/tokio) and [`futures`](https://docs.rs/futures).
```
sdroege, line 89: That sentence doesn't make much sense to me. tokio is also based on the futures crate, just that tokio has its own `Stream` trait.

Maybe say that there are multiple async runtimes in Rust: e.g. tokio which is most widely used, async-std and smol. And that another async runtime is provided by the glib crate.

Also that some crates expect a specific runtime to be used and you would have to implement interop between the different runtimes, e.g. via channels. And in the case of tokio, smol and glib you also need to make sure there is an actual runtime running because it doesn't happen implicitly.

sdroege: also async-std does not have any (stable) stream support at all, for example

book/src/main_event_loop.md
```diff
@@ -86,6 +86,12 @@ Filename: <a class=file-link href="https://github.com/gtk-rs/gtk4-rs/blob/master
...
+> Crates based on [`glib`](https://gtk-rs.org/gtk-rs-core/stable/latest/docs/glib/), [`smol`](https://docs.rs/smol/latest/smol/) or [`async-std`](https://docs.rs/async-std/1.12.0/async_std/) work with `futures`.
+> In order to use `tokio` based crates you can either:
+> - maintain a static variable of type `once_cell::sync::Lazy<tokio::runtime::Runtime>` and spawn tasks with it, or
+> - use the [`async-compat`](https://docs.rs/async-compat/latest/async_compat/) crate.
```
sdroege, line 93: async-compat only provides some translation between traits that are defined differently in futures and tokio. You still need to make sure a tokio runtime is running somewhere.

Usually you'd use it in combination with managing a tokio runtime.

sdroege:
I think this has to be written differently, and become a proper paragraph instead of a few bullet points :)

sdroege:
> To my knowledge, reqwest only works with tokio, but it doesn't seem to spell that out anywhere.

You can check the dependencies. That's what you'll have to do for using any async crates in addition to checking the docs. Most of the Rust ecosystem either depends on tokio or is runtime agnostic, very few crates require anything else.

Usually when you want to use some crate you'll notice it using tokio and then you don't have any choice anyway :)

> But also here, I don't see it spelled out that this is fine to poll something from the async-std ecosystem with glib. I am not 100% sure myself to be honest.

I think that's fine with async-std because it has a global runtime that is used for the actual work. When you await the future, you're just waiting for the result really, it would basically be like a oneshot channel.

I didn't notice that this is possible with async-std though, that's useful to know :)

Hofer-Julian:
> So yes, a chapter with multiple sections seems best. Explaining all of the above but also how to use tokio correctly with a bit of example code, and the whole thing about how to use channels.

Sounds good to me!

PR #1510, opened 2023-10-11
book: Rename action to correct name

Hofer-Julian:


book/listings/todo/8/window/mod.rs
```diff
@@ -387,11 +387,13 @@ impl Window {
 
         // ANCHOR: setup_actions
         // Create action to create new collection and add to action group "win"
-        let action_new_list = gio::SimpleAction::new("new-collection", None);
-        action_new_list.connect_activate(clone!(@weak self as window => move |_, _| {
-            window.new_collection();
-        }));
-        self.add_action(&action_new_list);
+        let action_new_collection = gio::SimpleAction::new("new-collection", None);
```
bilelmoussaoui, line 390: if you switch to ActionEntry::builder, you would no longer need the clone! usage here :)

Hofer-Julian: Could you give me an example how I would use that?
I don't get it from the docs alone

Hofer-Julian: Nevermind, found a couple of examples in other apps.

PR #1511, opened 2023-10-11
book: Extend main loop chapter with async section

Hofer-Julian:
See discussion in https://github.com/gtk-rs/gtk4-rs/pull/1508

- [x] Show how to await on `gio::spawn_blocking`
- [x] Await on ashpd
- [x] Use reqwest with tokio


sdroege:
Let me know when you want me to go over it for a review :)

Hofer-Julian:
> Let me know when you want me to go over it for a review :)

Thanks! Will do that :)

Hofer-Julian:
@sdroege @felinira this PR is now ready for review. Feel free to also comment on parts of the chapter that wasn't touched by this PR.

You can render it locally by running:

```
cargo install mdbook
git switch Hofer-Julian/book-async
mdbook serve book
```

CI fails because I reference listings that will be there after this PR is merged.

Hofer-Julian (commented):
Thanks for the review @felinira!
Your suggestions make sense to me and I applied most of the immediately :)

sdroege:
A bit late, but I went through everything now and don't have a single suggestion what to change. Good work, this is a lot more understandable and short than what I would've come up with :)

PR #1521, opened 2023-10-22
book: Move to async-channel

Hofer-Julian:
async-channel covers more use cases.
This also fits with my observation that it is popular within gtk-rs apps

### DIFF
--- a/book/listings/Cargo.lock
+++ b/book/listings/Cargo.lock
@@ -9,6 +9,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3b13c32d80ecc7ab747b80c3784bce54ee8a7a0cc4fbda9bf4cda2cf6fe90854"
 
 [[package]]
+name = "async-channel"
+version = "1.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


sdroege:
It might be worth mentioning in the book *why* you use `async-channel` (I'm usually also using that, or just the channels from the `futures` crate...).

Should we maybe remove the GLib channel? It mostly exists from the times before async was mature and because people instead used timeout callbacks to periodically check if a non-async channel contains something.

Hofer-Julian:
> It might be worth mentioning in the book why you use async-channel

Over the `futures` one? Tbh, I didn't know that one existed.
Could you tell me what's the reason to choose one over the other?

> Should we maybe remove the GLib channel? It mostly exists from the times before async was mature and because people instead used timeout callbacks to periodically check if a non-async channel contains something.

I think that would be good, yes.
Should I open an issue for that?

sdroege:
> Could you tell me what's the reason to choose one over the other?

`async-channel` is a more modern implementation. I think it's more lightweight, has lower overhead and IIRC there's also some minimal API differences?

Probably doesn't matter too much of the majority of applications which one you use.

> Should I open an issue for that?

Yes please

PR #1522, opened 2023-10-22
book: Use bounded channels instead of unbounded

Hofer-Julian:
They should be enough in most situations.


Hofer-Julian:
@sdroege most of the async channel chapter is already rendered here: https://gtk-rs.org/gtk4-rs/stable/latest/book/main_event_loop.html

Feel free to tell me if you think something should be changed here (like e.g. the explanation why `async-channel`) :)

PR #1546, opened 2023-12-17
Add clarification around creating tokio runtime

JakeStanger:
Off the back of https://github.com/gtk-rs/gtk-rs-core/issues/1251, this adds some additional detail to the Tokio section of the main event loop page that explains *why* the runtime is manually created, and the reasons to avoid doing it the 'normal' way.


Hofer-Julian (approved):
Sounds good. Thanks!

PR #1562, opened 2024-01-16
Update to proc-macro-crate 3

sdroege:


sdroege:
Ah, duplicate of https://github.com/gtk-rs/gtk4-rs/pull/1555

PR #1563, opened 2024-01-19
print_job: fix send() closure

pbor:
Pass a result, since error is set only on failure. Also avoid passing the job itself to the rust closure since that is redundant.

### DIFF
--- a/gtk4/src/print_job.rs
+++ b/gtk4/src/print_job.rs
@@ -8,22 +8,24 @@ use crate::PrintJob;
 
 impl PrintJob {
     #[doc(alias = "gtk_print_job_send")]
-    pub fn send<P: Fn(&PrintJob, &glib::Error) + 'static>(&self, callback: P) {
+    pub fn send<P: Fn(&PrintJob, Result<(), glib::Error>) + 'static>(&self, callback: P) {
         let callback_data: Box_<P> = Box_::new(callback);
-        unsafe extern "C" fn callback_func<P: Fn(&PrintJob, &glib::Error) + 'static>(
+        unsafe extern "C" fn callback_func<P: Fn(&PrintJob, Result<(), glib::Error>) + 'static>(
             print_job: *mut ffi::GtkPrintJob,
             user_data: glib::ffi::gpointer,
             error: *const glib::ffi::GError,
         ) {
             let print_job = from_glib_borrow(print_job);
-            let error = from_glib_borrow(error);
+            let result = if error.is_null() {
+                Ok(())
+            } else {
+                Err(from_glib_none(error))
+            };
             let callback: &P = &*(user_data as *mut _);
-            (*callback)(&print_job, &error);
+            (*callback)(&print_job, result);
         }
         let callback = Some(callback_func::<P> as _);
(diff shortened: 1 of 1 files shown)


gtk4/src/print_job.rs
```diff
@@ -6,20 +6,23 @@ use std::boxed::Box as Box_;
...
+            } else {
+                Err(from_glib_none(error))
+            };
             let callback: &P = &*(user_data as *mut _);
-            (*callback)(&print_job, &error);
+            (*callback)(result);
         }
         let callback = Some(callback_func::<P> as _);
-        unsafe extern "C" fn dnotify_func<P: Fn(&PrintJob, &glib::Error) + 'static>(
+        unsafe extern "C" fn dnotify_func<P: Fn(Result<(), glib::Error>) + 'static>(
```
sdroege, line 25:
```suggestion
        unsafe extern "C" fn dnotify_func<P>(
```

should be enough

gtk4/src/print_job.rs
```diff
@@ -6,20 +6,23 @@ use std::boxed::Box as Box_;
...
-            let print_job = from_glib_borrow(print_job);
-            let error = from_glib_borrow(error);
+            let result = if error.is_null() {
+                Ok(())
+            } else {
+                Err(from_glib_none(error))
+            };
             let callback: &P = &*(user_data as *mut _);
-            (*callback)(&print_job, &error);
+            (*callback)(result);
```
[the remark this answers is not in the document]
pbor: I figured that you can always pass it in the closure with `glib::clone(@strong job, ...` if one needs it. No strong opinion though

gtk4/src/print_job.rs
pbor: I kept it as it was before and mechanically replaced... I will wait for CI to pass before applying this since I cannot test locally

gtk4/src/print_job.rs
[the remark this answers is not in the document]
sdroege: That probably causes a reference cycle though

gtk4/src/print_job.rs
bilelmoussaoui: The CI now should be green, so feel free to rebase

gtk4/src/print_job.rs
```diff
@@ -8,22 +8,24 @@ use crate::PrintJob;
...
             print_job: *mut ffi::GtkPrintJob,
             user_data: glib::ffi::gpointer,
             error: *const glib::ffi::GError,
         ) {
             let print_job = from_glib_borrow(print_job);
-            let error = from_glib_borrow(error);
+            let result = if error.is_null() {
+                Ok(())
+            } else {
+                Err(from_glib_none(error))
```
bilelmoussaoui, line 22: This should be 
```suggestion
                Err(from_glib_full(error))
```

pbor: https://docs.gtk.org/gtk4/callback.PrintJobCompleteFunc.html

says that the error is owned by the caller of the callback, so I do not think we should do `from_glib_full`.
Note that the old code was doing `from_glib_borrow`, so we could keep that and pass `Result<(), &glib::Error>`, but `Result<(), glib::Error>` seems more consistent with other APIs in gtk4-rs

Issue #853, opened 2022-01-25
Add an example that makes use of async operations


bilelmoussaoui:
Something like sending a web request to retrieve some data and then display it. Maybe using tokio or some other async runtime.

sdroege:
There's an example in gtk-rs-core and I think also an example for dialogs here?

> Maybe using tokio or some other async runtime.

That will involve a bit of boilerplate for the cross-runtime communication then but probably a good idea to have an example for that.

bilelmoussaoui:
> There's an example in gtk-rs-core and I think also an example for dialogs here?

There is indeed, but it doesn't show how to integrate with a window without blocking the main thread

Claudio-code:
Iwrote it example https://github.com/gtk-rs/gtk4-rs/pull/1578

but CI falls because not found openssl dependency https://github.com/gtk-rs/gtk4-rs/actions/runs/7649773189/job/20844748624

PR #1578, opened 2024-01-25
Examples: async request no blocking main thread

Claudio-code:
Tried fix it https://github.com/gtk-rs/gtk4-rs/issues/853

- Make async web request to pokeapi and list all in window
- When scroll to end make new request and add new pokemons in list


bilelmoussaoui:
Thanks a lot for your contribution, please send a PR to add the needed dependencies to https://github.com/gtk-rs/gtk4-rs/blob/master/Dockerfile. 

But before doing so, the other option is to disable the default features of reqwest and use `rustls-tls` instead. That will make the example easily runnable on other platforms like windows

examples/Cargo.toml
```diff
@@ -13,6 +13,10 @@ glow = { version = "0.13.0", optional = true }
 epoxy = { version = "0.1.0", optional = true }
 libloading = { version = "0.8.0", optional = true }
 im-rc = { version = "15", optional = true }
+async-channel = "2.1.1"
+tokio = { version = "1", features = ["full"] }
+reqwest = { version = "0.11", features = ["json"] }
+serde = { version = "1.0", features = ["derive"] }
```
bilelmoussaoui, line 19: All of those should be optional, see how it is done for other examples

examples/Cargo.toml
```diff
@@ -195,3 +199,7 @@ path = "video_player/main.rs"
 [[bin]]
 name = "virtual_methods"
 path = "virtual_methods/main.rs"
+
+[[bin]]
+name = "request_no_blocking_main_thread"
+path = "request_no_blocking_main_thread/main.rs"
```
bilelmoussaoui, line 205: Keep the file ordered please. Also maybe just name it `tokio` ? as it basically shows how to integrate tokio in your gtk-rs app.

examples/request_no_blocking_main_thread/main.rs
```diff
@@ -0,0 +1,80 @@
+mod pokemon_client;
+
+use glib::clone;
+use gtk::glib::once_cell::sync::Lazy;
+use gtk::{glib, Application, ApplicationWindow, Label, ListBox, PolicyType, ScrolledWindow};
+use gtk::{prelude::*, PositionType};
```
bilelmoussaoui, line 6: Please avoid including types that are used once.

examples/request_no_blocking_main_thread/main.rs
```diff
@@ -0,0 +1,80 @@
...
+fn build_ui(app: &Application) {
+    let (sender, receiver) = async_channel::bounded::<Result<Vec<Pokemon>, reqwest::Error>>(1);
+
+    RUNTIME.spawn(clone!(@strong sender => async move {
+        let mut pokemon_client = PokemonClient::new();
+        let pokemon_vec = pokemon_client.get_pokemon_list().await;
+        sender.send(pokemon_vec).await.expect("The channel needs to be open.");
+    }));
+
+    let result_poke_vec = receiver.recv_blocking().unwrap();
```
bilelmoussaoui, line 30: This doesn't look good at all, instead wait for the response on  `glib::spawn_future_local`, the app won't start until a response is received

bilelmoussaoui: that would avoid you duplicating the same code below, actually this code can just be removed :)

examples/request_no_blocking_main_thread/main.rs
```diff
@@ -0,0 +1,80 @@
...
+                sender.send(pokemon_vec).await.expect("The channel needs to be open.");
+            }));
+        }
+    });
+
+    glib::spawn_future_local(async move {
+        while let Ok(response) = receiver.recv().await {
+            if let Ok(response) = response {
+                for pokemon in response {
+                    let label = Label::new(Some(&pokemon.name));
```
bilelmoussaoui, line 63: set halign on the label to start, using the builder pattern. That would make the UI slightly better

examples/request_no_blocking_main_thread/main.rs
```diff
@@ -0,0 +1,80 @@
...
+                }
+            } else {
+                println!("bad request");
+            }
+        }
+    });
+
+    let window = ApplicationWindow::builder()
+        .application(app)
+        .title("pokemon list infinite scroll")
```
bilelmoussaoui, line 74: I don't like this title, because it is a lie. GtkListBox doesn't scale for infinite scrolling and you will get issues around 1k items, probably less, if your widget is complex.

Instead, I would just say "Tokio integration" or whatever.

examples/request_no_blocking_main_thread/pokemon_client.rs
```diff
@@ -0,0 +1,41 @@
...
+impl PokemonClient {
+    const URI: &'static str = "https://pokeapi.co/api/v2/pokemon";
+    const PAGE_LIMIT: &'static str = "100";
+
+    pub fn new() -> Self {
+        Self { page: 0 }
+    }
+
+    pub async fn get_pokemon_list(&mut self) -> Result<Vec<Pokemon>, reqwest::Error> {
+        let current_offset = self.page.clone().to_string();
```
bilelmoussaoui, line 27:
```suggestion
        let current_offset = self.page.to_string();
```
you don't need to clone a copy type

examples/request_no_blocking_main_thread/pokemon_client.rs
```diff
@@ -0,0 +1,41 @@
...
+        let params: [(&str, &str); 2] = [
+            ("limit", Self::PAGE_LIMIT),
+            ("offset", current_offset.as_str()),
+        ];
+        let url = reqwest::Url::parse_with_params(Self::URI, params);
+        let body = reqwest::get(url.unwrap())
```
bilelmoussaoui, line 33: Move the unwrap above

examples/request_no_blocking_main_thread/pokemon_client.rs
```diff
@@ -0,0 +1,41 @@
+use serde::Deserialize;
```
bilelmoussaoui, line 1: The extra file doesn't contain much, so I would just put everything in a single container file

examples/request_no_blocking_main_thread/main.rs
```diff
@@ -0,0 +1,80 @@
...
+
+use glib::clone;
+use gtk::glib::once_cell::sync::Lazy;
+use gtk::{glib, Application, ApplicationWindow, Label, ListBox, PolicyType, ScrolledWindow};
+use gtk::{prelude::*, PositionType};
+use tokio::runtime::Runtime;
+
+use pokemon_client::{Pokemon, PokemonClient};
+
+const APP_ID: &str = "org.gtk_rs.pokemon.list";
```
bilelmoussaoui, line 11: it is used only once, drop the const please

bilelmoussaoui (commented):
overall looks good, few things though:
- Add a readme for the example
- The gif example is nice, but please record just the application window without anything else (to reduce the size) and using a light default theme to match all the other examples. The current gif is too large (1.18mb). 
- Please rename the gif to something like `screenshot.gif`
- Add the example to the main `examples/README.md`

examples/Cargo.toml
```diff
@@ -32,6 +36,10 @@ femtovg-support = ["epoxy", "femtovg", "glow", "libloading"]
 glium-support = ["glium", "epoxy", "libloading"]
 v4_12 = ["gtk/v4_12"]
 v4_14 = ["gtk/v4_14"]
+tokio-support = ["tokio"]
+async-channel-support = ["async-channel"]
+reqwest-support = ["reqwest"]
+serde-support = ["serde"]
```
bilelmoussaoui, line 42: For now, those features are only used by one example. So I would group them into a single feature and just call it `tokio = ["dep:tokio", "async-channel", "reqwest", "serde"]`

examples/tokio_async_request/main.rs
```diff
@@ -0,0 +1,104 @@
...
+        while let Ok(response) = receiver.recv().await {
+            if let Ok(response) = response {
+                for pokemon in response {
+                    let label = gtk::Label::new(Some(&pokemon.name));
+                    label.set_halign(gtk::Align::Start);
+                    list_box.append(&label);
+                }
+            } else {
+                println!("bad request");
+            }
```
bilelmoussaoui, line 92:
```suggestion
            if let Ok(response) = response {
                for pokemon in response {
                    let label = gtk::Label::new(Some(&pokemon.name));
                    label.set_halign(gtk::Align::Start);
                    list_box.append(&label);
                }
            } else {
                println!("bad request");
            }
```
Use a match pattern here and print the error along the `bad request` message

examples/tokio_async_request/main.rs
```diff
@@ -0,0 +1,107 @@
+use glib::clone;
+use gtk::glib::once_cell::sync::Lazy;
```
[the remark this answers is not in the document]
Claudio-code: I couldn't do this and it ends up giving errors because of the listbox, but if you instantiate it within the lambda there is no way for the ScrolledWindow to receive its value. Can you show me an example of what this implementation would look like?

error: 
Screenshot from 2024-01-28 15-23-46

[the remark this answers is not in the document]
Claudio-code: I achieved resolve

sdroege (approved):

bilelmoussaoui:
Looks good to me! Still not a fan of the gif, but I will update that later to have a consistent look across the examples.

Thank you for your contribution

PR #1580, opened 2024-01-27
Get rid of once_cell

bilelmoussaoui:
@Hofer-Julian, would be nice to update the book in the signals chapter as well


Hofer-Julian:
> @Hofer-Julian, would be nice to update the book in the signals chapter as well

Sure, will see if I find time this weekend

PR #1584, opened 2024-01-28
Add missing installation step for Windows

pieterdd:
Initially the instructions [didn't work for me](https://github.com/gtk-rs/gtk4-rs/discussions/1583), but I was able to figure out that I was missing a bit of linker configuration. I documented this requirement and rewrote other parts of the Windows documentation a little bit.

### DIFF
--- a/book/src/installation_windows.md
+++ b/book/src/installation_windows.md
@@ -6,7 +6,7 @@ You will want to go for the GNU toolchain if you depend on libraries that can on
 
 ## Install Rustup
 
-Install the rust toolchain via [rustup](https://rustup.rs/).
+Install the Rust toolchain via [rustup](https://rustup.rs/).
 
 ## Install GTK 4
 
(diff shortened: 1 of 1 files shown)


book/src/installation_windows.md
```diff
@@ -16,26 +16,39 @@ Install the rust toolchain via [rustup](https://rustup.rs/).
...
+
+### Ensure your linker can find GTK's library files
+
+Open `%HOMEPATH%\.cargo\config` for editing. You can add extra flags for your linker here to tell it where GTK's library files are. If you're using LLVM's linker with the `stable-x86_64-pc-windows-msvc` Rust toolchain, your config may look like this:
+
+```toml
+[target.x86_64-pc-windows-msvc]
+linker = "lld-link.exe"
+rustflags = ["-C", "link-args=/LIBPATH:C:\\gtk-build\\gtk\\x64\\release\\lib\\"]
+```
```
Hofer-Julian, line 51: I assume that's only necessary if you use `lld`?
If yes, please make that clear here

Hofer-Julian (commented):
One comment, apart from that, it looks good.
Thanks!

book/src/installation_windows.md
pieterdd: You also need to use it when using the default linker, I just didn't research the appropriate flag to do so. Here's what happens when I remove my `~/.cargo/config`:

```
error: linking with `link.exe` failed: exit code: 1181
(...)
  = note: LINK : fatal error LNK1181: cannot open input file 'gobject-2.0.lib'
```

This works fine for me:

```toml
[target.x86_64-pc-windows-msvc]
rustflags = ["-C", "link-args=/LIBPATH:C:\\gtk-build\\gtk\\x64\\release\\lib\\"]
```

It looks like both linkers support the same flag, so I'll update the documentation accordingly.

Hofer-Julian: I wonder why I didn't have to do that, but oh well.
Let's add it :)

Hofer-Julian (approved):

PR #1588, opened 2024-02-02
gtk::show_about_dialog: Set hide_on_close

### LINKED ISSUE #1544: AboutDialog hangs application on second invocation
I have added a simple Help/About menu item to show an AboutDialog, but have experienced a very weird behavior. The first time I show the dialog it works as expected, but if I show it a second time it will not close. Being a modal dialog, this locks up the whole application.

Simple closure attached to action
```
let action = SimpleAction::new("help-about", None);
    action.connect_activate(clone!(@weak window => move |_action, _parameter| {
        show_help_about(&window);
    }));
    app.add_action(&action);
```
And the function to show is equally simple
```
pub fn show_help_about(window: &Window) {
    let icon = Texture::from_resource(
        "/com/shartrec/kelpie_planner/images/kelpiedog_120x120_transparent.png");
    show_about_dialog(Some(window), &[
        ("program-name", &util::info::PROGRAM_NAME),
        ("version", &util::info::VERSION),
        ("website", &util::info::WEBSITE),
        ("license-type", &util::info::LICENSE_TYPE),
        ("title", &util::info::ABOUT_TITLE),
        ("authors", &[util::info::AUTHOR].as_ref()),
        ("logo", &icon)
        ]);
}
```

That is the standard Rust bindings for GTK4 function as documented here: [show_about_dialog in gtk4 - Rust ](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/fn.show_about_dialog.html)

The following error is emitted to the console when the AboutDialog is shown the second time
```
Gtk-WARNING **: 13:44:27.324: A window is shown after it has been destroyed. This will leave the window in an inconsistent state.
```

felinira:
Closes #1544

We should probably question whether we need this method at all vs. just using a builder directly, but as we already have it I propose to postpone this question to later and fix it for the time being.

### DIFF
--- a/gtk4/src/functions.rs
+++ b/gtk4/src/functions.rs
@@ -236,14 +236,16 @@ pub fn show_about_dialog<P: IsA<Window>>(parent: Option<&P>, properties: &[(&str
                 builder = builder.property(key, *value);
             }
             let about_dialog = builder.build();
-            about_dialog.set_transient_for(parent);
-            about_dialog.set_modal(true);
-            about_dialog.set_destroy_with_parent(true);
+            about_dialog.set_hide_on_close(true);
 
             // cache the dialog if a parent is set
             if let Some(dialog_parent) = parent {
+                about_dialog.set_modal(true);
+                about_dialog.set_transient_for(parent);
+                about_dialog.set_destroy_with_parent(true);
                 dialog_parent.set_qdata(quark, about_dialog.clone());
             }
+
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
we should probably deprecate it. I added it when I was young thinking that I can bind every single API of gtk and there would be no more work to do afterwards :)

PR #1589, opened 2024-02-02
examples: Add example for About Dialog

felinira:
While I was at it I added a small example for using `gtk::AboutDialog`. I didn't find the logo anywhere else, so I guess it needs to live in the example directory. I opted to use the builder instead of `show_about_dialog` because of type safety.


examples/about_dialog/main.rs
```diff
@@ -0,0 +1,44 @@
+use gtk::gdk;
+use gtk::gio;
+use gtk::glib;
+use gtk::prelude::*;
+
+fn main() {
+    let app = gtk::Application::new(None::<&str>, gtk::gio::ApplicationFlags::empty());
```
Hofer-Julian, line 7: Please use the builder pattern here. Also, similar to the other examples give it an app id

examples/about_dialog/main.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    app.connect_activate(|app| {
+        let window = gtk::ApplicationWindow::new(app);
```
Hofer-Julian, line 10: Let's also use the builder here

examples/about_dialog/main.rs
```diff
@@ -0,0 +1,44 @@
...
+
+        let button = gtk::Button::builder()
+            .label("Show About Dialog")
+            .margin_top(24)
+            .margin_bottom(24)
+            .margin_start(24)
+            .margin_end(24)
+            .build();
+
+        let logo_file = gio::File::for_path("./examples/about_dialog/gtk-rs.svg");
```
Hofer-Julian, line 21: This only works if my cwd is correct, right? Maybe embedding into the binary with a macro is more robust

examples/about_dialog/main.rs
```diff
@@ -0,0 +1,44 @@
...
+                .website("https://gtk-rs.org")
+                .license_type(gtk::License::MitX11)
+                .authors(["Author 1", "Author 2"])
+                .logo(&logo)
+                .build();
+
+            dialog.present();
+        }));
+
+        window.set_child(Some(&button));
```
Hofer-Julian, line 39: That statement van also be part of the builder

Hofer-Julian (commented):
Looks good mostly. I had a few minor comments.

examples/about_dialog/main.rs
Hofer-Julian: In that case, I don't mind either way :)

Issue #1590, opened 2024-02-03
Widget's add_binding bindings are a bit off 2


sophie-h:
The variants passed to the closure seems to be nothing than a weird variant of extra-data which are usually dropped from the bindings.

Issue #1542, opened 2023-11-30
Widget's add_binding bindings are a bit off


sophie-h:
One thing is, that it uses bool instead of glib::Propagate

Second thing is, that the variants passed to the closure seems to be nothing than a weird variant of extra-data which are usually dropped from the bindings.

bilelmoussaoui:
> Second thing is, that the variants passed to the closure seems to be nothing than a weird variant of extra-data which are usually dropped from the bindings.

Are you sure about this? =)

bilelmoussaoui:
Fixed in main

bilelmoussaoui:
> [6938071](https://github.com/gtk-rs/gtk4-rs/commit/69380715ea613981b7a3c9dcccd34eb961c8ec7f)
> 
> Did you …forget to stage it?

chances of mistakes increase significantly when you are half asleep and tired, the fix is in some other commit and i had already pushed to main...

bilelmoussaoui:
That comes from gtk::CallbackAction itself, which is not related to this issue which was about the required argument to pass when using `add_binding`

sophie-h:
Created a new issue #1590

PR #1591, opened 2024-02-03
gtk: Don't propogate unused argument

ZanderBrown:
Unlike the C version we are emulating, we dont set any ‘arguments’, and in any case they have little value in a language with native closures

It's therefore plain confusing to require users to accept this second parameter that will *always* be `None`

Fix: https://github.com/gtk-rs/gtk4-rs/issues/1590
See: https://github.com/gtk-rs/gtk4-rs/issues/1542


bilelmoussaoui:
gracias

PR #1644, opened 2024-03-12
Simplify library configuration step for Windows

nazar-pc:
This is what I use in CI and on Windows machines (just did fresh Windows 11 setup). No need to mess with linker flags in configuration files.


Hofer-Julian (approved):
Thanks!

Issue #939, opened 2022-02-19
book: Add chapter about accessibility


Hofer-Julian:

bilelmoussaoui:
@Hofer-Julian we can close this one or ?

Hofer-Julian:
Yes!

PR #1646, opened 2024-03-15
Book: First version of accessibility doc

Claudio-code:
Tried fix https://github.com/gtk-rs/gtk4-rs/issues/939

Added example how uses screen reader to make more easy experience of those who cannot see but I not can make example of high contrast because it works out of box


Hofer-Julian:
Thanks a lot for your contribution @Claudio-code, I appreciate the effort.
However, I don't currently accept whole chapters by external contributors. It is really hard to keep a consistent style otherwise.

Small fixes are always appreciated though :)

Issue #1671, opened 2024-04-02
Segfault in `content_provider_get_value` from dereferencing null `error_ptr`


awused:
**Bug description**

Drag and drop with a ContentProvider that doesn't implement `ContentProviderImpl::value` onto a gtk::Entry in the same process.

**Backtrace**

You can get the backtrace running your program with the environment variable `RUST_BACKTRACE=full`.

```
#0  0x000055acb0b54877 in gdk4::subclass::content_provider::content_provider_get_value<aw_fm::gui::clipboard::imp::ClipboardProvider>
    (ptr=0x55acb58fede0, value_ptr=0x55acb52dbcf0, error_ptr=0x0)
    at /storage/src/third_party/gtk4-rs/gdk4/src/subclass/content_provider.rs:409
409	            *error_ptr = e.into_glib_ptr();
[Current thread is 1 (Thread 0x7f701f2dbec0 (LWP 296838))]
warning: Missing auto-load script at offset 0 in section .debug_gdb_scripts
of file /cache/rust/debug/aw-fm.
Use `info auto-load python-scripts [REGEXP]' to list them.
Missing separate debuginfos, use: dnf debuginfo-install xorg-x11-drv-nvidia-libs-545.29.06-2.fc39.x86_64
(gdb) bt full
#0  0x000055acb0b54877 in gdk4::subclass::content_provider::content_provider_get_value<aw_fm::gui::clipboard::imp::ClipboardProvider>
    (ptr=0x55acb58fede0, value_ptr=0x55acb52dbcf0, error_ptr=0x0)
    at /storage/src/third_party/gtk4-rs/gdk4/src/subclass/content_provider.rs:409
        e = glib::error::Error {inner: glib::boxed::Boxed<glib_sys::GError, glib::error::Error> {inner: core::ptr::non_null::NonNull<glib_sys::GError> {pointer: 0x55acb5a8c080}, _dummy: core::marker::PhantomData<*mut glib::error::Error>}}
        ret = core::result::Result<glib::value::Value, glib::error::Error>::Err(glib::error::Error {inner: glib::boxed::Boxed<glib_sys::GError, glib::error::Error> {inner: core::ptr::non_null::NonNull<glib_sys::GError> {pointer: 0x55acb5a8c080}, _dummy: core::marker::PhantomData<*mut glib::error::Error>}})
[...]

PR #1672, opened 2024-04-02
Correctly handle `NULL` `GError**` out parameters

sdroege:
Fixes https://github.com/gtk-rs/gtk4-rs/issues/1671


bilelmoussaoui (approved):
looks good to me, could merge already but would be nice to get a confirmation from @awused

sdroege:
Well, it's correct in any case but there might be more things broken :)

PR #1678, opened 2024-04-04
Regenerate with latest gir

### LINKED ISSUE #1676: DmabufTextureBuilder has no build function
**Bug description**

Title

bilelmoussaoui:
Also include some missing bits here and there from v4_14. Hoping to do a release this weekend

Fixes #1676


gtk4/src/subclass/im_context.rs
```diff
@@ -67,6 +67,11 @@ pub trait IMContextImpl: IMContextImplExt + ObjectImpl {
     fn activate_osk(&self) {
         self.parent_activate_osk()
     }
+    #[cfg(feature = "v4_14")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "v4_14")))]
+    fn activate_osk_with_event(&self, event: Option<gdk::Event>) -> bool {
```
sdroege, line 72:
```suggestion
    fn activate_osk_with_event(&self, event: Option<&gdk::Event>) -> bool {
```

etc. It's all transfer none, so no need to pass an owned value around

gdk4/src/dmabuf_texture_builder.rs
```diff
@@ -0,0 +1,66 @@
...
+    #[doc(alias = "gdk_dmabuf_texture_builder_build")]
+    #[must_use = "The builder must be built to be used"]
+    #[allow(clippy::missing_safety_doc)]
+    pub unsafe fn build(self) -> Result<Texture, glib::Error> {
+        let mut error = std::ptr::null_mut();
+
+        let result = ffi::gdk_dmabuf_texture_builder_build(
+            self.to_glib_none().0,
+            None,
+            std::ptr::null_mut(),
```
sdroege, line 17: You want to do the same as with the GL texture builder here so that an optional `FnOnce() + Send + 'static` closure can be provided here

sdroege: Oh I should've scrolled down, sorry :)

gdk4/src/dmabuf_texture_builder.rs
```diff
@@ -0,0 +1,66 @@
...
+            }
+        } else {
+            Err(from_glib_full(error))
+        }
+    }
+
+    #[doc(alias = "gdk_dmabuf_texture_builder_build")]
+    #[must_use = "The builder must be built to be used"]
+    #[allow(clippy::missing_safety_doc)]
+    pub unsafe fn build_with_release_func<F: FnOnce() + 'static>(
```
sdroege, line 37:
```suggestion
    pub unsafe fn build_with_release_func<F: FnOnce() + Send + 'static>(
```

I think. Texture are supposed to be thread-safe or not? Same change I guess for the GL texture?

bilelmoussaoui: They are thread safe yes, would do the same for GL textures

PR #1688, opened 2024-04-10
macros: Drop macro-proc-error and upgrade syn to 2.0

### LINKED ISSUE #1409: Bump syn usage in gtk-macros

zecakeh:
This can be reviewed per commit.

I rewrote a bunch of stuff in attribute_parser.rs because I found it easier than to port the existing code. I could have rewrote more to be simpler but that would have involved changing the types that are exported from the module, which would have involved changes in other parts too. It can probably be done in a following PR.

This almost gets rid of the anyhow dependency in the macros crate, because of the need to use `syn::Error::into_compile_error`. It can probably be completely removed in a following PR, since it has only a single use and can be easily replaced.

Supersedes #1422.\
Closes #1409.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -392,7 +392,7 @@ checksum = "87750cf4b7a4c0625b1529e4c543c2182106e4dedc60a2a6455e00d212c489ac"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.57",
+ "syn",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


bilelmoussaoui:
Can't review right now but just wanted to say thanks a lot for the patch, so happy to this finally happening!

PR #1705, opened 2024-04-20
gtk4: Manually implement `GraphicsOffload` constructor for now

sdroege:
Needs to be transfer-none instead of transfer-full


sdroege:
Should get this in so it can be backported without pulling in all the new GTK 4.16 API at once too, and once we updated the .gir files this can be reverted.

PR #1707, opened 2024-04-21
Fix typo in todo_1.md

exi:
Text talksa bout Window::model when the code actually adds Window::tasks


Hofer-Julian (approved):
Nice find, thanks!

PR #1708, opened 2024-04-21
macros: Drop anyhow dependency

zecakeh:
Since #1688 was merged, it is used in a single place and it is easy to remove, so it doesn't seem worth it to keep depending on it.


bilelmoussaoui:
Thanks a lot!

PR #1719, opened 2024-04-27
Simplify reading file contents to a String

mjgarton:
This uses a simpler approach to reading the file contents to a String, making the code slightly more readable by removing code unrelated to the point of the example.


sdroege (approved):
I would've just `use std::fs` and then `fs::read_to_string()` to make it clearer where the function comes from but either way is fine.

This was indeed unnecessarily convoluted :)

PR #1789, opened 2024-07-04
gtk: Implementable implementation for AccessibleText

felinira:


bilelmoussaoui:
I don't have the time to review this right now but two questions:
- is it important for you to have this in the next release or waiting till a future .x would be okay?
- do you think you could add an example of how it could be used?

gtk4/src/accessible_text_range.rs
```diff
@@ -1,17 +1,24 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-#[derive(Copy, Clone)]
-#[doc(alias = "GtkAccessibleTextRange")]
-#[repr(transparent)]
-pub struct AccessibleTextRange(crate::ffi::GtkAccessibleTextRange);
+use glib::translate::UnsafeFrom;
```
bilelmoussaoui, line 3:
```suggestion
use glib::translate::*;
```

gtk4/src/accessible_text_range.rs
```diff
@@ -1,17 +1,24 @@
...
+use gtk4_sys::GtkAccessibleTextRange;
```
bilelmoussaoui, line 4: You import it, then only use it in certain places... Usually we do `use crate::ffi;` and call ffi types with `ffi::$Type` or functions as well.

gtk4/src/accessible_text_range.rs
```diff
@@ -1,17 +1,24 @@
...
+
+glib::wrapper! {
+    #[doc(alias = "GtkAccessibleTextRange")]
+    pub struct AccessibleTextRange(BoxedInline<crate::ffi::GtkAccessibleTextRange>);
+}
```
bilelmoussaoui, line 9: Should be moved to a separate commit.

gtk4/src/subclass/accessible_text.rs
```diff
@@ -0,0 +1,533 @@
...
+//! Traits intended for implementing the [`AccessibleText`](crate::AccessibleText) interface.
+
+use std::mem::MaybeUninit;
+
+use crate::accessible_text_range::AccessibleTextRange;
+use crate::subclass::prelude::*;
+use crate::{ffi, AccessibleText, AccessibleTextGranularity};
+use glib::object::Cast;
+use glib::translate::*;
+use glib::GString;
```
bilelmoussaoui, line 13:
```suggestion
use crate::{ffi,subclass::prelude::*, AccessibleText,AccessibleTextRange, prelude::*, AccessibleTextGranularity};
use glib::{translate::*, GString};
```

gtk4/src/subclass/accessible_text.rs
```diff
@@ -0,0 +1,533 @@
...
+
+    fn parent_contents(&self, start: u32, end: u32) -> Option<glib::Bytes> {
+        unsafe {
+            let type_data = Self::type_data();
+            let parent_iface = type_data.as_ref().parent_interface::<AccessibleText>()
+                as *const ffi::GtkAccessibleTextInterface;
+
+            let func = (*parent_iface)
+                .get_contents
+                .expect("no parent \"get_contents\" implementation");
```
bilelmoussaoui, line 150: Is this required? Usually we return None if there is no override/default implementation

gtk4/src/subclass/accessible_text.rs
```diff
@@ -0,0 +1,533 @@
...
+        *attribute_values = c_values.into_glib_ptr();
+    }
+}
+
+#[cfg(feature = "v4_16")]
+#[cfg_attr(docsrs, doc(cfg(feature = "v4_16")))]
+unsafe extern "C" fn accessible_text_get_extents<T: AccessibleTextImpl>(
+    accessible_text: *mut ffi::GtkAccessibleText,
+    start: libc::c_uint,
+    end: libc::c_uint,
```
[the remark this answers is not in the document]
bilelmoussaoui: we only use libc if it is a mut ptr

gtk4/src/accessible_text_range.rs
```diff
@@ -1,17 +1,24 @@
...
 
 impl AccessibleTextRange {
+    pub fn new(start: usize, length: usize) -> Self {
+        unsafe { AccessibleTextRange::unsafe_from(GtkAccessibleTextRange { start, length }) }
+    }
+
     pub fn start(&self) -> usize {
-        self.0.start
+        self.inner.start
```
sdroege, line 17: How about setters? Unnecessary?

gtk4/src/subclass/accessible_text.rs
```diff
@@ -0,0 +1,533 @@
...
+        self.parent_offset(point)
+    }
+
+    #[doc(alias = "get_selection")]
+    fn selection(&self) -> Vec<AccessibleTextRange> {
+        self.parent_selection()
+    }
+}
+
+mod sealed {
```
sdroege, line 65: We don't need that anymore, do we?

gtk4/src/subclass/accessible_text.rs
```diff
@@ -0,0 +1,533 @@
...
+            let is_set: bool = from_glib(func(
+                self.obj()
+                    .unsafe_cast_ref::<AccessibleText>()
+                    .to_glib_none()
+                    .0,
+                offset,
+                n_ranges.as_mut_ptr(),
+                &mut ranges,
+                &mut attribute_names,
+                &mut attribute_values,
```
sdroege, line 95: I don't think there are necessarily `n_ranges` attribute names/values. Both are just `NULL`-terminated string arrays. I'd assume both have the same length but otherwise they should be independent from the ranges

gtk4/src/subclass/accessible_text.rs
```diff
@@ -0,0 +1,533 @@
...
+            func(
+                self.obj()
+                    .unsafe_cast_ref::<AccessibleText>()
+                    .to_glib_none()
+                    .0,
+            )
+        }
+    }
+
+    fn parent_contents(&self, start: u32, end: u32) -> Option<glib::Bytes> {
```
sdroege, line 142: I think this must not return `NULL`? Or at least the C functions is not annotated as such

gtk4/src/subclass/accessible_text.rs
```diff
@@ -0,0 +1,533 @@
...
+                    .unsafe_cast_ref::<AccessibleText>()
+                    .to_glib_none()
+                    .0,
+                offset,
+                granularity.into_glib(),
+                start.as_mut_ptr(),
+                end.as_mut_ptr(),
+            );
+
+            if !bytes.is_null() {
```
sdroege, line 191: Not marked as nullable either

gtk4/src/subclass/accessible_text.rs
```diff
@@ -0,0 +1,533 @@
...
+unsafe extern "C" fn accessible_text_get_contents<T: AccessibleTextImpl>(
+    accessible_text: *mut ffi::GtkAccessibleText,
+    start: libc::c_uint,
+    end: libc::c_uint,
+) -> *mut glib::ffi::GBytes {
+    let instance = &*(accessible_text as *mut T::Instance);
+    let imp = instance.imp();
+
+    let contents = imp.contents(start, end);
+    contents.to_glib_full()
```
sdroege, line 361:
```suggestion
    contents.into_glib_ptr()
```

also for all other `to_glib_full()` here

gtk4/src/subclass/accessible_text.rs
```diff
@@ -0,0 +1,533 @@
...
+        if !start.is_null() {
+            *start = r_start;
+        }
+        if !end.is_null() {
+            *end = r_end;
+        }
+
+        bytes.to_glib_full()
+    } else {
+        std::ptr::null_mut()
```
sdroege, line 386: Not marked as nullable in C

gtk4/src/subclass/accessible_text.rs
```diff
@@ -0,0 +1,533 @@
...
+    end: libc::c_uint,
+    extents: *mut graphene::ffi::graphene_rect_t,
+) -> glib::ffi::gboolean {
+    let instance = &*(accessible_text as *mut T::Instance);
+    let imp = instance.imp();
+
+    let rect = imp.extents(start, end);
+
+    if let Some(rect) = rect {
+        std::ptr::copy_nonoverlapping(rect.into_glib_ptr(), extents, 1);
```
sdroege, line 502:
```suggestion
        *extents = *rect.as_ptr();
```

or so? `into_glib_ptr()` does a new heap allocation that you leak here

examples/accessible_text/main.rs
```diff
@@ -0,0 +1,24 @@
...
+    application.run()
+}
+
+fn build_ui(application: &gtk::Application) {
+    let window = gtk::ApplicationWindow::new(application);
+
+    window.set_title(Some("Accessible Text Example"));
+    window.set_default_size(260, 140);
+
+    let text_view: AccessibleTextView = glib::Object::new();
```
bilelmoussaoui, line 20:
```suggestion
    let text_view = glib::Object::new::<AccessibleTextView>();
```
turbo fish for the win

examples/accessible_text/text_view.rs
```diff
@@ -0,0 +1,113 @@
+use gtk::glib;
+use gtk::subclass::prelude::*;
+
+mod imp {
+    use glib::GStr;
+    use gtk::{
+        ffi::{GTK_ACCESSIBLE_ATTRIBUTE_OVERLINE, GTK_ACCESSIBLE_ATTRIBUTE_OVERLINE_SINGLE},
```
bilelmoussaoui, line 7: why using ffi? we don't have a safe way to set those?

bilelmoussaoui: actually they are available from `gtk::ACCESSIBLE_ATTRIBUTE_OVERLINE` and so on

bilelmoussaoui:
> Maybe two wrapper objects for (range, attr, value) triples wouldn't be such a bad idea. When writing the example it was very cumbersome to change attributes, adding attributes by constants, etc.

I think we can do that as a second step :) but yes, having to deal with constants is not ideal. But also not sure how we can do something that wouldn't be too far away from the original API.

Maybe we can make those attributes an enum? then you would only have a `Vec<(Range, TextAttribute)>`

gtk4/src/subclass/accessible_text.rs
bilelmoussaoui: Would it be too much to ask if you could fix those missing nullable annotations? otherwise i can take care of that :)

PR #1860, opened 2024-10-10
Book main function compiler error

redvimo:
// Suggesting this line at the end of main function,i get a compilation error without it
    return glib::ExitCode::SUCCESS;


bilelmoussaoui:
app.run() already returns glib::ExitCode https://gtk-rs.org/gtk-rs-core/stable/latest/docs/gio/prelude/trait.ApplicationExtManual.html#method.run

redvimo:
Thank you. It was my error, I had a semicolon after `app.run()`.

PR #1866, opened 2024-10-17
book: Make book listings work on macos by using glib user_data_dir

daxhuiberts:
`dirs::data_dir()` returns a different directory that glib expects for the schemas on macos (see https://docs.rs/dirs/latest/dirs/fn.data_dir.html)
Using `glib::user_data_dir()` instead will return the expected directory for glib.

### DIFF
--- a/book/listings/Cargo.lock
+++ b/book/listings/Cargo.lock
@@ -370,27 +370,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "dirs"
-version = "5.0.1"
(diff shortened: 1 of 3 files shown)


bilelmoussaoui:
Is the dirs crate still used elsewhere? If not, maybe remove it completely

daxhuiberts:
> Is the dirs crate still used elsewhere? If not, maybe remove it completely

Good suggestion. It was only used in the xtask, so I removed it.

bilelmoussaoui:
lgtm, thanks!

PR #1867, opened 2024-10-17
book: Include ashpd only on linux so that listings run on macos

daxhuiberts:
This will make `ashpd` only a dependency on linux.

An alternative would be something like `cfg(all(unix, not(target_os = "macos")))` to still include it for all other unix family.

### DIFF
--- a/book/listings/main_event_loop/7/main.rs
+++ b/book/listings/main_event_loop/7/main.rs
@@ -1,5 +1,3 @@
-use ashpd::desktop::account::UserInformation;
-use ashpd::WindowIdentifier;
 use glib::clone;
 use gtk::prelude::*;
 use gtk::{glib, Application, ApplicationWindow, Button};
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
The code using it should also be behind a similar guard

daxhuiberts:
> The code using it should also be behind a similar guard

I looked into disabling the `main_event_loop_7` bin, but there currently is no simple way to achieve that (see https://github.com/rust-lang/cargo/issues/9208)
Not sure what the appropriate approach would be in this case.

daxhuiberts:
I made a change to `main_event_loop_7` with a possible solution, but this approach is probably not desirable.

book/listings/main_event_loop/7/main.rs
```diff
@@ -65,3 +49,30 @@ fn build_ui(app: &Application) {
...
+        println!("User name: {}", response.name());
+    } else {
+        println!("Could not access user information.")
+    }
+}
+
+#[cfg(not(target_os = "linux"))]
+async fn fetch_user_information(_button: Button) {
+    println!("fetching user information not available outside target_os = \"linux\"");
+}
```
bilelmoussaoui, line 78: Missing newline at the end of the file

bilelmoussaoui:
Also, please fix the formatting, otherwise the suggested solution is the best we can do so it is fine.

daxhuiberts:
Thanks for the feedback.
Fixed the newline and formatting.

bilelmoussaoui:
Thanks!

PR #1868, opened 2024-10-18
Regenerate with dropped sealed auto traits

bilelmoussaoui:


sdroege:
You'll want to do the same with the manual traits

bilelmoussaoui:
> You'll want to do the same with the manual traits

Without sublassing traits, right? as that is handled in the other PR

sdroege:
Yes

PR #1870, opened 2024-10-18
gtk/templates: Add missing traits for TemplateChild

### LINKED ISSUE #1842: TemplateChild doesn't work as property
**Bug description**

When using this minimal example I get

> the trait bound `gtk4::TemplateChild<gtk4::Widget>: ToValue` is not satisfied

I need to specify the property as

`#[property(get = |s: &&Self| (*s.test_widget).clone(), type = gtk4::Widget)]`

to make it work.

**test_widget.rs**

```rust
use gtk4::glib;
use gtk4::prelude::*;
use gtk4::subclass::prelude::*;

mod imp {
    use super::*;

    #[derive(Default, glib::Properties, gtk4::CompositeTemplate)]
    #[template(file = "test_widget.ui")]
    #[properties(wrapper_type = super::TestWidget)]
    pub struct TestWidget {
        #[property(get)]
        #[template_child]
        test_widget: TemplateChild<gtk4::Widget>,
    }

    #[glib::object_subclass]
    impl ObjectSubclass for TestWidget {
        const NAME: &'static str = "TestWidget";
        type Type = super::TestWidget;
        type ParentType = gtk4::Widget;

            klass.bind_template();
        }

        fn instance_init(obj: &glib::subclass::InitializingObject<Self>) {
            obj.init_template();
        }
    }

    #[glib::derived_properties]
    impl ObjectImpl for TestWidget {}
    impl WidgetImpl for TestWidget {}
    impl TestWidget {}
}

glib::wrapper! {
    pub struct TestWidget(ObjectSubclass<imp::TestWidget>)
    @extends gtk4::Widget;
}
```

**test_widget.ui**
```xml
<interface>
  <template class="TestWidget" parent="GtkWidget">
    <child>
      <object class="GtkWidget" id="widget" />
    </child>
  </template>
</interface>
```

Apparently this worked once, considering #1495 was merged, but something must have either regressed, or it never really worked after all :smile:

bilelmoussaoui:
Fixes #1842

Only compile tested

cc @felinira @sdroege

### DIFF
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -1211,12 +1211,13 @@ pub unsafe trait WidgetClassExt: ClassStruct {
 unsafe impl<T: ClassStruct> WidgetClassExt for T where T::Type: WidgetImpl {}
 
 #[derive(Debug, PartialEq, Eq)]
-#[repr(transparent)]
+#[repr(C)]
 pub struct TemplateChild<T>
 where
     T: ObjectType + FromGlibPtrNone<*mut <T as ObjectType>::GlibType>,
 {
     ptr: *mut <T as ObjectType>::GlibType,
+    should_drop: bool,
 }
 
 impl<T> Default for TemplateChild<T>
@@ -1228,6 +1229,7 @@ where
 
         Self {
             ptr: std::ptr::null_mut(),
+            should_drop: false,
         }
     }
 }
@@ -1242,6 +1244,44 @@ where
 
     fn param_spec_builder() -> Self::BuilderFn {
         Self::ParamSpec::builder
+    }
+}
+
+impl<T> ToValue for TemplateChild<T>
+where
+    T: ToValue + ObjectType + FromGlibPtrNone<*mut <T as ObjectType>::GlibType>,
+{
+    #[inline]
+    fn to_value(&self) -> glib::Value {
+        T::to_value(&self.get())
+    }
+
+    #[inline]
+    fn value_type(&self) -> glib::Type {
+        T::static_type()
+    }
+}
+
+impl<T> glib::value::ValueType for TemplateChild<T>
+where
+    T: glib::value::ValueType + ObjectType + FromGlibPtrNone<*mut <T as ObjectType>::GlibType>,
+{
+    type Type = <T as glib::value::ValueType>::Type;
+}
+
+unsafe impl<'a, T> glib::value::FromValue<'a> for TemplateChild<T>
+where
+    T: ObjectType + FromGlibPtrNone<*mut <T as ObjectType>::GlibType>,
+{
+    type Checker = glib::value::GenericValueTypeChecker<T>;
+
+    #[inline]
+    unsafe fn from_value(value: &'a glib::Value) -> Self {
+        skip_assert_initialized!();
+        TemplateChild {
+            ptr: T::from_value(value).into_glib_ptr(),
+            should_drop: true,
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


sdroege (approved):

gtk4/src/subclass/widget.rs
```diff
@@ -1245,6 +1245,43 @@ where
...
+where
+    T: ObjectType + FromGlibPtrNone<*mut <T as ObjectType>::GlibType>,
+{
+    type Checker = glib::value::GenericValueTypeChecker<T>;
+
+    #[inline]
+    unsafe fn from_value(value: &'a glib::Value) -> Self {
+        skip_assert_initialized!();
+        TemplateChild {
+            ptr: T::from_value(value).as_ptr(),
```
sdroege, line 1280: Actually no, this looks wrong. The pointers becomes invalid after this call, you need `into_glib_ptr()` or so but then `TemplateChild` has no `Drop`...

bilelmoussaoui: Right because Gtk is keeping the ref of the child and drops it itself. If we do implement Drop, we would have to some bool to identify whether we should unref the ptr or not. Sounds a bit hackish but no ideas how to do this properly :)

sdroege: Do we need the `FromValue` impl?

bilelmoussaoui: Yes. Otherwise it can't really be a property

sdroege: Do we want writable properties to be possible for this? Seems strange to me. Readonly seems to make sense.

bilelmoussaoui: Well, you want to have writable properties yes. All container widgets nowadays have a "child" property that is readwrite.

sdroege: Oh I see. Well, then you'll need a drop boolean :)

bilelmoussaoui: Done that in the latest push

gtk4/src/subclass/widget.rs
```diff
@@ -1211,12 +1211,12 @@ pub unsafe trait WidgetClassExt: ClassStruct {
 unsafe impl<T: ClassStruct> WidgetClassExt for T where T::Type: WidgetImpl {}
 
 #[derive(Debug, PartialEq, Eq)]
-#[repr(transparent)]
```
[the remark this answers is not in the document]
sdroege: I hope nothing was relying on it being the same representation as `T` :)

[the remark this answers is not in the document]
sdroege: Maybe this needs to be repr-C now so that we at least guarantee that the pointer is stored at the beginning? How is GtkBuilder filling this?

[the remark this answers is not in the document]
bilelmoussaoui: > Maybe this needs to be repr-C now so that we at least guarantee that the pointer is stored at the beginning? How is GtkBuilder filling this?

By calling gtk_widget_class_bind_template_child_full, which is used from rust with bind_template_child_with_offset. But yes, maybe it should be repr-C.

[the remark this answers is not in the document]
sdroege: Yes, it currently works with the offset to `TemplateChild` so you need to make it repr-C at least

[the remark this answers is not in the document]
bilelmoussaoui: Done, let us wait and see if @felinira got whatever use case they have working with this patch first. I tested it locally by changing one of the examples but still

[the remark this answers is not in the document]
bilelmoussaoui: > I'm a little surprised why this needs a FromValue implementation though. Isn't this readonly?

The requirement seems to come from `ObjectExt::property`
```rust
fn property<V: for<'b> FromValue<'b> + 'static>(&self, property_name: &str) -> V;
```

So it is expected? not sure why you are surprised about it, unless you meant the ToValue? in such case, FromValue, requires ValueType, which itself requires ToValue....

bilelmoussaoui:
As this fixes the issue, let us land it already, any further fixes could be done separately

PR #1886, opened 2024-10-24
template_child: Implement PropertyGet for TemplateChild

felinira:
This circumvents the drop issue for now by just not allowing property setters at all. Then we can have the field be transparent again. :)

As we don't have public setters on TemplateChild it creates an inconsistency where setting via Property would work, but setting the field directly wouldn't.

I think we should either:

* Generally allow setting the value of TemplateChild via a public set method
* Not allow it at all (which this opts to do)

This is technically breaking, because it unimplements HasParamSpec, although I can't really see any reason to rely on this in application code. :woman_shrugging:

### DIFF
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -5,7 +5,13 @@
 
 use std::{boxed::Box as Box_, collections::HashMap, fmt, future::Future};
 
-use glib::{clone::Downgrade, subclass::SignalId, translate::*, GString, Variant};
+use glib::{
+    clone::Downgrade,
+    property::{Property, PropertyGet},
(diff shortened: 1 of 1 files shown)


bilelmoussaoui (approved):
As I mentioned on the Matrix channel, I don't use this feature on any of the apps I maintain, nor have an idea about the expectations from GTK side on whether you can modify the pointer that gets initialized with gtk_init_template.

So I am fine with merging this change. It would be nice to confirm if template children widgets as properties can not be writable before doing so though.

sdroege (approved):

sdroege:
Your call but I think this makes sense. You could implement `PropertySet` if you want writable properties at a later time

PR #1903, opened 2024-11-08
0.9 backports

sdroege:


bilelmoussaoui:
I would wait till we land the latest gir files before doing a release as it contains some annotations fixes

sdroege:
Ok, then let's wait for that also for gtk-rs-core

bilelmoussaoui:
https://github.com/gtk-rs/gtk4-rs/pull/1907 would need to be backported as well

sdroege:
Done

PR #1909, opened 2024-11-13
add gdk4-macos bindings

bilelmoussaoui:
This still needs moving the gdk4-macos gir file to the gir-files repository. Which requires the gtk merge request to land first and then some sort of script to fetch it from GTK's gitlab CI.


gsk4/src/auto/path_point.rs
```diff
@@ -85,7 +85,6 @@ impl PartialOrd for PathPoint {
         Some(self.cmp(other))
     }
 }
-
```
nacho, line 88: why the removal of these empty lines?

nacho:
LGTM. Thanks @bilelmoussaoui

gsk4/src/auto/path_point.rs
bilelmoussaoui: fixed

PR #1938, opened 2024-12-09
gdk4-macos: manually implement native_window method

nacho:
For now we need to depend on the cocoa crate in order to implement this method since it is the only decent crate implementing the cocoa api.

### DIFF
--- a/gdk4-macos/Cargo.toml
+++ b/gdk4-macos/Cargo.toml
@@ -15,6 +15,7 @@ version.workspace = true
 
 [features]
 v4_8 = ["gdk4-macos-sys/v4_8"]
+cocoa = ["dep:cocoa"]
 
 [dependencies]
 gdk4-macos-sys.workspace = true
@@ -22,6 +23,7 @@ gdk.workspace = true
 gio.workspace = true
 glib.workspace = true
 libc.workspace = true
+cocoa = { version = "0.26", default-features = false, optional = true }
 
 [dev-dependencies]
 gir-format-check.workspace = true
--- a/gdk4-macos/Gir.toml
+++ b/gdk4-macos/Gir.toml
@@ -17,7 +17,6 @@ generate = [
     "GdkMacos.MacosGLContext",
     "GdkMacos.MacosKeymap",
     "GdkMacos.MacosSeat",
-    "GdkMacos.MacosSurface",
 ]
 
 manual = [
@@ -37,3 +36,14 @@ status = "generate"
     [[object.function]]
     name = "get_geometry"
     ignore = true # The function does not exists
+
+[[object]]
+name = "GdkMacos.MacosSurface"
+status = "generate"
+    [[object.function]]
+    name = "get_native_window"
+    manual = true
+    rename = "native"
+    [[object.property]]
+    name = "native"
+    generate = ["notify"]
--- a/gdk4-macos/src/auto/macos_surface.rs
+++ b/gdk4-macos/src/auto/macos_surface.rs
@@ -20,18 +20,6 @@ glib::wrapper! {
 }
 
 impl MacosSurface {
-    //#[cfg(feature = "v4_8")]
-    //#[cfg_attr(docsrs, doc(cfg(feature = "v4_8")))]
-    //#[doc(alias = "gdk_macos_surface_get_native_window")]
-    //#[doc(alias = "get_native_window")]
-    //pub fn native_window(&self) -> /*Unimplemented*/Option<Basic: Pointer> {
-    //    unsafe { TODO: call ffi:gdk_macos_surface_get_native_window() }
-    //}
-
-    //pub fn native(&self) -> /*Unimplemented*/Basic: Pointer {
-    //    ObjectExt::property(self, "native")
-    //}
-
     #[doc(alias = "native")]
     pub fn connect_native_notify<F: Fn(&Self) + 'static>(&self, f: F) -> SignalHandlerId {
         unsafe extern "C" fn notify_native_trampoline<F: Fn(&MacosSurface) + 'static>(
--- a/gdk4-macos/src/macos_surface.rs
+++ b/gdk4-macos/src/macos_surface.rs
@@ -1,0 +1,29 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+#[cfg(feature = "v4_8")]
+use crate::ffi;
+#[cfg(not(feature = "v4_8"))]
+use crate::prelude::*;
+use crate::{id, MacosSurface};
+#[cfg(feature = "v4_8")]
+use glib::translate::*;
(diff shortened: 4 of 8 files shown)


gdk4-macos/src/auto/macos_surface.rs
```diff
@@ -20,14 +20,6 @@ glib::wrapper! {
...
 impl MacosSurface {
-    //#[cfg(feature = "v4_8")]
-    //#[cfg_attr(docsrs, doc(cfg(feature = "v4_8")))]
-    //#[doc(alias = "gdk_macos_surface_get_native_window")]
-    //#[doc(alias = "get_native_window")]
-    //pub fn native_window(&self) -> /*Unimplemented*/Option<Basic: Pointer> {
-    //    unsafe { TODO: call ffi:gdk_macos_surface_get_native_window() }
-    //}
-
     //pub fn native(&self) -> /*Unimplemented*/Basic: Pointer {
```
bilelmoussaoui, line 23: This property has to be ignored as well, and you can use it as a fallback for pre v4.8

gdk4-macos/Cargo.toml
```diff
@@ -22,6 +22,7 @@ gdk.workspace = true
 gio.workspace = true
 glib.workspace = true
 libc.workspace = true
+cocoa = "0.26"
```
bilelmoussaoui, line 25: Same like we do for all external crates integrations, it has to be optional.

gdk4-macos/src/auto/macos_surface.rs
bilelmoussaoui: basically getting the property as a pointer and doing the same as you do when calling gdk_macos_surface_get_native_window

gdk4-macos/Cargo.toml
```diff
@@ -15,13 +15,15 @@ version.workspace = true
 
 [features]
 v4_8 = ["gdk4-macos-sys/v4_8"]
+cocoa = ["cocoa"]
```
bilelmoussaoui, line 18:
```suggestion
cocoa = ["dep:cocoa"]
```
or so, the new feature has to be also added to https://github.com/gtk-rs/gtk4-rs/tree/main/gdk4-macos#features

gdk4-macos/src/auto/macos_surface.rs
nacho: Fixed

gdk4-macos/Cargo.toml
nacho: ah! nice thanks

gdk4-macos/Gir.toml
```diff
@@ -36,4 +35,14 @@ name = "GdkMacos.MacosMonitor"
...
+
+[[object]]
+name = "GdkMacos.MacosSurface"
+status = "generate"
+    [[object.function]]
+    name = "get_native_window"
+    manual = true
+    [[object.property]]
+    name = "native"
+    ignore = true # Same as native_window
```
bilelmoussaoui, line 48:
```suggestion
    generate = ["notify"]
```
and add the getter manually. The cocoa helper, when the future is not enabled, should also be there but giving you a pointer instead of the cocoa id type.

gdk4-macos/Cargo.toml
nacho: done

gdk4-macos/Gir.toml
nacho: do you think we should just rename native_window to native?

nacho: done

gdk4-macos/Cargo.toml
nacho: done

examples/custom_orientable/custom_orientable/imp.rs
```diff
@@ -62,7 +62,7 @@ impl ObjectImpl for CustomOrientable {
     }
 
     fn dispose(&self) {
-        // Child widgets need to be manually unparented in `dispose()`.
+        // Child widgets need to be manually parentless in `dispose()`.
```
bilelmoussaoui, line 65: this should be un-parented instead.

gdk4-macos/src/macos_surface.rs
```diff
@@ -0,0 +1,26 @@
...
+use cocoa::base::id;
+#[cfg(feature = "v4_8")]
+use glib::translate::*;
+
+impl MacosSurface {
+    #[cfg(feature = "v4_8")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "v4_8")))]
+    #[doc(alias = "gdk_macos_surface_get_native_window")]
+    #[doc(alias = "get_native_window")]
+    pub fn native(&self) -> Option<id> {
```
bilelmoussaoui, line 16: This still doesn't have two fallbacks.
- If the cocoa feature is disabled, you should have this function that gives you access to the raw pointer.
- If you don't have v4_8 enabled, this function should go through ObjectExt::property and give you the native window.

examples/custom_orientable/custom_orientable/imp.rs
nacho: the tool was asking explicitely for parentless :D which seemed weird but I did what the tool wanted blindly. I'll fix it

gdk4-macos/src/macos_surface.rs
nacho: I think this should be now ready

bilelmoussaoui (approved):

bilelmoussaoui:
The changes looks good to me, but the docs CI as being run on a linux machine seems to fail at linking cocoa, which is weird as it is an optional feature...

nacho:
> The changes looks good to me, but the docs CI as being run on a linux machine seems to fail at linking cocoa, which is weird as it is an optional feature...

I wonder if it has something to do with the fact that the feature name is cocoa and the dependency is also cocoa...

sdroege:
The docs build uses `--all-features`, that's why

nacho:
I wonder how this works for windows then... checking

gdk4-macos/Cargo.toml
```diff
@@ -15,13 +15,15 @@ version.workspace = true
 
 [features]
 v4_8 = ["gdk4-macos-sys/v4_8"]
+cocoa = ["dep:cocoa"]
```
sdroege, line 18: I think this is unnecessary but doesn't hurt for making it more explicit (every optional dependency is automatically also a feature)

bilelmoussaoui: I prefer them being explicit personally :)

sdroege (approved):

gdk4-macos/Cargo.toml
sdroege: Same

PR #1939, opened 2024-12-10
Update per latest gir/gir-files

bilelmoussaoui:


CHANGELOG.md
```diff
@@ -93,6 +93,7 @@ yuraiz:
 ```rust
 #[derive(Debug, Default, gtk::CompositeTemplate)]
 #[template(string = "
+using Gtk 4.0;
```
sdroege, line 96: This is of course a breaking change so we can't backport it. Unless you think nobody is using the blueprint support :)

PR #1945, opened 2024-12-18
0.9 backports

sdroege:


gtk4-macros/src/blueprint.rs
```diff
@@ -1,28 +1,36 @@
...
         .stdout(Stdio::piped())
         .spawn()
-        .unwrap_or_else(|_| panic!("blueprint-compiler not found"));
-
+        .map_err(|e| format!("blueprint-compiler couldn't be spawned: {e}"))?;
     let mut stdin = compiler.stdin.take().unwrap();
-    stdin.write_all(b"using Gtk 4.0;\n")?;
-    stdin.write_all(blueprint)?;
+    if let Err(e) = stdin
+        .write_all(b"using Gtk 4.0;\n")
```
sdroege, line 17: @bilelmoussaoui This still inserts the line for backwards compat. Your call but removing this seems like it would break existing code

bilelmoussaoui: i wouldn't remove it, no

PR #2013, opened 2025-03-18
book: describe custom structs and multithreading

charlesrocket:


Hofer-Julian:
Hey @charlesrocket, thanks for your contribution.

I appreciate the work you put into this, but I don't think I am going to merge it. We already cover channels which are preferred over mutexes in most situations.

charlesrocket:
The book is not very helpful when dealing with custom structs, so that was the reason for writing this example.

PR #2014, opened 2025-03-22
Update container to adwaita to 1.7

swsnr:


swsnr:
Out of curiosity, why does the container image specify the version just for Adwaita?  Gtk just builds of main, as far as I can see...

bilelmoussaoui:
> Out of curiosity, why does the container image specify the version just for Adwaita? Gtk just builds of main, as far as I can see...

because this is only meant for the usage in the gtk4-rs book. Not for anyone's CI.

swsnr:
Fair enough, then I'll just build my own image :shrug:

PR #2026, opened 2025-04-06
Adjust install of `ashpd` to avoid `tokio`

Jwashton:
Since version 0.10 `ashpd` has shipped with the `tokio` feature enabled by default. This slightly undermines this section of the book where the author is discussing non-tokio ways to run async code.

This commit adjusts the command for adding `ashpd` to exclude `tokio` and use the previously expected `async-std` feature instead.

Without this change, the example code compiles but panics with a `there is no reactor running` message. With this change, the example code runs successfully.


Hofer-Julian:
Thanks @Jwashton !

PR #2031, opened 2025-04-09
femtovg area : scale by the scale factor to fill the screen

### LINKED ISSUE #2015: femtovg example uses half the screen on 2x scaling
**Bug description**

Run 
```
cargo run --bin femtovg_area --features="femtovg-support
```
on windows with 175% of 200% scaling on windows 11.

Upon resizing, there are also artefacts on the 3/4 of the canvas not used.

https://github.com/user-attachments/assets/b1289b1e-a7fa-4326-bf53-3d722107dba6

Doublonmousse:
fixes #2015

### DIFF
--- a/examples/femtovg_area/femtovg_area/imp.rs
+++ b/examples/femtovg_area/femtovg_area/imp.rs
@@ -89,8 +89,9 @@ impl GLAreaImpl for FemtoVGArea {
         let canvas = canvas.as_mut().unwrap();
 
         let area = self.obj();
-        let w = area.width() as u32;
-        let h = area.height() as u32;
+        let scale_factor = self.obj().scale_factor();
+        let w = (area.width() * scale_factor) as u32;
+        let h = (area.height() * scale_factor) as u32;
         canvas.reset_transform();
         canvas.clear_rect(0, 0, w, h, Color::rgba(0, 0, 0, 0));
 


examples/femtovg_area/femtovg_area/imp.rs
```diff
@@ -89,8 +89,9 @@ impl GLAreaImpl for FemtoVGArea {
         let canvas = canvas.as_mut().unwrap();
 
         let area = self.obj();
-        let w = area.width() as u32;
-        let h = area.height() as u32;
+        let scale_factor = self.obj().scale_factor();
```
sdroege, line 92: Like I said in the issue, this should really use the (fractional) scale factor of the GDK surface. That this doesn't work is either a bug in GTK or something else is wrong here.

Doublonmousse: Okay, that's enough.

I'm beyond fed up of you two @sdroege and @bilelmoussaoui disagreeing on every single point and being on the receiving end of your opposite comments (that ultimately ends with no resolution). This is incredibly frustrating and make me want to never contribute to your project again.

Another example here: https://gitlab.gnome.org/World/Rust/poppler-rs/-/issues/1#note_2242804 vs https://github.com/gtk-rs/gir/issues/1605 vs https://gitlab.gnome.org/World/Rust/poppler-rs/-/merge_requests/4#note_2367084

Please sort this between yourselves.

Maybe this is a gtk bug, though this is the correct fix for this to work on any os with any renderer on 4.16 and 4.18.

In any case I'm not interested in pursuing this further (be it changing this patch, opening a bug report on gtk or discussing this further).

Have a good day

Doublonmousse: Sorry for lashing out yesterday ...

This is kinda me bottling up 6 months of worn patience from dealing with a broken `poppler-rs` release (you can understand my frustration when upstream publishes a new package, does not check CI to verify this builds anymore on target platforms -windows was broken and seemingly it's not a CI platform ?-,  breaks everyone depending on this, does not follow your advice that `gir` is correct since a while ago and the gir file is to be updated and takes 6 months to realize that the `gir` file is outdated. As a matter of fact, the update that broke build was already obsolete by the time the bad version was pushed 9 months ago. All the while, I believed all this time that this was not just a gir update because of these conversations. The result is an issue that should have been caught in the first place and could be fixed in 20 minutes -this is the time I took me to open the gir book and find out the gir file was indeed outdated- took 6 months).

So when I saw once again a disagreement between you two, I wanted out. That you would disagree on a PR _after_ it was merged signals to me you didn't communicate well beforehand. (I certainly wouldn't want this to happen to a contributor because it's an uncomfortable situation to be in).

For the fractional scaling part, I don't really know if something is wrong or not (I'm not totally surprised that there would still be plenty of integer scaling going on with gtk4 as not all of the fractional scaling stuff is pure fractional scaling, and will probably never truly be until gtk5).

Printing the surface size from the gl context directly, I have that the size shown in the debugger is the integer scale factor * the size given the the surface width/height method (and not multiplied by the fractional scale factor, regardless of the renderer of dmabuf status)

Screenshot_20250411_212150

I will open an issue on the gtk gitlab so that this is clarified.

And yes, I won't be surprised this won't work in the future (a lot of things change or can break between major gtk 4 version)  though I'm content with this working until the next big gtk 4.20 release next year (if it breaks by then maybe I'll take another look).

bilelmoussaoui: First of all, for a bit of context, I only took of the maintainership of poppler-rs because no one really updates it after it was initially created by someone else. I don't use it, I don't need it and I don't use anything that depends on it. What I am doing on all the crates in World/Rust is beyond what anyone could during their free time, which is why things progress very slowly especially lately.

If you want to take over the maintainership of poppler-rs just le me know and will give it to you for free.

Doublonmousse: @sdroege I've asked here for the original issue https://gitlab.gnome.org/GNOME/gtk/-/issues/7489#note_2413072 so it seems using the scaling factor is actually correct here, though more for historic reasons.

I reckon more recent developments (dma-buf) use the fractional scaling instead.

Maybe that will change with an ulterior gtk4 (or 5 ?) version though (but that would be a breaking change + we can deal with it when the time comes).

@bilelmoussaoui I'm verry verry sorry for the messages (I think they deserve to be deleted). I'll try to do something constructive now (instead of inexcusable verbal abuse) and send more PR your way for `poppler-rs` and generally get more involved in the project. 

There may be other users interested in becoming contributors/maintainers, though no certainty here.

I can't really say  "let me take maintainership" now, cause that would probably leave a bad taste and I don't want that. I'd rather we have regular contributors/contributions (others or me) first before maintainer status can be discussed. It's true as well that there wasn't _that_ much activity on the repo, I was pretty surprised to open the first ever issue on the gitlab, so that's unfair to fault you on that.

 (I understand  that maintaining something that you don't need/use means that it'll not be where you'll spend most of your time, and it might be compared to spending time on something you really care about. Hence the need for users -> contributors)

PR #2039, opened 2025-04-26
image: Install python3-devel

bilelmoussaoui:
needed for g-i


bilelmoussaoui:
actually maybe we don't need it anymore since the updated image of gtk-rs-core

PR #2056, opened 2025-05-16
fix(event_controller): set_static_name UAF fixed, name now has static lifetime

### LINKED ISSUE #2053: EventControllerKey.set_static_name SEGFAULT
**Bug description**

When I use the method `EventControllerKey.set_static_name` with a string literal param, then I add this controller to the window of the app, and then I open the GTK Inspector and visit the `Controllers` tab, I see one controller (I pressume it's the one I setted) with garbage characters (non-UTF-8), after navigating for a while, the application crashes with the Segmentation Fault (core dump) message.

I think it's because of the lifetime of the param not being `'static` and the compiler optimizing it somehow.
```rust
fn set_static_name(name: Option<&str>) { ... }
```

**Please provide a code sample as small as possible which reproduces the bug.**
```rust
fn main() -> glib::ExitCode {
    let app: gtk::Application = glib::Object::builder().build();

    app.connect_activate(|app| {
        let window = gtk::ApplicationWindow::new(app);
        let key_controller = gtk::EventControllerKey::new();
        key_controller.set_static_name(Some("foo"));
        window.add_controller(key_controller);
        window.present()
    });

    app.run()
}
```

**Backtrace**

~~N/A because it doesn't panic, the process just get killed by the OS.~~

So the error is non-deterministic, I tried another run and the app didn't crash at any moment, but the garbage characters were still there

otaxhu:
Fixes #2053

### DIFF
--- a/gtk4/Gir.toml
+++ b/gtk4/Gir.toml
@@ -103,7 +103,6 @@ generate = [
     "Gtk.EditableProperties",
     "Gtk.EmojiChooser",
     "Gtk.EntryIconPosition",
-    "Gtk.EventController",
     "Gtk.EventControllerFocus",
     "Gtk.EventControllerMotion",
     "Gtk.EventControllerScrollFlags",
@@ -1173,6 +1172,14 @@ status = "generate"
     name = "get_entry"
     rename = "entry"
     manual = true # upcast to Entry
+
+[[object]]
+name = "Gtk.EventController"
+status = "generate"
+manual_traits = ["EventControllerExtManual"]
+    [[object.function]]
+    name = "set_static_name"
+    manual = true # to make use of a static lifetimed string
 
 [[object]]
 name = "Gtk.EventControllerKey"
--- a/gtk4/src/event_controller.rs
+++ b/gtk4/src/event_controller.rs
@@ -11,7 +11,7 @@ pub trait EventControllerExtManual: IsA<EventController> + 'static {
     #[cfg(feature = "v4_8")]
     #[cfg_attr(docsrs, doc(cfg(feature = "v4_8")))]
     #[doc(alias = "gtk_event_controller_set_static_name")]
-    fn set_static_name(&self, name: Option<&'static str>) {
+    fn set_static_name(&self, name: Option<&'static glib::GStr>) {
         unsafe {
             ffi::gtk_event_controller_set_static_name(
                 self.as_ref().to_glib_none().0,
--- a/gtk4/src/event_controller.rs
+++ b/gtk4/src/event_controller.rs
@@ -1,0 +1,24 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+#[cfg(feature = "v4_8")]
+use glib::translate::*;
+
+#[cfg(feature = "v4_8")]
+use crate::ffi;
+use crate::{prelude::*, EventController};
+
+pub trait EventControllerExtManual: IsA<EventController> + 'static {
+    #[cfg(feature = "v4_8")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "v4_8")))]
+    #[doc(alias = "gtk_event_controller_set_static_name")]
+    fn set_static_name(&self, name: Option<&'static str>) {
+        unsafe {
+            ffi::gtk_event_controller_set_static_name(
+                self.as_ref().to_glib_none().0,
+                name.to_glib_none().0,
+            );
+        }
+    }
+}
+
+impl<O: IsA<EventController>> EventControllerExtManual for O {}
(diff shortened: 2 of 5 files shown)


otaxhu:
Ready for review @bilelmoussaoui

gtk4/src/event_controller.rs
```diff
@@ -0,0 +1,24 @@
+// Take a look at the license at the top of the repository in the LICENSE file.
+
+#[allow(unused_imports)]
+use glib::translate::*;
+
+use crate::prelude::*;
+#[allow(unused_imports)]
+use crate::{ffi, EventController};
```
bilelmoussaoui, line 8:
```suggestion
#[cfg(feature = "v4_8")]
use glib::translate::*;

#[cfg(feature = "v4_8")]
use crate::{ffi, prelude::*, EventController};
```

gtk4/Gir.toml
```diff
@@ -1174,6 +1173,15 @@ status = "generate"
...
     manual = true # upcast to Entry
 
+[[object]]
+name = "Gtk.EventController"
+status = "generate"
+manual_traits = ["EventControllerExtManual"]
+    [[object.function]]
+    name = "set_static_name"
+    manual = true # to make use of a static lifetimed string
+    doc_trait_name = "EventControllerExtManual"
```
bilelmoussaoui, line 1183:
```suggestion
```

bilelmoussaoui:
Please rebase for the CI to pass.

bilelmoussaoui:
& squash the commits into a single one.

gtk4/src/event_controller.rs
otaxhu: prelude::* needs to be in scope in order for `IsA<EventController>` to work

bilelmoussaoui (approved):

gtk4/src/event_controller.rs
```diff
@@ -0,0 +1,24 @@
...
+
+#[cfg(feature = "v4_8")]
+use crate::ffi;
+use crate::{prelude::*, EventController};
+
+pub trait EventControllerExtManual: IsA<EventController> + 'static {
+    #[cfg(feature = "v4_8")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "v4_8")))]
+    #[doc(alias = "gtk_event_controller_set_static_name")]
+    fn set_static_name(&self, name: Option<&'static str>) {
```
sdroege, line 14:
```suggestion
    fn set_static_name(&self, name: Option<&'static glib::GStr>) {
```

it needs to be a NUL-terminated string, and `&str` is not

sdroege:
```suggestion
    fn set_static_name(&self, name: impl Into<Option<&'static glib::GStr>>) {
```

if you want it to be a bit more generic

otaxhu: https://gtk-rs.org/gtk-rs-core/stable/latest/docs/src/glib/translate.rs.html#558-590

That's already covered by the function `to_glib_none`

otaxhu: It's better to have API simplicity and accept a `&str`.

sdroege: This doesn't fix anything though. The temporary string created by `to_glib_none()` is freed and then you have a dangling pointer inside GTK.

otaxhu: Do you think we should ***remove*** this function and encourage people to use `set_name`?

The reason being that API users will have to convert the string into a `CStr` then into a `GStr`, resulting in very verbose code.

`GStr` doesn't have `From` traits, but it has `TryFrom<&'a CStr>`, so even if the function signature is `fn set_static_name(name: impl Into<Option<&'static glib::GStr>>)` it won't work.

An example of calling `set_static_name`:
```rust
let controller: gtk::EventController = ...;
controller.set_static_name(Some(c"Hello".try_into().unwrap()));
```

Doing so will represent a breaking change

otaxhu: Nobody has reported this error since 3 years ago, so I think is less of a problem to break the API.

https://github.com/gtk-rs/gtk4-rs/blame/main/gtk4/src/auto/event_controller.rs#L148

otaxhu:
PR converted to draft due what @sdroege commented https://github.com/gtk-rs/gtk4-rs/pull/2056#discussion_r2094142830

gtk4/src/event_controller.rs
sdroege: You can do `glib::gstr!("blablabla")` to get a `&'static glib::GStr`

otaxhu: Ok, I'm gonna add that to the PR and re-open it

PR #2064, opened 2025-05-22
Don't skip `Option` for nullable collections

fengalin:
Consequence of https://github.com/gtk-rs/gir/pull/1622


gsk4/Gir.toml
```diff
@@ -99,12 +99,6 @@ status = "generate"
     [[object.function]]
     name = "new"
     manual = true
-    [[object.function]]
```
bilelmoussaoui, line 102: please remove this commit, those functions are not generated automatically and as shown below, they cannot be yet.

fengalin: This is automatically detected by `Gir` now, so the annotation is no longer necessary. We discussed it there: https://github.com/fengalin/gtk4-rs/commit/4bf3298cacb6654b6834c50153c4095d72a2648f#r157572515, or did I miss something?

bilelmoussaoui: They should not be generated as commented code, that has always been the case...

bilelmoussaoui: and without it we don't generate docs

fengalin: ok removed

gtk4/src/auto/string_list.rs
```diff
@@ -16,7 +16,7 @@ glib::wrapper! {
 
 impl StringList {
     #[doc(alias = "gtk_string_list_new")]
-    pub fn new(strings: &[&str]) -> StringList {
+    pub fn new(strings: Option<&[&str]>) -> StringList {
```
bilelmoussaoui, line 19: those two don't make sense to have as nullable

fengalin: done

fengalin:
Postponing these changes to a future Rust bindings generator rewrite

PR #2073, opened 2025-06-09
fix misleading comments in book GObject memory management

jackaperkins:


jackaperkins:
realized this was correct when viewing the entire code file as written in the listing, so i'll remove this pr

PR #2107, opened 2025-07-30
gdk: Add more memory formats

rmader:
See https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/8330

---

Required for https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/merge_requests/2408

cc @bilelmoussaoui @sdroege - not sure if this could have been autogenerated - opening the MR anyway as I need it locally for testing :)


sdroege:
This would be autogenerated when the new gir files are imported, which I just did. See https://github.com/gtk-rs/gtk4-rs/pull/2101 :)

PR #2127, opened 2025-08-25
ci/msvc: Pick meson 1.8.4

bilelmoussaoui:
As 1.9.0 seems to have a regression


sdroege:
Seems to be a regression in the glib or gtk build system instead

PR #2133, opened 2025-09-03
Fix Panic due to wrong image path in examples/clipboard

rafialhamd:
Updated with correct image path to FIX **Panic Error:**

thread 'main' panicked at examples\clipboard\main.rs:111:58:
called `Result::unwrap()` on an `Err` value: Error { domain: g-io-error-quark, code: 1, message: "Error opening file C:\\github.com\\gtk-rs\\gtk4-rs\\examples\\examples\\clipboard\\asset.png: No such file or directory" }

### DIFF
--- a/examples/clipboard/main.rs
+++ b/examples/clipboard/main.rs
@@ -107,7 +107,7 @@ fn build_ui(application: &gtk::Application) {
         .spacing(24)
         .build();
 
-    let file = gio::File::for_path("./examples/clipboard/asset.png");
+    let file = gio::File::for_path("./clipboard/asset.png");
     let asset_paintable = gdk::Texture::from_file(&file).unwrap();
 
     let image_from = gtk::Image::builder()


bilelmoussaoui:
The path would always be wrong based on from where you run the demo...

sdroege:
Maybe the file can just be `include_bytes!` and then we use a memory input stream here?

rafialhamd:
It's common to run the 'examples ' from the 'examples' directory.
So, here is where it is run:
```
$ pwd
/c/github.com/gtk-rs/gtk4-rs/examples
```
Below works fine. 'Diplays image' in the about dialog
```
$ cargo run --bin about_dialog --features="v4_6"
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.05s
     Running `C:\github.com\gtk-rs\gtk4-rs\target\debug\about_dialog.exe`
```
Since the above can pickup image correctly, I expect the below to do so.
But, below one fails, as it is attempting to open `C:\\github.com\\gtk-rs\\gtk4-rs\\examples\\examples\\clipboard\\asset.png` while it should be `C:\\github.com\\gtk-rs\\gtk4-rs\\examples\\clipboard\\asset.png`

```
$ cargo run --bin clipboard
   Compiling gtk4-rs-examples v0.11.0 (C:\github.com\gtk-rs\gtk4-rs\examples)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.76s
     Running `C:\github.com\gtk-rs\gtk4-rs\target\debug\clipboard.exe`

thread 'main' panicked at examples\clipboard\main.rs:111:58:
called `Result::unwrap()` on an `Err` value: Error { domain: g-io-error-quark, code: 1, message: "Error opening file C:\\github.com\\gtk-rs\\gtk4-rs\\examples\\examples\\clipboard\\asset.png: No such file or directory" }

```
And this PR has the fix.

rafialhamd:
> Maybe the file can just be `include_bytes!` and then we use a memory input stream here?

I believe that should be solution, and NOT my PR.

> The path would always be wrong based on from where you run the demo...

Just found that my PR code works in `examples` BUT NOT from the parent directory.
So, the PR fixes one issue and causes another.

This needs a separate issue/fix.

PR #2135, opened 2025-09-08
gdk4-win32: Allow windows 0.61 and 0.62

sdroege:


sdroege:
@bilelmoussaoui This was not supposed to be merged as-is

sdroege:
I.e. I don't even know if this passes without changes on Windows :)

sdroege:
It does not, but already because meson or GTK are broken.

bilelmoussaoui:
> It does not, but already because meson or GTK are broken.

Right, one of the reasons why I went ahead and merged things anyway because there is no way to check...

sdroege:
Let me just revert the second commit for now then.

PR #2154, opened 2025-09-30
fix(gdk4-wayland): `egl` feature does not compile without `wayland-crate` feature

JakeStanger:
Previously enabling `egl` (with or without a version feature flag) would fail to compile, as some required imports were incorrectly gated behind the `wayland_crate` flag only.

This amends the feature flags in `wayland_display` to amend this.


JakeStanger:
CI docs failure appears unrelated. Have confirmed docs compile for all of the following combinations:

- No features
- `v4_4`, `egl`
- `v4_4`, `egl`, `wayland_crate`
- `wayland_crate`

sdroege (approved):

PR #2161, opened 2025-10-17
fix(gtk4): remove deprecated warnings on style_context methods

JakeStanger:
The crate incorrectly marks the bindings to `gtk_style_context_add_provider_for_display` and `gtk_style_context_remove_provider_for_display` as deprecated, linking to the same methods as alternatives.

The official docs do not mark this functions as deprecated:
https://docs.gtk.org/gtk4/type_func.StyleContext.add_provider_for_display.html
https://docs.gtk.org/gtk4/type_func.StyleContext.remove_provider_for_display.html

### DIFF
--- a/gtk4/src/style_context.rs
+++ b/gtk4/src/style_context.rs
@@ -5,8 +5,6 @@ use glib::translate::*;
 use crate::{ffi, prelude::*, StyleContext, StyleProvider};
 
 impl StyleContext {
-    #[deprecated(note = "Use gtk::style_context_add_provider_for_display instead.")]
-    #[doc(alias = "gtk_style_context_add_provider_for_display")]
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
Why was this merged? this is incorrect. Those are just functions and are not part of that type.

sdroege:
This is indeed confusing. Sorry for missing that, I'll revert

sdroege:
Reverted

JakeStanger:
Sorry I'm not sure I understand?

I have spotted I've removed the doc aliases which is definitely incorrect.

sdroege:
`gtk::style_context_add_provider_for_display()` vs. `gtk::StyleContext::add_provider_for_display()`

PR #2174, opened 2025-11-23
book: Fix main event loop examples

zhao-gang:


Hofer-Julian:
Thanks. Seems like we have an outdated field in mdbook.toml. Can you please remove that as well?

zhao-gang:
Hi @Hofer-Julian , updated book.toml and now the book can be built.

Hofer-Julian (approved):
Thank you!

PR #2178, opened 2025-11-26
ci: enhance Docker workflow with QEMU and Buildx setup

hamkido:
Added steps to set up QEMU and Docker Buildx in the GitHub Actions workflow for improved multi-platform builds. Updated the platforms to include linux/amd64 and linux/arm64.


bilelmoussaoui:
What is the purpose of those images if they are not even used by our CI?

PR #2191, opened 2025-12-02
Rust 2024 edition preparations: Run cargo fix --edition

A6GibKm:
This should be squashed.

### DIFF
--- a/gtk4/src/subclass/accessible_text.rs
+++ b/gtk4/src/subclass/accessible_text.rs
@@ -598,10 +598,12 @@ unsafe extern "C" fn accessible_text_set_caret_position<T: AccessibleTextImpl>(
     accessible_text: *mut ffi::GtkAccessibleText,
     position: u32,
 ) -> glib::ffi::gboolean {
-    let instance = &*(accessible_text as *mut T::Instance);
-    let imp = instance.imp();
+    unsafe {
+        let instance = &*(accessible_text as *mut T::Instance);
+        let imp = instance.imp();
 
-    imp.set_caret_position(position).into_glib()
+        imp.set_caret_position(position).into_glib()
+    }
 }
 
 #[cfg(feature = "v4_22")]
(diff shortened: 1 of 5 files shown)


sdroege:
Needs updating gir

A6GibKm:
> Needs updating gir

Done.

sdroege:
@A6GibKm Can you update similar to the gtk-rs-core PR?

A6GibKm:
Rebased. Lets see what CI says.

sdroege:
Looks like there's still something wrong with the namespaced version features:

```
error: unexpected `cfg` condition value: `gio_v4_18`
  --> gtk4/src/auto/shortcuts_shortcut.rs:62:16
   |
62 |     #[cfg_attr(feature = "gio_v4_18", deprecated = "Since 4.18")]
   |                ^^^^^^^^^^-----------
```

But apart from that this is probably ready?

sdroege:
This also needs rebasing

sdroege:
So where does this `gio_v4_18` come from exactly, what is even deprecated there since when? :)

A6GibKm:
> So where does this `gio_v4_18` come from exactly, what is even deprecated there since when? :)

This https://docs.gtk.org/gtk4/property.ShortcutsShortcut.icon.html got deprecated. I suppose the namespace is wrongly set to gio since the return type is GIcon.

sdroege:
You also want to set `resolver = "3"` in `Cargo.toml` and change `expr_2021` to `expr` in macro definitions (`cargo fix` changed them from `$foo:expr` to `$foo:expr_2021`).

A6GibKm:
There are no uses of expr_2021 at the moment. I am not sure why the --fix did not pull those in.

A6GibKm:
> You also want to set `resolver = "3"` in `Cargo.toml` and change `expr_2021` to `expr` in macro definitions (`cargo fix` changed them from `$foo:expr` to `$foo:expr_2021`).

Set to `resolver = 3`.

sdroege:
> There are no uses of expr_2021 at the moment. I am not sure why the --fix did not pull those in.

Was not necessary here then. Good!

A6GibKm:
CI passes now.

sdroege (approved):

PR #2215, opened 2026-01-11
cargo: Set edition to 2024

A6GibKm:
Followup to https://github.com/gtk-rs/gtk4-rs/pull/2191.


sdroege:
Please fold that into the other PR :)

PR #2230, opened 2026-01-30
book: use `Rc::clone` to make reference-counting explicit

azzamsa:
>  By using Rc::clone for reference counting, we can visually distinguish between the deep-copy kinds of clones and the kinds of clones that increase the reference count. 

The Rust Programming Language, p. 640


Hofer-Julian:
Thanks, but I'd rather leave it as is to stay consistent with the rest of the book

azzamsa:
what if I change all related bit to `Rc::clone` in the book?

Hofer-Julian:
> what if I change all related bit to `Rc::clone` in the book?

Please don't :)

Most `clone` calls in the book, also aren't `Rc::clone`

azzamsa:
I see, Thanks!

PR #2273, opened 2026-03-21
main_event_loop.md: Port ashpd features from async-std to async-io

ratijas:
`async-std` is no longer a feature of `ashpd` crate, and `async-io` replaced it. In fact, the whole `async-std` is deprecated in favor of `smol`.

### DIFF
--- a/book/src/main_event_loop.md
+++ b/book/src/main_event_loop.md
@@ -214,7 +214,7 @@ It can be configured to depend on `async-std`.
 We can add it to our dependencies by running the following command.
 
 ```
-cargo add ashpd --no-default-features --features "gtk4 async-std"
+cargo add ashpd --no-default-features --features "gtk4 async-io"
 ```
 
 You need to use a Linux desktop environment in order to run the following example locally.
--- a/book/src/main_event_loop.md
+++ b/book/src/main_event_loop.md
@@ -214,7 +214,7 @@ It can be configured to depend on `async-std`.
 We can add it to our dependencies by running the following command.
 
 ```
-cargo add ashpd --no-default-features --features "gtk4 async-io"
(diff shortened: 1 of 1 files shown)


ratijas:
Looking into `ashpd` source code:

```rs
#[cfg(all(all(feature = "tokio", feature = "async-io"), not(doc)))]
compile_error!("You can't enable both async-io & tokio features at once");
#[cfg(all(not(feature = "tokio"), not(feature = "async-io"), not(doc)))]
compile_error!("Either the `async-io` or the `tokio` feature has to be enabled");
```

I don't know the whole story, but it looks like `tokio` XOR `async-io`. I'm not sure how `smol` fits into this picture; I guess `smol` is a direct alternative to `gio` event loop?

Also, `account` feature needs to be enabled for the presented example to work.

```rs
#[cfg_attr(docsrs, doc(cfg(feature = "account")))]
#[cfg(feature = "account")]
pub mod account;
```

ratijas:
Also the example uses `reqwest@0.12`, but `0.13.2` is available, and [`0.13` has brought breaking changes](https://github.com/seanmonstar/reqwest/releases/tag/v0.13.0) in features:

> `rustls` is now the default TLS backend, instead of `native-tls`.
> `rustls-tls` has been renamed to `rustls`.

Hofer-Julian:
Can you please check this out @bilelmoussaoui?

ratijas:
There are couple more recommendations of `async-std` on the page, they should probably be either dropped or replaced with `async-io` depending on context. I'm still figuring this out.

bilelmoussaoui:
Updated the deps in main. Thanks! any further clarifications can be handled separately

PR #2275, opened 2026-03-23
Fix directory path of asset.png

Oelnbod:
This fixes the directory path of `asset.png` in the clipboard example. Otherwise, this program panics when running from the examples or clipboard directory


bilelmoussaoui:
this depends on from where you run the example, it will always break in some scenario given it is part of a workspace. Not sure we can do much about it but this is not a solution

PR #2316, opened 2026-06-12
Add bindings to TryExpression

antmh:
This allows having fallbacks in case evaluating an expression fails. Also added a helper method on Expression to chain these expressions similarly to property and closure expressions.


bilelmoussaoui:
thanks for your patch, let us wait till we land https://github.com/gtk-rs/gtk4-rs/pull/2317 so the ci can pass here

bilelmoussaoui:
Please rebase

antmh:
Rebased, also fixed an issue with the `chain_try` method not passing `self` and an unused generic (I think I copied another chain method while adding the new one and I forgot to update it)

PR #2320, opened 2026-06-13
subclass: Add missing BuildableImpl vfuncs

bilelmoussaoui:


sdroege (approved):
Not very nice API but can't think of anything better either

PR #2321, opened 2026-06-13
gdk-wayland: Add _raw wrappers

### LINKED ISSUE #2071: Wayland objects without wayland_crate feature
I want to access the libwayland pointers returned by gdk-wayland but I have no need for the dependencies pulled in by the wayland_crate feature.

I do not want to compile these dependencies and I do not want to spend effort unwrapping the returned objects just to get the pointers that were returned by gdk-wayland. I also do not want to use the C functions directly as it would require me to write additional unsafe code.

I do want to use thin safe wrappers that do nothing more than call the gdk functions and return the pointers.

bilelmoussaoui:
Fixes #2071

### DIFF
--- a/gdk4-wayland/src/wayland_surface.rs
+++ b/gdk4-wayland/src/wayland_surface.rs
@@ -1,13 +1,13 @@
 // Take a look at the license at the top of the repository in the LICENSE file.
 
-#[cfg(feature = "wayland_crate")]
-#[cfg_attr(docsrs, doc(cfg(feature = "wayland_crate")))]
+use std::ffi::c_void;
+use std::ptr::NonNull;
+
 use glib::translate::*;
 #[cfg(feature = "wayland_crate")]
 #[cfg_attr(docsrs, doc(cfg(feature = "wayland_crate")))]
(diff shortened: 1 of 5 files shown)


gdk4-wayland/src/wayland_surface.rs
```diff
@@ -1,20 +1,28 @@
...
 use crate::ffi;
 use crate::{WaylandSurface, prelude::*};
 
 // rustdoc-stripper-ignore-next
 /// Trait containing manually implemented methods of
 /// [`WaylandSurface`](crate::WaylandSurface).
 pub trait WaylandSurfaceExtManual: IsA<WaylandSurface> + 'static {
+    #[doc(alias = "gdk_wayland_surface_get_wl_surface")]
+    #[doc(alias = "get_wl_surface")]
+    fn wl_surface_raw(&self) -> Option<NonNull<c_void>> {
```
sdroege, line 20: Is there some -sys crate we could use for the pointer types? Or should we define type aliases for them? `c_void` is not very nice :)

Otherwise looks good to me

bilelmoussaoui: nop there isn't and i think c_void is good enough if people want the raw pointer.

PR #2348, opened 2026-07-29
Change to_string() to as_str() in widget.rs

lq259:
The method `to_string()` allocates new memory even if you are not writing to the string. `to_string()` is well suited for events where you are writing to the string, but because this function just reads an immutable variable, `as_str()` is more appropriate, mainly because `as_str()` is meant for read-only operations and does not allocate new memory.

Cargo test and cargo check passed.

### DIFF
--- a/gtk4/src/subclass/widget.rs
+++ b/gtk4/src/subclass/widget.rs
@@ -1049,7 +1049,7 @@ pub unsafe trait WidgetClassExt: ClassStruct {
                         .unwrap();
                     let activate_callback = *internal
                         .actions
-                        .get(&action_name.to_string())
+                        .get(action_name.as_str())
                         .unwrap_or_else(|| {
                             panic!("Action name '{}' was not found", action_name.as_str());
                         });


lq259 (commented):
Just realized this affected other places... Look at the failed checks.
Edit: Overreaction, had never seen CI before. Thanks!

sdroege:
> Just realized this affected other places... Look at the failed checks.

How are the failed checks related to this change?

sdroege:
Anyway, change looks correct to me and feel free to also add a commit to this PR that fixes the CI. Otherwise I'll get to it at some point

lq259:
Idk, just looked at it closer now. Clippy complaining, seems unrelated. Thanks!
Edit: And security vulnerabilities...

lq259:
Still complaining... The rest is beyond my capability :)

gtk4-macros/src/lib.rs
A6GibKm, line 75: This one also does an unnecessary allocation.

sdroege:
Rebased so the CI passes. That also dropped one of the changes as that was in main already.

PR #2349, opened 2026-08-01
gtk-macros: Remove unnecessary allocation

A6GibKm:
The field.indent should only be allocated if the iterator did not yield any value.


A6GibKm:
It is unfortunate that we cannot use a 

    value_id.unwrap_or_else(|| &indent.to_string());

sdroege:
Rebased so the CI passes.

PR #2357, opened 2026-08-29
Remove GL examples making use of epoxy

sdroege:
The epoxy crate is unmaintained for >8 years, depends on lots of outdated dependencies and now prevents us from running `cargo update` because xml-rs 0.7 was yanked from crates.io because someone doesn't know what yanking is meant for.


sdroege:
@bilelmoussaoui I have a local change that keeps the examples while dropping epoxy. Cleaning up now

sdroege:
See https://github.com/gtk-rs/gtk4-rs/pull/2358

